PR #7, opened 2021-02-25
Add Github Workflows

philipp-fensch:
Adds 3 Workflows:

- container.yaml: Builds Docker-Container from Dockerfile.dev-centos nad pushes it to Docker-Hub
- build_test.yaml: Builds dpsim on Windows & Linux, executes one example and runs cppcheck
- generate.yaml: Generate Docs


n-eiling:
What still has to be changed:
- in build_test.yml und generate.yml: "container:" should use the correct tag name
- container.yml: use correct tag name and add gitlab secrets in the projects settings. Maybe the URL to the gitlab registry has to be added as well.

m-mirz:
I am continuing this work in a new branch: https://github.com/sogno-platform/dpsim/tree/github-workflows

PR #22, opened 2021-07-29
Create new python interface and convert python examples

JTS22:
This PR greatly expands on the [pybind11](https://pybind11.readthedocs.io/en/stable/)-interface for DPsim, exposing most of the classes and methods required for setting up, running and interacting with simulations in python, including
- SP, DP and EMT single and three phase component classes
- Functions to print out attribute information for component objects
- Simulation, RealTimeSimulation and SystemTopology classes
- DecouplingLine classes
- Event classes
- Bindings between STL- and Eigen-Containers and their python / numpy equivalents

Additionally, most of the Jupyter notebooks which previously just ran simulations by calling the compiled C++-examples or by using the old python interface got converted to use the new interface.

This PR also sets the C++-version for compilation to C++17. For the python interface to compile correctly, a minimum version of C++14 is required. For compilation to work with C++11, further changes to the binding code would be required, a description of which can be found in the top-level CMakeLists.txt


Include/dpsim/pybind/DPComponents.h
```diff
@@ -0,0 +1,31 @@
...
+#include <pybind11/complex.h>
+#include <pybind11/stl.h>
+#include <pybind11/functional.h>
+#include <pybind11/eigen.h>
+
+#include <dpsim/Simulation.h>
+#include <dpsim/RealTimeSimulation.h>
+#include <cps/IdentifiedObject.h>
+#include <cps/CIM/Reader.h>
+#include <DPsim.h>
```
m-mirz, line 20: @JTS22 would it be possible to move these dpsim includes to the cpp files?

CMakeLists.txt
```diff
@@ -214,15 +218,18 @@ add_subdirectory(models)
...
 
+
 if(DPSIM_BUILD_EXAMPLES)
 	add_subdirectory(Examples)
 endif(DPSIM_BUILD_EXAMPLES)
 
 if(WITH_PYBIND)
-	set(PYBIND11_CPP_STANDARD -std=c++11)
+	set(DPSIMPY_INCLUDE_DIRS
+		${CMAKE_CURRENT_BINARY_DIR}/Include/dpsim/pybind
```
m-mirz, line 228: @JTS22 just for curiosity... why have you added the binary directory here?

JTS22: I probably just copied it from the other include directives. It should not be necessary, so i removed the binary dir include in 786a0ab2985ce510ac46da2823dd4a1f94899d07

Include/dpsim/pybind/DPComponents.h
JTS22: yes, did that in 5c0d595dbba0a90d0a5e84122cf42474193f1c42

PR #39, opened 2021-09-16
ci: add minimal build of DPsim

stv0g:
Signed-off-by: Steffen Vogel <post@steffenvogel.de>


stv0g:
Every time I am trying to build DPsim it fails as the CI only tests a full-featured build.
Often CIM or Sundial headers are included without proper `#ifdefs` to check for their support.

PR #47, opened 2021-09-17
rm gitlab docker build and use gh actions

m-mirz:
Signed-off-by: Markus Mirz <mmirz@eonerc.rwth-aachen.de>


stv0g:
Agreed. But there is stull a lot of gruft in the old gitlab ci comfig here

m-mirz:
Since we have moved to github actions, this is not necessary anymore.

PR #60, opened 2021-12-16
fix spdlog formatting string

JTS22:
This PR fixes a formatting string for log messages in the EMT Ph1 VoltageSource. Previously, this log call would throw an error because a float is expected for the first format parameter, but an integer is given.


m-mirz:
This PR fixes a formatting string for log messages in the EMT Ph1 VoltageSource. Previously, this log call would throw an error because a float is expected for the first format parameter, but an integer is given.

PR #64, opened 2022-01-19
Reduced-order VBR models

dinkelbachjan:
This PR includes the reduced-order VBR synchronous generator models. It encompasses the model orders 3 (one-axis model), 4 (two-axis model), 6a (Marconato's model) and 6b (Anderson-Fouad's model). The models in the SP domain are validated against PSAT results, while the models in DP/EMT domain are validated against the SP domain models in a dedicated scenario.

4th order PCM and DCIM implementations will be part of #79.


Examples/Cxx/Circuits/SP_SynGen3OrderVBR_SMIB_Fault.cpp
```diff
@@ -0,0 +1,208 @@
...
+Real setPointActivePower=300e6;
+Real setPointVoltage=1.05*VnomMV;
+
+// HV line parameters referred to MV side
+Examples::Grids::CIGREHVAmerican::LineParameters lineCIGREHV;
+Real lineLength = 100;
+Real lineResistance = lineCIGREHV.lineResistancePerKm * lineLength * std::pow(ratio,2);
+Real lineInductance = lineCIGREHV.lineReactancePerKm * lineLength * std::pow(ratio,2) / nomOmega;
+Real lineCapacitance = lineCIGREHV.lineSusceptancePerKm * lineLength / std::pow(ratio,2) / nomOmega;
+Real lineConductance = 1e-15;
```
dinkelbachjan, line 28: Why did you choose here `1e-15` instead of `8e-2`? Maybe it's better practice to add the alternative value for the conductance parameter as well to `Examples::Grids::CIGREHVAmerican::LineParameters`, to place a comment there what its purpose is and use this alternative parameter (e.g. named `lineConductancePerKmAlternative`) then everywhere in order to harmonise

Examples/Cxx/Circuits/SP_SynGen4OrderDQ_SMIB_Fault.cpp
```diff
@@ -0,0 +1,205 @@
+#include <DPsim.h>
```
dinkelbachjan, line 1: File should be renamed to `Examples/Cxx/Circuits/SP_SynGen4OrderDCIM_SMIB_Fault.cpp`

Examples/Cxx/Circuits/SP_SynGen4OrderDQ_SMIB_Fault.cpp
```diff
@@ -0,0 +1,205 @@
...
+	Real initMechPower = initActivePower;
+
+	// Nodes
+	std::vector<Complex> initialVoltage_n1{ n1PF->voltage()(0,0), 0.0, 0.0};
+	std::vector<Complex> initialVoltage_n2{ n2PF->voltage()(0,0), 0.0, 0.0};
+	auto n1SP = SimNode<Complex>::make("n1SP", PhaseType::Single, initialVoltage_n1);
+	auto n2SP = SimNode<Complex>::make("n2SP", PhaseType::Single, initialVoltage_n2);
+
+	// Components
+	auto genSP = SP::Ph1::SynchronGenerator4OrderDQ::make("SynGen", Logger::Level::debug);
```
dinkelbachjan, line 102: Model should be renamed as well to `SynchronGenerator4OrderDCIM`

Examples/Cxx/Circuits/SP_SynGen4OrderDQ_SMIB_Fault.cpp
```diff
@@ -0,0 +1,205 @@
...
+	
+	simSP.run();
+}
+
+int main(int argc, char* argv[]) {	
+
+	// Command line args processing
+	CommandLineArgs args(argc, argv);
+	//Real SwitchClosed = 1e-3 * (24*24/555);
+	Real SwitchClosed = 1e3;
```
dinkelbachjan, line 172: Why here no per-unit conversion, as in the line before?

models/Include/cps/Base/Base_SimpSynchronousGenerator.h
```diff
@@ -0,0 +1,209 @@
+/* Copyright 2017-2021 Institute for Automation of Complex Power Systems,
```
dinkelbachjan, line 1: This is the class that you name as `ReducedOrderSynchronousGenerator` in your thesis (Fig 3.1), right? I find this term clearer, so I would also use it here in the code

Examples/Cxx/Circuits/SP_ReducedOrderSG_VBR_Load_Fault.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+	// Simulation parameters
+	Real SwitchClosed = GridParams.SwitchClosed;
+	Real SwitchOpen = GridParams.SwitchOpen;
+	Real startTimeFault = 1.0;
+	Real endTimeFault   = 1.1;
+	Real finalTime = 5;
+	Real timeStep = 1e-3;
+	Real H = syngenKundur.H;
+	int SGModel = 4;
```
dinkelbachjan, line 126: I think it's better that `SGModel` is of type `String`. Then you can use `'6a'` and `'6b'` instead of `6` and `7`, which can become confusing in particular if we have models `6c`, `6d` etc. in the future

Examples/Notebooks/Circuits/SP_Validation_ReducedOrderSG_VBR_SMIB_Fault.ipynb
```diff
@@ -0,0 +1,1156 @@
...
+   "metadata": {},
+   "source": [
+    "# Validation of the 3th, 4th and 6th Order VBR SynGen against PSAT"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This example is taken from: P. Kundur, \"Power System Stability and Control\", Example 13.2, pp. 864-869."
```
dinkelbachjan, line 14: Is this correct? In [SP_ReducedOrderSG_SMIB_Fault](Examples/Cxx/Circuits/SP_ReducedOrderSG_SMIB_Fault.cpp) you are using according to [this](https://github.com/sogno-platform/dpsim/blob/3dc0f557d3d55ed6dc06f9333ee326fe3e59dc44/Examples/Cxx/Circuits/SP_ReducedOrderSG_SMIB_Fault.cpp#L12) line the `555MW` machine (`H=3.7`), instead of the `2220MW` machine (`H=3.5`), which is defined under the here written Kundur reference according to [Examples.h](https://github.com/sogno-platform/dpsim/blob/3dc0f557d3d55ed6dc06f9333ee326fe3e59dc44/Examples/Cxx/Examples.h#L89).

models/Source/Base/Base_ReducedOrderSynchronGenerator.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+	mRightVector = Matrix::Zero(leftVector->get().rows(), 1);
+	mMnaTasks.push_back(std::make_shared<MnaPreStep>(*this));
+	mMnaTasks.push_back(std::make_shared<MnaPostStep>(*this, leftVector));
+}
+
+template <typename VarType>
+void Base::ReducedOrderSynchronGenerator<VarType>::MnaPreStep::execute(Real time, Int timeStepCount) {
+	mSynGen.mSimTime = time;
+	mSynGen.stepInPerUnit(); //former system solve (trapezoidal)
```
dinkelbachjan, line 294: Comment unclear. Can be removed?

models/Source/SP/SP_Ph1_SynchronGenerator4OrderDCIM.cpp
```diff
@@ -0,0 +1,124 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *********************************************************************************/
+
+#include <cps/SP/SP_Ph1_SynchronGenerator4OrderDCIM.h>
+
+using namespace CPS;
+
+// !!! TODO: 	Adaptions to use in SP_Ph1 models phase-to-ground peak variables
```
dinkelbachjan, line 13: I assume this is a copy/paste mistake and should be removed.

Examples/Cxx/Circuits/SP_ReducedOrderSG_VBR_Load_Fault.cpp
martinmoraga: I agree but it is not possible to use strings with the option -o of the command line. Any idea how I can do it?

dinkelbachjan: This will be possible after I have resolved #77

models/Source/SP/SP_Ph1_SynchronGeneratorVBR.cpp
```diff
@@ -0,0 +1,85 @@
...
+	mIntfCurrent(0, 0) = Math::complexFromVectorElement(leftVector, mVirtualNodes[1]->matrixNodeIndex());
+	Matrix Iabc = Matrix::Zero(2,1);
+	Iabc << mIntfCurrent(0, 0).real(), mIntfCurrent(0, 0).imag();
+	mIdq = mComplexAToDq * Iabc / mBase_I_RMS;
+}
+
+Matrix SP::Ph1::SynchronGeneratorVBR::get_DqToComplexATransformMatrix() {
+	Matrix dqToComplexA(2, 2);
+	dqToComplexA <<
+		cos(mThetaMech),	-sin(mThetaMech),
```
dinkelbachjan, line 81: Why is here the argument of `sin` and `cos` set to `mThetaMech` and not `mThetaMech-mNomOmega*mSimTime`?

dinkelbachjan: I would assume that due to the negligence of `mBase_OmMech*mSimTime` in the interface transformation, your voltages/currents in the grid are currently oscillating with `mBase_OmMech` (60Hz). Maybe @martinmoraga haven't seen this, because the focus was on the machine variables (especially the electrical torque), which are in the machines dq reference frame. It also should not have had any impact on the static grid simulation, but it should be corrected, so that the angles of the grid voltage/currents are really the ones in the SP domain.

dinkelbachjan: This was fixed in 15657fe1e34048455929da970fb5e99750740364. The varying voltage in SP at the SG terminal was compensated by a varying voltage of the external network injection due to wrong parametrisation. Both issues are fixed now.

Examples/Cxx/Circuits/SP_ReducedOrderSG_VBR_Load_Fault.cpp
dinkelbachjan: #77 has been resolved and merged into this branch. It can be implemented now using
https://github.com/sogno-platform/dpsim/blob/02babaea27ffc515c35b3b6404d719954c7b10cb/Include/dpsim/Utils.h#L146

PR #68, opened 2022-02-02
Mat2dpsimpy

gnakti:
This PR implements:

- Matpower import and powerflow simulation in DPsim using dpsimpy
- DPsim adaptation: Add node apparent power in order to compare results with Matpower.
- Pybind interface adaptation: Add missing set_base_voltage of transformer SPcomponents


m-mirz:
This is included in #87

PR #74, opened 2022-02-04
Update VILLASnode hash in install script

dinkelbachjan:
Signed-off-by: Jan Dinkelbach <jdinkelbach@eonerc.rwth-aachen.de>

### DIFF
--- a/Packaging/Shell/install-fedora-deps-dev.sh
+++ b/Packaging/Shell/install-fedora-deps-dev.sh
@@ -1,3 +1,5 @@
+# Note: Make sure to execute the shell script from the root of the dpsim directory
+
 VILLAS_VERSION=0.8.0
 CIM_VERSION=CGMES_2.4.15_16FEB2016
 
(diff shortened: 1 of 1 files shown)


stv0g:
Is this still up to date?

dinkelbachjan:
It is up-to-date in the sense that it should still be avoided that we use a branch reference from VILLASnode in our setup script. But the Commit-Hash is deprecated here and should be instead the same as used in the [Dockerfile.dev](https://github.com/sogno-platform/dpsim/blob/master/packaging/Docker/Dockerfile.dev).

stv0g:
Thanks @dinkelbachjan for the update.

> [...] the sense that it should still be avoided that we use a branch reference from VILLASnode

Okay, yes thats probably a better idea.. In #168, I added the dpsim-branch again. But I will revert this to a commit-hash.
In #168, I also tested the build with the most recent commit of VILLASnode which works fine. So we can take this one.

Lets close this PR and update the hash by #168

PR #87, opened 2022-03-30
restructure python packaging

m-mirz:
This PR adds a new dpsim python package wrapping the pybind package. The new package includes pure python code to read matpower case files.


m-mirz:
Update image in Dockerfile

.github/workflows/build_test_linux_fedora.yaml
```diff
@@ -89,12 +89,21 @@ jobs:
       run: |
         chmod -R +x ./build/Examples/Cxx
 
+    - name: Debug
```
m-mirz, line 92: remove the debugging part

PR #110, opened 2022-08-11
State-Space Nodal Solver

dinkelbachjan:
This branch focusses on the implementation of a State Space Nodal simulation approach for real time analysis within the EMT domain based on the paper “A Combined State-Space Nodal Method for the Simulation of Power System Transients” ,doi: [10.1109/TPWRD.2010.2090364](https://doi.org/10.1109/TPWRD.2010.2090364) of Dufour et al. .

Linear State Space Nodal (SSN) models of basic components and linear circuits like an RLC circuit are implemented as single SSN components. Additionally to Resistive Companion MNA models they hold internal states. The simulation sequence and structure does not have to be changed for these components, only the used methods itself differ in the executed calculations and quantity updates.

For non-linear components an iterative solution approach is implemented in a new solver which is derived from the MnaEigenSparseSolver and is called SSNSolver. The Newton-Raphson numerical root finding algorithm is used to find system solutions for each step. Within each iteration, non-linear SSN components are updated in a dedicated function call. Non-linear components such as diodes or synchronous generators possess additional methods which are called in this function. These are used to update their states as well as their Jacobian and actual output function values. The last two are used in the Netwon-Raphson iteration.


gnakti:
@m-mirz 
This PR can be closed. The work was merged as part of PR https://github.com/sogno-platform/dpsim/pull/175.
The [ssn](https://github.com/sogno-platform/dpsim/tree/ssn) branch can be deleted.

PR #112, opened 2022-08-31
Fix devcontainer.json

PJungkamp:
The change in capitalization of top-level directories broke the
`devcontainer.json` file.

Fix the path to `Dockerfile.dev`.

Signed-off-by: Philipp Jungkamp <philipp.jungkamp@rwth-aachen.de>


JTS22:
Oops, seems like I missed that because I rebased onto your first PR only after changing all the pathnames...

https://github.com/sogno-platform/dpsim/blob/85a17d3041ecef57d03bdf60ead735e2f5e2799f/.devcontainer/devcontainer.json#L10-L11
Is there a specific reason for starting the container as privileged and without seccomp restrictions? What is the usecase that depends on these settings?

m-mirz:
@stv0g is `--privileged` required for villas in general or only for the shmem interface?

stv0g:
It should not be required anymore

m-mirz:
The update branch button here did not work as I expected. I will merge this into a new branch to fix this before merging into the master.

PR #113, opened 2022-08-31
Revision of power flow  solver

gnakti:
This PR revises the PF solver of DPsim and adapts the MATPOWER import script & cim++ Reader class to the new changes.
matpower cases 9, 14, 39, 118 are also validated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # <img src="docs/images/dpsim.png" width=40 /> DPsim
 
-[![Build & Test CentOS](https://github.com/sogno-platform/dpsim/actions/workflows/build_test_linux_centos.yaml/badge.svg)](https://github.com/sogno-platform/dpsim/actions/workflows/build_test_linux_centos.yaml)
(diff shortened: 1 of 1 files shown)


gnakti:
- The revision started in an older branch: powerflow-revision. The branch has two commits that have bee cherry-picked in this new branch revise-pf-solver . powerflow-revision can be removed.

martinmoraga:
I had to comment the [lines 264 - 269 of the file PFSolver.cpp](https://github.com/sogno-platform/dpsim/blob/7f3857d0b235a075c62d96ae5bb5f2c4b569dbb8/dpsim/src/PFSolver.cpp#L264) because when you try to run the examples [case9bus ](https://github.com/sogno-platform/dpsim/blob/revise-pf-solver/examples/Notebooks/Grids/case9.ipynb) or [case14bus](https://github.com/sogno-platform/dpsim/blob/revise-pf-solver/examples/Notebooks/Grids/case14.ipynb) you get that the voltage of all nodes with Loads are 0. That is because the base voltage of the EnergyConsumers are 0 (error in cim files?). 

In my opinion we should improve the funcion `PFSolver::determineNodeBaseVoltages()` to check that the base voltages are not zero. And display some error if that is the case (or throw an exception) because find the reason of the error was really difficult!

Issue #75, opened 2022-02-06
Refactor VBR Models


dinkelbachjan:
- [ ] Reuse the order-specific auxiliar constants in the domain specific VBR models
  - Example 4th order: 
    - [EMT_Ph3_SynchronGenerator4OrderVBR](https://github.com/sogno-platform/dpsim/blob/3dc0f557d3d55ed6dc06f9333ee326fe3e59dc44/models/Source/EMT/EMT_Ph3_SynchronGenerator4OrderVBR.cpp#L66)
    - The same constants are also used by `DP_Ph1_SynchronGenerator4OrderVBR` and `SP_Ph1_SynchronGenerator4OrderVBR`
  - It should be possible to define the order-specific auxiliar constants already in `Base_ReducedOrderSynchronGenerator`
  - Same goes for 3rd and 6th order

- [ ] Remove `Ef` from auxiliar constants
  - Example 4th order:
    - [EMT_Ph3_SynchronGenerator4OrderVBR](https://github.com/sogno-platform/dpsim/blob/3dc0f557d3d55ed6dc06f9333ee326fe3e59dc44/models/Source/EMT/EMT_Ph3_SynchronGenerator4OrderVBR.cpp#L71)
  - Remove `Ef` and rename `Cq` to `Dq`

PR #115, opened 2022-09-06
Fix implementation of exciter and new turbine governor model

martinmoraga:
@dinkelbachjan @m-mirz 

- fix implementation of exciter DC1A model 
  - https://github.com/sogno-platform/dpsim/blob/master/dpsim/examples/cxx/signals/Exciter.cpp is out-of-date 
- new turbine governor TGOV1 (Ref: Milano, Power system modelling and scripting p. 358)
- partially solves points of #75


m-mirz:
Follow up PR #120

PR #117, opened 2022-09-08
Reduced order SGs extensions

dinkelbachjan:
This PR provides further examples with reduced-order synchronous generators: a SMIB system with a load step event and the WSCC benchmark system. Besides, it includes the following:
- The logging of the indices of the variable matrix entries for reduced-order VBR models.
- The removal of the uncofigurable magnetizing reactance from the powerflow transformer model to align the model with the ones used in dynamic simulations.
- The use of the SG's initial active power from the terminals as initial mechanical input power instead of the predefined mechanical power when this value is set by means of initializeFromNodesAndTerminals.
- The reduction of the required accuracy for a time event from 1000ns to 100ns.


m-mirz:
I think that the new C++ examples being used in jupyter test notebooks still need to be added here:
https://github.com/sogno-platform/dpsim/blob/cfb4e2cf63ded5384169dc1d09e7604a98a6e785/dpsim/examples/cxx/CMakeLists.txt#L108

dinkelbachjan:
I have checked that all the C++ examples, which are currently used in the Jupyter notebooks, are added under `TEST_SOURCES`. Probably additional Jupyter test notebooks will follow in the future for some of the C++ examples added by this PR.

dinkelbachjan:
The rebase has been done as well. What is actually the motivation behind listing the sources additionally under `TEST_SOURCES`?

m-mirz:
We are not building all C++ examples in the CI anymore to improve the speed. Only examples in `TEST_SOURCES` are are built and can be used in jupyter tests.

m-mirz (approved):

PR #119, opened 2022-09-12
Vs signal gen follow up rebase7

dinkelbachjan:
This pull request contains all commits about follow-up work regarding the signal generator that had been added to dpsim earlier.

The models are now validated by the following two validation notebooks:
- EMT_DP_SP_Slack_PiLine_PQLoad_FrequencyRamp_CosineFM
- DP_Slack_PiLine_VSI_Ramp

In addition to that, the scalingKp parameter of the PV control has been adapted for the CIGRE MV grid in order to speed up the control.


JTS22:
The newly rebased version of this can be found on branch https://github.com/sogno-platform/dpsim/tree/vs-signal-gen-follow-up-rebase8. Therefore, this PR should either be force-pushed to the rebase8-branch or a new PR should be created.

PR #123, opened 2022-10-10
Reimplement the interface using a lock-free queue

### LINKED ISSUE #111: lock-free queue for interface to villas etc.
Could be based on this implementation:
https://moodycamel.com/blog/2013/a-fast-lock-free-queue-for-c++.htm

Requirements:
- provide lock-free communication between two threads in dpsim, independent of communication interface implementation
- C++ not C

JTS22:
This PR provides a new implementation for the DPsim `Interface` and `InterfaceVillas` classes. While in before, the read and write actions of the interface were run in sync with the simulation, they are now executed in their own threads asynchronously. To communicate with the simulation, a [lock-free queue](https://github.com/cameron314/readerwriterqueue) is used. To accommodate for the changes to the interface system, the procedure for creating and configuring an interface also has changed slightly, now including some new options for controlling blocking and non-blocking operations.

Requires #121 to be merged first.
Closes #111


dpsim-villas/include/dpsim-villas/InterfaceSampleBased.h
```diff
@@ -59,7 +59,8 @@ namespace DPsim {
 				for (auto attr : intf.mImportAttrs) {
 					mModifiedAttributes.push_back(attr);
 				}
-				mModifiedAttributes.push_back(Scheduler::external);
+				//TODO: Is this necessary / what effect does a dependency on external have?
```
m-mirz: Main reason for indicating the external dependency is to stop the scheduler from dropping the task.

dpsim/src/Simulation.cpp
```diff
@@ -177,6 +171,7 @@ void Simulation::sync() {
 	mLog->info("Start synchronization with remotes on {} interfaces", numOfSyncInterfaces);
 
 	for (auto ifm : mInterfaces) {
+		//TODO: Change or document this synchronization behavior
 		if(ifm.syncStart) {
```
m-mirz: This is used to synchronize two real-time simulators in the beginning of a co-simulation. During the co-simulation, the two simulators may rely only on their own clocks and no synchronization mechanism to minimize the impact of communication on the real-time performance / variance of the timestep.

dpsim/src/Simulation.cpp
```diff
@@ -177,6 +171,7 @@ void Simulation::sync() {
 	mLog->info("Start synchronization with remotes on {} interfaces", numOfSyncInterfaces);
 
 	for (auto ifm : mInterfaces) {
+		//TODO: Change or document this synchronization behavior
```
m-mirz: This is used to synchronize two real-time simulators in the beginning of a co-simulation. During the co-simulation, the two simulators may rely only on their own clocks and no synchronization mechanism to minimize the impact of communication on the real-time performance / variance of the timestep.

cmake/GetReaderWriterQueue.cmake
```diff
@@ -0,0 +1,18 @@
+include(FetchContent)
+
+FetchContent_Declare(
+  readerwriterqueue
+  GIT_REPOSITORY    https://github.com/cameron314/readerwriterqueue
+  GIT_TAG           v1.0.6
+)
+
+# TODO: Is this really still necessary? Or can we just use cmake 3.14?
```
m-mirz, line 9: If we fix and update the rocky linux setup / test, we could require cmake 3.14 as minimum.

dpsim/src/CMakeLists.txt
```diff
@@ -23,7 +24,9 @@ set(DPSIM_INCLUDE_DIRS
 	)
 
 set(DPSIM_LIBRARIES
-	dpsim-models)
+	dpsim-models
+	readerwriterqueue #TODO: Is there any need to only add this + the interface classes conditionally / introduce a new cmake variable for this?
```
m-mirz, line 28: I think it is ok to add this unconditionally.

dpsim-villas/include/dpsim-villas/InterfaceSampleBased.h
JTS22: I think, as long as there are other tasks who depend on the attributes in mModifiedAttributes, the task should be scheduled either way (I just tried that and it seems to work). If there are no other tasks which depend on this tasks, not scheduling it should actually be the proper choice. For this specific use case it should not matter though, so we can just leave the dependency on external in.

dpsim/src/Simulation.cpp
JTS22: Knowing that it makes sense to have that mechanism in place, however it was not really documented before how this can be en-/disabled and what exactly it does. Through the comments in the `InterfaceVillas` class, the option to turn it on or off for every attribute is explained now, the exact procedure however (having two dpsim-exports and one import) still remains undocumented.

cmake/GetReaderWriterQueue.cmake
JTS22: Sounds great! I actually have no idea what's even the issue there. I might look into it in the near future.

dpsim/src/CMakeLists.txt
JTS22: I would agree. Removing the comment...

dpsim/include/dpsim/Interface.h
```diff
@@ -34,38 +34,78 @@ namespace DPsim {
...
+
+	public:
+		class PreStep : public CPS::Task {
+		public:
+			PreStep(Interface& intf) :
+				Task(intf.mName + ".Read"), mIntf(intf) {
+				for (auto attr : intf.mImportAttrs) {
+					mModifiedAttributes.push_back(attr);
+				}
+				//TODO: Is this necessary / what effect does a dependency on external have?
```
m-mirz: As mentioned before, the external dependency avoids that the scheduler drops the task.

dpsim-models/include/dpsim-models/Attribute.h
```diff
@@ -361,9 +361,6 @@ namespace CPS {
 		 * @return true if the copy operation was successful, false otherwise
 		 */
 		virtual bool copyValue(AttributeBase::Ptr copyFrom) {
-			if (!this->isStatic()) {
```
m-mirz: What problem occurred for dynamic attributes and how have you solved it?

dpsim/src/Simulation.cpp
m-mirz: I am collecting open points related to docs in #135

dpsim-villas/src/InterfaceWorkerVillas.cpp
```diff
@@ -215,15 +220,29 @@ void InterfaceWorkerVillas::readValuesFromEnv(std::vector<Interface::AttributePa
 						mCurrentSequenceInterfaceToDpsim++;
 					}
 				}
+
+				if (!pollFds.empty()) {
+					//Manually clear the event file descriptor since Villas does not do that for some reason
```
m-mirz, line 224: Do we have an issue for that in villas?

m-mirz (approved):

dpsim-models/include/dpsim-models/Attribute.h
JTS22: I think, I excluded dynamic attributes at first because they are usually meant to be computed from their dependencies and not set directly. However, since setting them directly is necessary when importing to a dynamic attribute with no dependencies configured (e. g. `mVoltageRef`), I removed the requirement for static attributes in 80f8f9eca423a1aadbc0a8fc24d9afa088a574e4. This works in general, but when importing to a dynamic attribute which actually has dependencies, the copied value will be overwritten immediately the next time the attribute is read, so the copy has no effect in this case.

dpsim-villas/src/InterfaceWorkerVillas.cpp
JTS22: I do not think so, I will create one and link it in the code.

JTS22: Linked to issue in d2ca027e3d0446a8fc232bf4f5c2ad120f0c9f22

JTS22: https://git.rwth-aachen.de/acs/public/villas/node/-/issues/347

PR #134, opened 2022-11-11
Fix EMT_Ph1_Inductor initialization

martinmoraga:
This PR fixes a small error in the calculation of the initial current of EMT_Ph1_Inductor and adapts the example VS_RL1 so that the inductor current is validated now with respect to Modelica reference results.


m-mirz (dismissed):

dinkelbachjan:
I agree. Do you know any test case where this makes a difference and which could be used for future validation of this model in the CI?

martinmoraga:
The initialization errors are invisible for the CI because the node voltages of the examples which are used to validate the inductor ([/dpsim/examples/Notebooks/Circuits/VS_RL1.ipynb](https://github.com/sogno-platform/dpsim/blob/master/examples/Notebooks/Circuits/CS_R2CL.ipynb), https://github.com/sogno-platform/dpsim/blob/master/examples/Notebooks/Circuits/VS_R2L3.ipynb and https://github.com/sogno-platform/dpsim/blob/master/examples/Notebooks/Circuits/VS_RL1.ipynb) are not being initialized which means that all quantities (currents and voltages) of all components (exept the voltage of the voltage source) are initialized to zero. Therefore, the initial values of the components does not matter in the validation.

martinmoraga:
> Can we add for example in VS_RL1 a validation section with initialization?

@dinkelbachjan 

In my last commit I adapted the example VS_RL1 so that the current is also validated. The file with the reference results must be changed:
 - old reference file: the results were validated against simulink and all initial values were set to 0. ([https://git.rwth-aachen.de/acs/public/simulation/reference-results/raw/master/Simulink/Circuits/SL_VS_RL1.csv](https://git.rwth-aachen.de/acs/public/simulation/reference-results/raw/master/Simulink/Circuits/SL_VS_RL1.csv)). 
  - new reference file: now the results are validated against modelica using the right initial values ([MOD_VS_RL1.csv](https://github.com/sogno-platform/dpsim/files/10186713/MOD_VS_RL1.csv))

dinkelbachjan (approved):

Issue #131, opened 2022-11-09
Check, improve and document subcomponent handling


dinkelbachjan:
- [x] Check that explicit subcomponent tasks are only defined when reasonable in terms of expected computational effort, otherwise let the subcomponent's pre- and poststep routines be handled by the supercomponent
- [x] Every `SimPowerComp` has a vector `mSubComponents` of subcomponents but most classes derived from `SimPowerComp` have their subcomponents as separate member variables, not utilizing this list. 
  - Make use of `mSubComponents` to harmonize calling pre- and poststep routines, as for example shown [here](https://github.com/sogno-platform/dpsim/blob/a4c9b5cfc3f40eb0eb80d555d9562455e1f489d0/dpsim-models/src/DP/DP_Ph1_AvVoltageSourceInverterDQ.cpp#L347-L352) 
- [x] Document the different concepts of how to handle pre- and poststep routines of subcomponents

JTS22:
Classes which include subcomponents:
| Class | Uses `mSubComponents` | Pre- and Post-step called by parent? |
|--------------|-----------|------------|
|DP_Ph1_VoltageSourceInverterDQ|yes|yes|
|DP_Ph1_NetworkInjection|yes|yes|
|DP_Ph1_PiLine|no|yes|
|DP_Ph1_PQLoadCS|no|no|
|DP_Ph1_RxLine|no|no|
||||

JTS22:
In addition to the changes made in #141, I propose another way to further unify the behavior of components with subcomponents: By creating a shared superclass MNASimPowerComp that is derived from both SimPowerComp and MNAInterface, it becomes possible to extract the various MNA method calls on the subcomponents into this common superclass. This means that the parent components only register their subcomponents once (e. g. through `addMNASubComponent`) and then only contain methods concerned with the MNA operations in the parent. The calls to the MNA methods of the child are handled by the superclass instead, drastically cutting down on the amount of duplicated code (especially the `for (auto comp : mSubComponents)` loop) in the parent classes. The `MNAPreStep` and `MNAPostStep` classes contained in almost all parent components can also be merged into the common superclass. A first working example of this can be found in https://github.com/sogno-platform/dpsim/tree/unify-mna-subcomponents (Comparison with #141 in https://github.com/sogno-platform/dpsim/compare/subcomponents...unify-mna-subcomponents).

dinkelbachjan:
[...]

PR #144, opened 2022-11-28
Introduce a common subclass for composite components

JTS22:
Introduces two new base classes, `MNASimPowerComp<T>` and `CompositePowerComp<T>`. For all composite components (as defined in #131), `MNASimPowerComp` provides default implementations for all methods in `MNAInterface`, as well as default `MnaPreStep` and `MnaPostStep` tasks, such that `MNAInterface` is only inherited once in the parent class and not in each component individually. Additionally, `CompositePowerComp` provides `mnaParent...` methods for most of the MNA methods in `MNAInterface`. Via the `addMNASubComponent` method, a component can register subcomponents for which the relevant MNA methods are called automatically by the `CompositePowerComp` parent class. For each MNA method, the child component class can decide to either override the normal `mna...` method and provide subcomponent handling themselves, or override just the `mnaParent...` method and let the subcomponents (and some other common operations) get handled by the parent class methods.

While the `MNASimPowerComp` class is not used for any non-composite components in this PR, the unification of MNA methods for these components is covered in #150 .

Based on #142 
Related to #131

### DIFF
--- a/dpsim-models/include/dpsim-models/AttributeList.h
+++ b/dpsim-models/include/dpsim-models/AttributeList.h
@@ -28,7 +28,7 @@ namespace CPS {
 
 		AttributeList() { };
 
-		virtual ~AttributeList() {};
+		virtual ~AttributeList() = default;
 
 		const AttributeBase::Map & attributes() { return mAttributes; };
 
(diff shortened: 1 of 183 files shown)


m-mirz:
Renaming the branch did not work as expected. I cannot reopen the PR so I created a new one #177. For the following PRs I have force push the branch after cherry picking as I did for #142

PR #150, opened 2022-12-05
Introduce common base class MNASimPowerComp - deprecated

JTS22:
Unifies the behavior of all components that previously implemented `MNAInterface` by providing a common base class `MNASimPowerComp<T>`. This base class implements  `MNAInterface` and all its methods in such a way, that common actions such as the creation and registration of the pre- and post-step tasks are done uniformly in the base class. The various `mna...` methods are replaced by a new set of `mnaComp...` methods that can be implemented by the child component classes to provide their own MNA behavior. 

Based on #144


m-mirz:
This one is a litte bit complicated to merge after #120. I am working on it in #178

dinkelbachjan:
Closing, since this has been replaced by #178

PR #152, opened 2022-12-08
PMU signal

gnakti:
This PR implements a measurement device as a signal class ...


gnakti:
Issue: the scheduler drops the tasks of the new PMU signal class

BeijiaYang:
### This is a brief explanation.
The class "PMUSignalDevice" is a new constructed signal-model class which produces the measurement error on the signals from system components. 
We attempt to applying it in a three bus simulation to simulate the measurement error attached on the SimNode. 
Recent problem is: the schedueler keeps the tasks of the new PMUSignalDevice class.

leonardocarreras:
closing it in favour of #258

PR #175, opened 2023-01-17
Linear SSN

dinkelbachjan:
This branch focusses on the implementation of a State Space Nodal (SSN) simulation approach for real time analysis within the EMT domain limited to linear components. The approach is based on the paper “A Combined State-Space Nodal Method for the Simulation of Power System Transients” ,doi: [10.1109/TPWRD.2010.2090364](https://doi.org/10.1109/TPWRD.2010.2090364) of Dufour et al. .

Linear State Space Nodal models of basic components and linear circuits like an RLC circuit are implemented as single SSN components. To achieve this a state space representation for the electrical element or circuit is derived. Those equations are discretized using the trapezoidal rule of integration. The system of linear equations is manipulated such that current states in the output equations of the state space model can be replaced with known terms, i.e. state values of the previous simulation step and input quantities. Previous states are additionally stored within the SSN component and updated when the system has been solved for the current time step. Now the output equations (nodal equations for V-Type SSN-Groups, mesh equations for I-Type SSN-Groups) can be used in the Modified Nodal Analysis approach by the already implemented stamping and task systems. 

Because of that the simulation sequence and structure does not have to be changed for these components, only the used methods itself differ in the executed calculations and the quantity updates for the internally stored states.

Regarding general changes not limited to the SSN approach, a setter method (setParameters) has been added for the EMT::Ph3::CurrentSource. The EMT::Ph3::CurrentSource class and the new EMT::Ph3::SSN::Full_Serial_RLC class were made usable within the pybind python module.

### DIFF
--- a/dpsim-models/src/EMT/EMT_Ph3_CurrentSource.cpp
+++ b/dpsim-models/src/EMT/EMT_Ph3_CurrentSource.cpp
@@ -22,6 +22,30 @@ EMT::Ph3::CurrentSource::CurrentSource(String uid, String name,
   **mIntfVoltage = Matrix::Zero(3, 1);
   **mIntfCurrent = Matrix::Zero(3, 1);
 }
+SimPowerComp<Real>::Ptr EMT::Ph3::CurrentSource::clone(String name) {
+  auto copy = CurrentSource::make(name, mLogLevel);
+  copy->setParameters(attributeTyped<MatrixComp>("I_ref")->get(),
+                      attributeTyped<Real>("f_src")->get());
+  return copy;
+}
+
+void EMT::Ph3::CurrentSource::setParameters(MatrixComp currentRef,
+                                            Real srcFreq) {
+  auto srcSigSine = Signal::SineWaveGenerator::make(**mName + "_sw");
+  // Complex(1,0) is used as initialPhasor for signal generator as only phase is used
+  srcSigSine->setParameters(Complex(1, 0), srcFreq);
+  mSrcSig = srcSigSine;
+
+  **mCurrentRef = currentRef;
+  mSrcFreq->setReference(mSrcSig->mFreq);
+
+  mSLog->info("\nCurrent reference phasor [I]: {:s}"
+              "\nFrequency [Hz]: {:s}",
+              Logger::matrixCompToString(currentRef),
+              Logger::realToString(srcFreq));
+
+  mParametersSet = true;
+}
 
 void EMT::Ph3::CurrentSource::initializeFromNodesAndTerminals(Real frequency) {
   SPDLOG_LOGGER_INFO(
@@ -68,13 +92,6 @@ void EMT::Ph3::CurrentSource::initializeFromNodesAndTerminals(Real frequency) {
   SPDLOG_LOGGER_INFO(
       mSLog, "\n--- Initialization from node voltages and terminal ---");
   mSLog->flush();
-}
-
-SimPowerComp<Real>::Ptr EMT::Ph3::CurrentSource::clone(String name) {
-  auto copy = CurrentSource::make(name, mLogLevel);
-  // TODO: implement setParameters
-  // copy->setParameters(attributeTyped<MatrixComp>("I_ref")->get(), attributeTyped<Real>("f_src")->get());
-  return copy;
 }
 
 void EMT::Ph3::CurrentSource::mnaCompInitialize(
--- a/dpsim-models/src/EMT/EMT_Ph3_SSN_Capacitor.cpp
+++ b/dpsim-models/src/EMT/EMT_Ph3_SSN_Capacitor.cpp
@@ -1,0 +1,238 @@
+/* Copyright 2017-2021 Institute for Automation of Complex Power Systems,
+ *                     EONERC, RWTH Aachen University
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *********************************************************************************/
+
+#include <dpsim-models/EMT/EMT_Ph3_SSN_Capacitor.h>
+
+using namespace CPS;
+
+EMT::Ph3::SSN::Capacitor::Capacitor(String uid, String name,
+                                    Logger::Level logLevel)
+    : MNASimPowerComp<Real>(uid, name, true, true, logLevel),
+      Base::Ph3::Capacitor(mAttributes) {
+  mPhaseType = PhaseType::ABC;
+  setVirtualNodeNumber(1);
+  setTerminalNumber(2);
+  mHistoricVoltage = Matrix::Zero(3, 1);
+  **mIntfVoltage = Matrix::Zero(3, 1);
+  **mIntfCurrent = Matrix::Zero(3, 1);
+}
+
+SimPowerComp<Real>::Ptr EMT::Ph3::SSN::Capacitor::clone(String name) {
+  auto copy = Capacitor::make(name, mLogLevel);
+  copy->setParameters(**mCapacitance);
+  return copy;
+}
+void EMT::Ph3::SSN::Capacitor::initializeFromNodesAndTerminals(Real frequency) {
+
+  Real omega = 2 * PI * frequency;
+  MatrixComp admittance = MatrixComp::Zero(3, 3);
+  admittance << Complex(0, omega * (**mCapacitance)(0, 0)),
+      Complex(0, omega * (**mCapacitance)(0, 1)),
+      Complex(0, omega * (**mCapacitance)(0, 2)),
+      Complex(0, omega * (**mCapacitance)(1, 0)),
+      Complex(0, omega * (**mCapacitance)(1, 1)),
+      Complex(0, omega * (**mCapacitance)(1, 2)),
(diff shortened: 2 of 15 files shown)


dpsim-models/include/dpsim-models/EMT/EMT_Ph1_SSN_Full_Serial_RLC.h
```diff
@@ -0,0 +1,92 @@
...
+        public Base::Ph1::Capacitor,
+        public Base::Ph1::Resistor
+		{
+	protected:
+        Matrix State = Matrix::Zero(2, 1);
+        Matrix yHistory =  Matrix::Zero(1, 1);
+
+		Matrix Dufour_u_n_t = Matrix::Zero(1, 1);
+
+        Matrix Dufour_A_k_hat = Matrix::Zero(2, 2);
```
gnakti, line 44: test

dpsim-models/src/EMT/EMT_Ph3_SSN_Full_Serial_RLC.cpp
georgii-tishenin, line 1: Would it be possible to do something similar to the idea of [Aggregation of Swing Equation](https://github.com/sogno-platform/dpsim/issues/235), but for SSN?
What I was thinking is you can move the logic, that is common for all state-space-nodal components, into a separate class, e.g., `StateSpaceGroup.`  The `StateSpaceGroup` would encapsulate the discrete-state space model and would be responsible for calculating the state,  $y_{khist}$ and $W_{kn}$. 
Then, each of the SSN components: `SSN_Inductor`, `SSN_Capacitor` and `SSN_Full_Serial_RLC`  can have an object of `StateSpaceGroup` class, and set the needed parameters for `StateSpaceGroup` (like, A, B, C, D matrices), get $y_{khist}$ and $W_{kn}$ from `StateSpaceGroup` and use them to update conductance matrix and right-hand-side vector.

dpsim-models/src/EMT/EMT_Ph3_SSN_Capacitor.cpp
```diff
@@ -0,0 +1,162 @@
...
+	mHistoricVoltage = Dufour_B_k_hat * **mIntfCurrent  + **mIntfVoltage;
+	Dufour_B_k_hat = timeStep * (2.0 * **mCapacitance).inverse();
+	Dufour_W_k_n = Dufour_B_k_hat;
+
+	**mRightVector = Matrix::Zero(leftVector->get().rows(), 1);
+}
+
+void EMT::Ph3::SSN::Capacitor::mnaCompApplySystemMatrixStamp(SparseMatrixRow& systemMatrix) {
+	if (terminalNotGrounded(0)) {
+		Math::addToMatrixElement(systemMatrix, matrixNodeIndex(0, 0), mVirtualNodes[0]->matrixNodeIndex(PhaseType::A), -1);
```
georgii-tishenin, line 74: In terms of Dufour's paper terminology: is `SSN_Inductor` a V_Type group, `SSN_Capacitor` an I_Type group and  `SSN_Full_Serial_RLC` a Mixed type group? I've noticed that the logic in `mnaCompApplySystemMatrixStamp()` for `SSN_Capacitor` is different from logic in `SSN_Inductor` and `SSN_Full_Serial_RLC.` Would it be possible to refactor the `SSN_Capacitor` code to have the same stamping logic, as you have managed to do it for `SSN_Full_Serial_RLC?`

dpsim-models/include/dpsim-models/EMT/EMT_Ph1_SSN_Full_Serial_RLC.h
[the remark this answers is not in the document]
MarvinTollnitschRWTH: "Full" just refers to all three components being in series, the voltage being measured across all three components and the current (of course) flowing through all components. But: C in series with parallel L and R would also be an RLC circuit, and even with all components in series the output voltage could be one not measured across all components.

examples/Notebooks/Circuits/EMT_Ph1_SSN_Dufour_L2_C_R_SW_RLC_Validation.ipynb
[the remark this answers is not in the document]
MarvinTollnitschRWTH: As it was part of my thesis, I just wanted to make clear what everything I do is based on. I guess it can be removed now.

dpsim-models/src/EMT/EMT_Ph3_SSN_Capacitor.cpp
MarvinTollnitschRWTH: SSN_Inductor is a V_Type group because the output equation uses the unknown voltage to define the output current. The SSN_Capacitor is an I_Type group because it is the other way around; the unknown quantity which ends up in the left side vector sets the type.

For V_Type goups, the current is defined and the voltage is unknown; thus it has the exact same structure as the MNA approach, GV = I and can be directly stamped. For I_type groups we have the same stamping structure as for voltage sources, where the matrix is expanded by 1x1 (or for three phases 3x3). The extra line contains the compnent's output equation which adds a voltage to the excitation vector (rather than a current) and a current to the left side vector of unknowns (rather than a voltage); it is not a nodal but a mesh equation, so the system gets an extra "virtual node" to include the defining equation.

Following that logic, the SSN_Full_Serial_RLC is an I_Type group, not a mixed type group; if we had two output equations for example, with one output quantity being a current and one being a voltage, then we would have a mixed type group. The output equation for the RLC circuit defines an output current; we just have two state equations for two internal states.

In the Dufour paper it is stated that you can convert the group types into each other, but it would take extra steps and as we follow the MNA approach and not just the nodal approach without mesh equations, I did not look at how that would work, unfortunately.

dpsim-models/src/EMT/EMT_Ph3_SSN_Full_Serial_RLC.cpp
MarvinTollnitschRWTH: In DPsim, we do not have one MnaComponent class for every component because the stamping varies for each component type. Similarily, the stamping structure of different SSN components differ. If we were to outsource the logic for calculating the stamps, we would need to give the StateSpaceGroup member the information about how to do so. If we do that for each component, then we can just do it in the component itself.

Maybe I did not quite understand the task StateSpaceGroup would take over and the benefits that would give. Feel free to correct me, or we will also be able to talk about it in the meeting!

dpsim-models/include/dpsim-models/EMT/EMT_Ph1_Switch.h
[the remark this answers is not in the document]
MarvinTollnitschRWTH: I guess that's possible, but it was a really spontaneous implementation based on the Ph3::Switch because I quickly needed one for Ph1. It also does not implement a "zero cross" mechanic, which I briefly talked about with Ghassen and Jan.

dpsim-models/src/EMT/EMT_Ph3_SSN_Full_Serial_RLC.cpp
georgii-tishenin: Thanks for your reply. You are right, we still need `SSN_Inductor`, `SSN_Capacitor` and `SSN_Full_Serial_RLC`, because calculation of A,B,C,D state-space matrices is different for all of these components, and also stamping is different between NortonEquivalent and TheveninEquivalent.
What I was thinking is that there is still some logic and data, that are common for all State-Space Groups and they can be encapsulated into an object of `StateSpaceGroup` class, that is aggregated by `SSN_Inductor`, `SSN_Capacitor` and `SSN_Full_Serial_RLC`. 
In terms of data, you would then have one `mStateSpaceGroup` member in each of the SSN component classes instead of having `historicCurrent`, `Dufour_B_k_hat`, `Dufour_W_k_n`. SSN component classes can then get these values from `mStateSpaceGroup`.
In terms of logic, `StateSpaceGroup` could take over the responsibility for state-space model numerical integration and calculation of $y_{kHist}$ and  $W_{kn}$.
Another advantage I see, is that the `StateSpaceGroup` itself can be a pretty simple class, implementing the state-space model and equations to calculate $y_{kHist}$ and  $W_{kn}$. It will not have a logic to communicate with the `MNASolver` and will be simpler to understand.

Btw, I've created an issue to [Reuse conductance stamp code](https://github.com/sogno-platform/dpsim/issues/288).

dpsim-models/src/EMT/EMT_Ph3_SSN_Capacitor.cpp
georgii-tishenin: Thank you for a very nice and clear explanation of different types of groups, and for how you are integrating mesh equations into MNA. I am more of a Nodal Analysis guy myself😆, so for me that's a learning as well. This is pretty cool and powerful. Great work😉.
It should be quite easy to convert Thevenin Equivalent to Norton Equivalent, but not necessary though.

gnakti (changes requested):
@MarvinTollnitschRWTH

dpsim-models/src/EMT/EMT_Ph3_CurrentSource.cpp
georgii-tishenin, line 1: Double check/test if implementation of` initializeFromNodesAndTerminals()` is correct

gnakti (approved):

PR #196, opened 2023-03-08
Iterative syngen models

dinkelbachjan:
This PR covers the implementation of iterative reduced-order synchronous generator models, namely the PCM and TPM models.

Based on #171


dinkelbachjan:
Continued by #207

PR #207, opened 2023-03-18
Reduced-order syngen models

dinkelbachjan:
This PR covers the implementation of iterative reduced-order synchronous generator models, namely the PCM and TPM models.
Removes DCIM implementations in DP_Ph1 as they are covered now by PCM models with zero iterations.
Implements a new DP-DQ interface, which is used now by all DP reduced-order synchronous generator models (PCM, TPM, VBR).

Based on #194 
Rebase of #196 

Related to #75


dinkelbachjan:
This PR has been covered by #213

PR #210, opened 2023-03-27
Fix consistency of calculations at first time step

### LINKED ISSUE #195: One-step delay in Voltage source
@dinkelbachjan @gnakti @m-mirz 

When a simulation starts the initial values of all components are calculated for t=0. Hence, the first simulation step should calculate the results for the time $t=\Delta t$ because the results for $t=0$ are the initial values. However, this is not the case for the voltage sorces. In the first simulation step, voltage sources are stamping in the right side vector the value of $t=0$ and not for $t=\Delta t$ which introduces a one step delay between the voltage sources and all other components of DPsim. This behavior translates to transients in the first steps of the simulation. For examples, in a simple 3 phase RLC circuit (initialized with consistent values) we can observe transients in phase b and c of the current of the capacitor:

alt text

To fix this error, the parameter "time" of the functions [EMT::Ph1::VoltageSource::updateVoltage(Real time)](https://github.com/sogno-platform/dpsim/blob/758d47c4c25db49c5dfae4bd3355d40e2b90d144/dpsim-models/src/EMT/EMT_Ph1_VoltageSource.cpp#L65) and [EMT::Ph3::VoltageSource::updateVoltage(Real time)](https://github.com/sogno-platform/dpsim/blob/758d47c4c25db49c5dfae4bd3355d40e2b90d144/dpsim-models/src/EMT/EMT_Ph3_VoltageSource.cpp#L141) must be updated to `time = time + timestep`.

In addition, the log must be adapted accordingly. Currently, the column time of the log files also has a one-step delay, as shown in the table below:
| time  | variable1 | variable2 | ... |
| ------------- | ------------- |  ------------- |  ------------- |
| t=0  | value of variable1 for t=timestep  | value of variable 2 for t=timestep| ... |
| t=timestep | value of variable1 for t=timestep*2 | value of variable1 for t=timestep*2 | ... |
| ... | ... | ... | ... |
| t=k | value of variable1 for t=timestep*(k+1) |value of variable2 for t=timestep*(k+1)| ... |

In my opinion, for time 0 we should log the initial values and the column time should not have a one-step delay:
| time  | variable1 | variable2 | ... |
| ------------- | ------------- |  ------------- |  ------------- |
| t=0  | initial value of variable1  | initial value of variable 2 | ... |
| t=timestep | value of variable1 for t=timestep |value of variable2 for t=timestep | ... |
| ... | ... | ... | ... |
| t=k | value of variable1 for t=timestep*(k) |value of variable2 for t=timestep*(k)| ... |

martinmoraga:
This PR ensures that the calculation of the first simulation step means the calculation of the values for the time instant `t=mTimeStep`, while `t=0` is covered by the initialization values. This was not consistently the case, in particular for `EMT::Ph3::VoltageSource`, which caused corresponding transients at the beginning of some simulations. Moreover, a small error in the intialization of the EMT Power Transformer is fixed. Besides, the scaling of voltages and currents has been harmonised, which is documented now under `docs/Development/Guidelines`.

Closes #195


dpsim-models/src/SimNode.cpp
```diff
@@ -39,10 +39,10 @@ SimNode<VarType>::SimNode(PhaseType phaseType)
 
 template <>
 void SimNode<Real>::initialize() {
-	(**mVoltage)(0,0) = (**mInitialVoltage)(0,0).real();
-	if (phaseType() == PhaseType::ABC) {
-		(**mVoltage)(1,0) = (**mInitialVoltage)(1,0).real();
-		(**mVoltage)(2,0) = (**mInitialVoltage)(2,0).real();
+	if (phaseType() == PhaseType::Single) {
+		(**mVoltage)(0,0) = (**mInitialVoltage)(0,0).real();
```
dinkelbachjan, line 43: Doesn't this need also the multiplication of `mInitialVoltage` with `RMS3PH_TO_PEAK1PH`?

dpsim-models/src/SystemTopology.cpp
```diff
@@ -75,6 +75,7 @@ void SystemTopology::addComponents(const IdentifiedObject::List& components) {
 void SystemTopology::initWithPowerflow(const SystemTopology& systemPF) {
 	for (auto nodePF : systemPF.mNodes) {
 		if (auto node = this->node<TopologicalNode>(nodePF->name())) {
+			// Initiation of phase B and C in 3 Phase systems ?
```
dinkelbachjan, line 78: Needs more explanation. Is this a TODO?

dpsim/src/Simulation.cpp
```diff
@@ -379,7 +379,7 @@ void Simulation::stop() {
 }
 
 Real Simulation::next() {
-	if (mTime < (**mFinalTime + **mTimeStep))
+	if (mTime < **mFinalTime + 1e-12)
```
dinkelbachjan, line 382: This number should specified via a #define directive for example in `dpsim/include/dpsim/Definitions.h`

dpsim-models/src/SimNode.cpp
dinkelbachjan: This has been changed accordingly. The [guideline for the scaling of voltages and currents](https://github.com/sogno-platform/dpsim/blob/4a79f7f95c4149935b0fa1d5f6a3c6050b1b342f/docs/hugo/content/en/docs/Development/Guidelines.md) is now documented.

PR #232, opened 2023-07-17
Disable VILLASfpga to remove libxil dependency

### LINKED ISSUE #231: VILLASnode Examples do not run in current Dev container
When using the dev-container from the current `Dockerfile.dev`, none of the VILLASnode-Examples can be executed, because they require the presence of the `libxil` shared library (https://github.com/VILLASframework/libxil). Because the last push of the dev image to Docker hub happened over 6 months ago, the pipeline still uses an outdated image, in which this dependency was apparently not present. Therefore, the VILLASnode-Examples still work in the GitHub-Pipeline, but not locally.
Options to resolve this are:
1. Compile VILLASnode with the `WITH_FPGA=OFF` option to disable the dependency (as in #232 )
2. Somehow install the dependency manually in the dpsim-Dockerfile
3. Update the VILLASnode version used in the DPsim-containers (see #233), as the issue has been fixed there (https://github.com/VILLASframework/node/pull/694)

JTS22:
Sets the `WITH_FPGA=OFF` compile flag for VILLASnode in the Dockerfile.dev to remove a dependency on the shared library `libxil`, which is currently not installed in the container. Merging this PR should probably be followed by an update for the images on Docker Hub via the `container` workflow.
Closes #231

### DIFF
--- a/packaging/Docker/Dockerfile.dev
+++ b/packaging/Docker/Dockerfile.dev
@@ -91,6 +91,7 @@ RUN cd /tmp && \
 	git checkout ${VILLAS_VERSION} && \
 	cmake ${CMAKE_OPTS} .. \
 		-DCMAKE_INSTALL_LIBDIR=/usr/local/lib64 \
+		-DWITH_FPGA=OFF \
 		-DDOWNLOAD_GO=OFF && \
 	make ${MAKE_OPTS} install && \
 	rm -rf /tmp/villas-node


JTS22:
Great, thank you @n-eiling @stv0g ! I added the use of the updated VILLASnode version as a solution in #231.

dinkelbachjan:
We will go for another solution as described in #231

PR #238, opened 2023-07-27
Fix power flow initialisation of synchronous generators

### LINKED ISSUE #223: Fix power flow initialisation & updating for the terminals
- The terminals of the components are not correctly initialized with initial power and updated with power flow solutions. This might lead to an initialization error in the dynamic simulation (if the terminal power is used in _initializeFromNodesAndTerminals_ e.g. _terminal(0)->singlePower()_)

martinmoraga:
In powerflow simulations, the power of synchronous generators is not being always correctly calculated, since the power of the loads connected to nodes is not considered in the calculation. This PR fixes the calculation of the synchronous generators' power in powerflow simulations and adapts the funcion `initWithPowerFlow` to initialialize the terminal power of the synchronous generators after the powerflow calculation.

Closes #223

### DIFF
--- a/dpsim/src/PFSolverPowerPolar.cpp
+++ b/dpsim/src/PFSolverPowerPolar.cpp
@@ -368,21 +368,22 @@ Real PFSolverPowerPolar::Q(UInt k) {
 }
 
 void PFSolverPowerPolar::calculatePAndQAtSlackBus() {
+	// calculates apparent power injection at slack buses flowing to other nodes (i.e. S_inj_to_other = S_inj - S_shunt, with S_inj = S_gen - S_load)
     for (auto topoNode: mVDBuses) {
         auto node_idx = topoNode->matrixNodeIndex();
 
-        // calculate power flowing out of the node into the admittance matrix (not into PQ Loads)
+        // calculate power flowing out of the node into the admittance matrix (i.e. S_inj)
         CPS::Complex I(0.0, 0.0);
         for (UInt j = 0; j < mSystem.mNodes.size(); ++j)
             I += mY.coeff(node_idx, j) * sol_Vcx(j);
         CPS::Complex S = sol_Vcx(node_idx) * conj(I);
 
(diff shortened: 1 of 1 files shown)


dpsim/src/PFSolverPowerPolar.cpp
```diff
@@ -353,47 +368,96 @@ Real PFSolverPowerPolar::Q(UInt k) {
...
+        for (UInt j = 0; j < mSystem.mNodes.size(); ++j)
+            I += mY.coeff(node_idx, j) * sol_Vcx(j);
+        CPS::Complex S = sol_Vcx(node_idx) * conj(I);
+
+        // GenPower = Power flowing out of node + PQLoadsPower
+        for(auto comp : mSystem.mComponentsAtNode[topoNode])
+            if (auto loadPtr = std::dynamic_pointer_cast<CPS::SP::Ph1::Load>(comp))
+                S += Complex(**(loadPtr->mActivePowerPerUnit), **(loadPtr->mReactivePowerPerUnit));
+
+        // Set power of external network injections of SG (VD)
```
dinkelbachjan, line 385: @martinmoraga 
`// Set power of either VD-type external network injection or VD-type synchronous generator depending on what is connected`
Is this is what you mean here?

dpsim/src/PFSolverPowerPolar.cpp
dinkelbachjan, line 1: It should be reconsidered what we eventually store in `sol_P` and `sol_Q`, as described in more detail in the respective TODO comment lines of his file. Besides, we could consider combining `calculatePAndQAtSlackBus` , `calculateQAtPVBuses` und `calculatePAndQInjectionPQBuses`.

dinkelbachjan (approved):

dpsim/src/PFSolverPowerPolar.cpp
dinkelbachjan: This will not be covered by this PR, but instead will be followed up on by closing #266.

PR #250, opened 2023-10-10
Change default linear solver to KLU

dinkelbachjan:
This PR changes the default linear solver from SparseLU to KLU. KLU performs per default a complete refactorisation. Besides, for partial refactorisation, which can be chosen as an option by the user, a fix is included for `mVariableSystemMatrixEntries` of `DP_Ph1_ReducedOrderSynchronGeneratorVBR` when modelled as Norton equivalent.


dpsim/src/KLUAdapter.cpp
```diff
@@ -18,6 +18,7 @@ KLUAdapter::~KLUAdapter()
         klu_free_symbolic(&mSymbolic, &mCommon);
     if (mNumeric)
         klu_free_numeric(&mNumeric, &mCommon);
+    mSLog->info("Number of Pivot Faults: {}", mPivotFaults);
```
leonardocarreras, line 21: This line might need to use the macro like `SPDLOG_LOGGER_INFO(mSLog, ...);`

dinkelbachjan: Indeed, thanks!

martinmoraga (approved):

Issue #256, opened 2023-10-23
Compilation warnings and some code improvements


leonardocarreras:
**Description of the bug**
During compilation in a new environment, errors related to `"-Woverloaded-virtual"` problems with the compiler arise in various files related to SynchronGenerator* classes and the missing argument `Int timeStepCount` in the `DIakopticsSolver` logger

**To Reproduce**
1. Install a new environment
2. Install dependencies
3. Compile with `cmake --build . -j4 --target dpsimpy`

**Expected behavior**
Build successful

**Desktop (please complete the following information):**
 - OS: Fedora 38

**Additional context**
This warning could also be ignored with pragma instructions, but we actually want to override it, as reading the intention in the code.

The warning is of type:
`warning: 'DerivedClass::function' hides overloaded virtual function [-Woverloaded-virtual]`

n-eiling:
fixed by #293

PR #257, opened 2023-10-23
Fix compilation missing override warnings and function declarations 

leonardocarreras:
This PR changes the signature of the `log` function in the `DiakopticsSolver` files to match the current ones used in the other solvers, where we have now two arguments.

On the other side, the overrides on the `SynchronGenerator` files of the function `clone` had a different signature compared to all the rest of the classes (and did not override the Base class virtual function.

This PR fixes issue #256

### DIFF
--- a/dpsim-models/src/Base/Base_ReducedOrderSynchronGenerator.cpp
+++ b/dpsim-models/src/Base/Base_ReducedOrderSynchronGenerator.cpp
@@ -471,6 +471,10 @@ void Base::ReducedOrderSynchronGenerator<
 }
 
 template <typename VarType>
+void Base::ReducedOrderSynchronGenerator<
+    VarType>::mnaCompApplySystemMatrixStamp(SparseMatrixRow &systemMatrix) {}
+
+template <typename VarType>
 void Base::ReducedOrderSynchronGenerator<VarType>::mnaCompInitialize(
     Real omega, Real timeStep, Attribute<Matrix>::Ptr leftVector) {
 
--- a/dpsim/include/dpsim/DAESolver.h
+++ b/dpsim/include/dpsim/DAESolver.h
@@ -42,6 +42,9 @@ protected:
   /// Nodes of the Problem
   CPS::SimNode<Complex>::List mNodes;
 
+  // Initial time t0
+  Real mT0;
+
   // IDA simulation variables
   /// Memory block allocated by IDA
   void *mem = NULL;
@@ -72,11 +75,12 @@ protected:
 
 public:
   /// Create solve object with given parameters
-  DAESolver(String name, const CPS::SystemTopology &system, Real dt, Real t0);
+  DAESolver(String name, const CPS::SystemTopology &system, Real dt, Real mT0);
   /// Deallocate all memory
   ~DAESolver();
   /// Initialize Components & Nodes with inital values
-  void initialize(Real t0);
+  void initialize() override;
+
   /// Solve system for the current time
   Real step(Real time);
 
--- a/dpsim/src/DAESolver.cpp
+++ b/dpsim/src/DAESolver.cpp
@@ -66,10 +66,11 @@ DAESolver::DAESolver(String name, const CPS::SystemTopology &system, Real dt,
   }
 
   std::cout << "Nodes Setup Done" << std::endl;
-  initialize(t0);
+  mT0 = t0;
+  initialize();
 }
 
-void DAESolver::initialize(Real t0) {
+void DAESolver::initialize() {
   int ret;
   int counter = 0;
   realtype *sval = NULL, *s_dtval = NULL;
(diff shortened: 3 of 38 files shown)


stv0g (approved):
LGTM. Lets merge this.

But in general: we should prefer passing `const std::string &` in favour of `String` as it would copy the string and is a lot slower due to additional copies.

martinmoraga (dismissed):

stv0g:
> Should we make that as a new issue? In a [quick search](https://github.com/search?q=repo%3Asogno-platform%2Fdpsim%20SimPowerComp%3CComplex%3E%3A%3APtr%20clone(String%20name)&type=code), we get this way of doing it by copying in 75 files, versus the only the 4 files modified here...

I can confirm this. All these are actually detected by cppcheck. However, we dont show these performance warnings in the CI log due to:

https://github.com/sogno-platform/dpsim/blob/b17546ac5aa83588b5f854f5f4321d6986ed2935/.github/workflows/build_test_linux_fedora.yaml#L221-L226

m-mirz:
@leonardocarreras sorry, for not merging sooner. Could you please rebase and I will merge it. The changes look good to me too.

leonardocarreras:
@m-mirz now is WIP, the rebase becomes a bit more interesting than expected

dpsim/include/dpsim/DAESolver.h
```diff
@@ -76,6 +76,7 @@ class DAESolver : public Solver {
   /// Deallocate all memory
   ~DAESolver();
   /// Initialize Components & Nodes with inital values
+  void initialize() override;
   void initialize(Real t0);
```
leonardocarreras, line 80: There is a conflict due to the use of `Real t0` in here...

https://github.com/sogno-platform/dpsim/blob/0f4d17c6dae7efa5b6c27f6c1f1aca767a50aa20/dpsim/src/DAESolver.cpp#L156

I do not know if we could do this differently so it does actually override

dpsim/include/dpsim/DAESolver.h
```diff
@@ -42,6 +42,9 @@ class DAESolver : public Solver {
   /// Nodes of the Problem
   CPS::SimNode<Complex>::List mNodes;
 
+  // t0
+  Real mt0;
```
stv0g, line 46:
```suggestion
  Real mT0;
```

dpsim/src/DAESolver.cpp
```diff
@@ -153,7 +154,7 @@ void DAESolver::initialize(Real t0) {
   //	}
   std::cout << "Call IDAInit" << std::endl;
 
-  ret = IDAInit(mem, &DAESolver::residualFunctionWrapper, t0, state, dstate_dt);
+  ret = IDAInit(mem, &DAESolver::residualFunctionWrapper, mt0, state, dstate_dt);
   //	if (check_flag(&ret, "IDAInit", 1)) {
   //		throw CPS::Exception();
   //	}
```
stv0g, line 160: Maybe we should remove this dead code as well?

dpsim-models/src/Base/Base_ReducedOrderSynchronGenerator.cpp
```diff
@@ -470,6 +470,10 @@ void Base::ReducedOrderSynchronGenerator<
   this->mSLog->flush();
 }
 
+template <typename VarType>
+void Base::ReducedOrderSynchronGenerator<VarType>::mnaCompApplySystemMatrixStamp(SparseMatrixRow &systemMatrix){
```
stv0g, line 474: Is your change formatted with clang-format?
```suggestion
void Base::ReducedOrderSynchronGenerator<VarType>::mnaCompApplySystemMatrixStamp(SparseMatrixRow &systemMatrix) {
```

stv0g (changes requested):
LGTM, I only have nits.

leonardocarreras:
Thanks for the suggestions, will adopt them... Just to keep track, there is now also an alternate solution in #293 (that also do some extra things)

dpsim-models/src/Base/Base_ReducedOrderSynchronGenerator.cpp
leonardocarreras: It was not, now will be applied

dpsim/src/DAESolver.cpp
leonardocarreras: Do you mean the actual file or the commented lines below?

n-eiling:
sorry. I didn't look if someone else was already working on it. I hope it doesn't create too much work if we merge my PR first.

leonardocarreras:
If your PR is ready to merge I can rebase after that

dpsim/include/dpsim/DAESolver.h
leonardocarreras: Done

dpsim/src/DAESolver.cpp
```diff
@@ -153,7 +154,7 @@ void DAESolver::initialize(Real t0) {
...
   std::cout << "Call IDAInit" << std::endl;
 
-  ret = IDAInit(mem, &DAESolver::residualFunctionWrapper, t0, state, dstate_dt);
+  ret = IDAInit(mem, &DAESolver::residualFunctionWrapper, mt0, state, dstate_dt);
```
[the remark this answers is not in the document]
leonardocarreras: Done

leonardocarreras:
Moved to #294

PR #258, opened 2023-10-24
PMU signal implementation using GSL

leonardocarreras:
This is the continuation of #152 started by @BeijiaYang, but the changes over time made it obsolete, and we are substituting it with this.

The goal is:

1. To create a new class that provides a measurement-like output.
2. To be able to control the parameters of the measurement signal.

This is still a draft WIP.


pipeacosta:
The PMU implementation requires to build with GSL. This is an option which does not come by default and we don't want to make it a default option. Therefore, we have two options:
- Include a new option in CMakeLists which triggers the GSL installation, something like WITH_PMU
- Leave it as it is, and build with WITH_GSL for PMU support
Both options would require an extra action in the CI

PR #284, opened 2024-04-15
Revision of power flow solver

martinmoraga:
Continuation of PR #113 (it was closed after the rebase of the master branch)

This PR fix some errors of the powerflow solver


python/src/dpsim/matpower.py
```diff
@@ -158,6 +208,49 @@ def create_dpsim_objects(self):
                 # add connections
                 dpsimpy_comp_dict[extnet_name].append([dpsimpy_busses_dict[bus_index]]) # [to bus]
 
+                # # set initial Q is not implemented for slack in DPsim. The mapping was changed here to test cas39.
```
m-mirz, line 211: Why do you include all this code in comments? And why are there two ‚#‘?

python/src/dpsim/matpower.py
```diff
@@ -1,38 +1,32 @@
 import pandas as pd
 import numpy as np
 import scipy.io
-
+import sys
+sys.path.append('/dpsim/build')
```
m-mirz, line 5: This looks like a hack to make it work without installing dpsim or adding dpsim to pythonpath. It only works in the dev container I assume.

stv0g:
```python
import sys
from os.path import dirname as dn

dpsim_dir = dn(dn(dn(dn(__file__))))
sys.path.append(dpsim_dir)
```

Should be more robust?

gnakti: @m-mirz This was added when developing the Matpower reader python module exactly for the two reasons you mention.

python/src/dpsim/matpower.py
gnakti: This PR contains some changes of the Matpower reader. I think all the Matpower reader changes can be removed since the PR focuses on the Power flow solver changes. The Matpower reader latest version is in another PR. Right @martinmoraga ?

martinmoraga: @m-mirz good idea 👍 
I will move this commits

python/src/dpsim/matpower.py
m-mirz: I discussed this with @stv0g  and we concluded that the best solution would be to not include any workaround like this. Please just update your pythonpath when developing… it is just copy & paste from the docs.

python/src/dpsim/matpower.py
martinmoraga: @m-mirz @gnakti 

The commits related to the matpower reader are now in the branch benchmark-networks

dpsim/src/PFSolver.cpp
```diff
@@ -321,6 +321,20 @@ void PFSolver::determineNodeBaseVoltages() {
             "Choose base voltage {}V of {} to convert pu-solution of {}.",
             baseVoltage_, gen->name(), node->name());
         break;
+      }
+      /*else if (std::shared_ptr<CPS::SP::Ph1::Load> load = std::dynamic_pointer_cast<CPS::SP::Ph1::Load>(comp)) {
```
m-mirz, line 325: Why is this code commented?

dpsim/src/PFSolverPowerPolar.cpp
```diff
@@ -98,6 +109,7 @@ void PFSolverPowerPolar::generateInitialSolution(Real time,
             mBaseApparentPower;
         sol_V(pv->matrixNodeIndex()) =
             extnet->attributeTyped<CPS::Real>("V_set_pu")->get();
+        // sol_Q(pv->matrixNodeIndex()) += extnet->attributeTyped<CPS::Real>("q_inj")->get() / mBaseApparentPower; //Todo allow initialisation of p_inj and q_inj in SP_NetworkInjection (use updatePowerInjection?)
```
m-mirz, line 112: @gnakti @martinmoraga please consider moving this to a separate PR as well.

dpsim/src/PFSolver.cpp
```diff
@@ -321,6 +321,19 @@ void PFSolver::determineNodeBaseVoltages() {
             "Choose base voltage {}V of {} to convert pu-solution of {}.",
             baseVoltage_, gen->name(), node->name());
         break;
+      } else if (std::shared_ptr<CPS::SP::Ph1::Load> load =
+                     std::dynamic_pointer_cast<CPS::SP::Ph1::Load>(comp)) {
+        baseVoltage_ = load->attributeTyped<CPS::Real>("V_nom")->get();
+        mSLog->info("Choose base voltage of {}V to convert pu-solution of {}.",
```
leonardocarreras, line 327: Maybe is better with `SPDLOG_LOGGER_INFO(`

dpsim/src/PFSolver.cpp
```diff
@@ -321,6 +321,19 @@ void PFSolver::determineNodeBaseVoltages() {
...
+                    baseVoltage_, load->name(), node->name());
+        break;
+      } else if (std::shared_ptr<CPS::SP::Ph1::NetworkInjection> extnet =
+                     std::dynamic_pointer_cast<CPS::SP::Ph1::NetworkInjection>(
+                         comp)) {
+        baseVoltage_ = extnet->attributeTyped<CPS::Real>("base_Voltage")->get();
+        mSLog->info("Choose base voltage of {}V to convert pu-solution of {}.",
```
leonardocarreras, line 334: I think here is also `SPDLOG_LOGGER_INFO(`

leonardocarreras (approved):

dpsim-models/src/DP/DP_Ph1_PiLine.cpp
georgii-tishenin, line 1: Now there are 23 changed files. But it looks like in most of the files there are just a couple of lines with changed formatting. And a new formatting is different from the .clang-format rules, applied to the rest of the codebase.
It seem that the files with changes, that have influence, are:

- SP_Ph1_NetworkInjection.h,
- SP_Ph1_NetworkInjection.cpp,
- PFSolver.cpp
- PFSolverPowerPolar.cpp

I would then suggest to limit this PR to these 4 files (hope, I did not miss anything)

dpsim-models/include/dpsim-models/SP/SP_Ph1_NetworkInjection.h
```diff
@@ -39,7 +39,7 @@ class NetworkInjection : public CompositePowerComp<Complex>,
 
   // #### Powerflow section ####
   /// Base voltage [V]
-  Real mBaseVoltage;
+  const Attribute<Real>::Ptr mBaseVoltage;
```
georgii-tishenin, line 42: I see this is done the same way for the other components, but does `mBaseVoltage` really need to be of `Attribute<Real>::Ptr` type? ls there a specific functionality of Attributes that you'd like to use for `mBaseVoltage`, like logging, interfacing or task scheduling?
If you just want to get the value of `mBaseVoltage` in PFSolver.cpp a more readable solution could be to have a getter function:
```
public:
    // Getter for mBaseVoltage
    Real getBaseVoltage() const {
        return mBaseVoltage;
    }
```

dpsim/src/PFSolver.cpp
```diff
@@ -321,6 +321,19 @@ void PFSolver::determineNodeBaseVoltages() {
             "Choose base voltage {}V of {} to convert pu-solution of {}.",
             baseVoltage_, gen->name(), node->name());
         break;
+      } else if (std::shared_ptr<CPS::SP::Ph1::Load> load =
+                     std::dynamic_pointer_cast<CPS::SP::Ph1::Load>(comp)) {
+        baseVoltage_ = load->attributeTyped<CPS::Real>("V_nom")->get();
```
georgii-tishenin, line 326: If `mNomVoltage` is intended to be of type `Attribute::Ptr` and is publicly accessible, consider directly accessing its value through the member variable instead of using a hard-coded string. For example:
`baseVoltage_ = load->mNomVoltage->get()`
This will ensure compile time-error checking. Now, if there is a typo in the string "V_nom" the error would not be caught at compile time.

dpsim/src/PFSolver.cpp
m-mirz: Yes, the macro variant can be optimized better if the log level is less than info.

dpsim/src/PFSolver.cpp
leonardocarreras: @georgii-tishenin would you agree that we can do this in a separate PR? I am not saying that the proposed way is not better in some sense... But for this time, if you see the other cases of this else-if-block, all of them are being called the same way, and this will not be consistent with the existing code... see the previous lines for example https://github.com/sogno-platform/dpsim/blob/15b1faf77eeac56532a2fa5e46b1a54853891f62/dpsim/src/PFSolver.cpp#L318 https://github.com/sogno-platform/dpsim/blob/15b1faf77eeac56532a2fa5e46b1a54853891f62/dpsim/src/PFSolver.cpp#L308 https://github.com/sogno-platform/dpsim/blob/15b1faf77eeac56532a2fa5e46b1a54853891f62/dpsim/src/PFSolver.cpp#L299-L300

dpsim-models/include/dpsim-models/SP/SP_Ph1_NetworkInjection.h
leonardocarreras: The same as for the other case, would it be a good place to discuss this in a different PR? If you have a strong opinion, I am inclined to take over and change this. Since this specific time, it looks like there is no specific reason to use attributes... However, I still think maybe #330 can be a better place for discussing the attributes and rules of usage...

dpsim/src/PFSolver.cpp
georgii-tishenin: hey @leonardocarreras. Agree with you.

dpsim-models/src/DP/DP_Ph1_PiLine.cpp
georgii-tishenin: done👍

dpsim-models/src/SP/SP_Ph1_Load.cpp
```diff
@@ -36,8 +36,10 @@ void SP::Ph1::Load::setParameters(Real activePower, Real reactivePower,
   **mReactivePower = reactivePower;
   **mNomVoltage = nominalVoltage;
 
-  SPDLOG_LOGGER_INFO(mSLog, "Active Power={} [W] Reactive Power={} [VAr]",
-                     **mActivePower, **mReactivePower);
+  SPDLOG_LOGGER_INFO(
+      mSLog,
+      "Active Power={}[W]\nReactive Power={} [VAr]\nNominal Voltage={} [V]",
```
stv0g, line 41: I dont think adding newlines to log output is a good practice, as it those new lines are not prefixed with the loggers time, severity and name fields.

dpsim/src/PFSolver.cpp
```diff
@@ -321,6 +321,21 @@ void PFSolver::determineNodeBaseVoltages() {
...
+        SPDLOG_LOGGER_INFO(
+            mSLog, "Choose base voltage of {}V to convert pu-solution of {}.",
```
stv0g, line 328: I think we should add a space between numbers and their unit:
```suggestion
            mSLog, "Choose base voltage of {} V to convert pu-solution of {}.",
```

dpsim/src/PFSolver.cpp
stv0g: I think the macros should only be used in critical code sections. This part of the PF solver is not critical...

stv0g (changes requested):
LGTM with a few nits

dpsim/src/PFSolver.cpp
leonardocarreras: Suggestion adopted

dpsim-models/src/SP/SP_Ph1_Load.cpp
leonardocarreras: Suggestion adopted

stv0g (approved):

Issue #291, opened 2024-05-08
Fix VILLASnode in Dockerfiles


pipeacosta:
After some changes in VILLASnode's structure, mainly related to the removal of VILLAScommon and VILLASfpga as submodules, building the DPsim's docker image from Dockerfile.dev issues an error. The issue will be solved by changing the `VILLAS_VERSION` variable to a more up-to-date commit of VILLASnode.

leonardocarreras:
I think part of your solution is [here](https://github.com/sogno-platform/dpsim/compare/master...fix-devcontainer)

leonardocarreras:
Closed #291 via #292

PR #292, opened 2024-05-08
Fix VILLASnode in Dockerfiles

pipeacosta:
This PR solves #291


stv0g:
LGTM, but can you elaborate why you've choosen a VILLASnode version from February '24 while the last commit is just a week old?

pipeacosta:
Hi @stv0g, There is no specific reason for having used that commit, we only need a stable version of VILLASnode. Do you have any recommendation in this regard?

stv0g:
@pipeacosta I would usually say we should use the latest available commit which is not broken. But I think we can also go ahead with this one and update again in some time..

stv0g (approved):

leonardocarreras (dismissed):
Looks good

Should we also do the same with the scripts? 

https://github.com/sogno-platform/dpsim/blob/f07066ab9941dfcb35e4acaf7086168f2b3462fb/packaging/Shell/install-ubuntu-deps.sh#L4

https://github.com/sogno-platform/dpsim/blob/f07066ab9941dfcb35e4acaf7086168f2b3462fb/packaging/Shell/install-fedora-deps.sh#L4

https://github.com/sogno-platform/dpsim/blob/f07066ab9941dfcb35e4acaf7086168f2b3462fb/packaging/Shell/install-fedora-deps-dev.sh#L4

stv0g:
@pipeacosta Could you also adjust the commit refs in the locations mentioned by @leonardocarreras ?

pipeacosta:
@stv0g done

stv0g (approved):

stv0g:
@m-mirz I think this ready to be merged

PR #293, opened 2024-05-13
fix clang compilation.

n-eiling:
This used to work before but the CI test was not migrated to github and therefore we lost the compatibility. We need clang compilation to make the clangd language server work.

### DIFF
--- a/dpsim/include/dpsim/MNASolverFactory.h
+++ b/dpsim/include/dpsim/MNASolverFactory.h
@@ -59,14 +59,18 @@ public:
 
   /// sovlerImpl: choose the most advanced solver implementation available by default
   template <typename VarType>
-  static std::shared_ptr<MnaSolver<VarType>>
-  factory(String name, CPS::Domain domain = CPS::Domain::DP,
-          CPS::Logger::Level logLevel = CPS::Logger::Level::info,
-          DirectLinearSolverImpl implementation = DirectLinearSolverImpl::KLU,
-          String pluginName = "plugin.so") {
+  static std::shared_ptr<MnaSolver<VarType>> factory(
+      String name, CPS::Domain domain = CPS::Domain::DP,
+      CPS::Logger::Level logLevel = CPS::Logger::Level::info,
+      DirectLinearSolverImpl implementation = mSupportedSolverImpls().back(),
+      String pluginName = "plugin.so") {
     //To avoid regression we use KLU in case of undefined implementation
     if (implementation == DirectLinearSolverImpl::Undef) {
+#ifdef WITH_KLU
       implementation = DirectLinearSolverImpl::KLU;
+#else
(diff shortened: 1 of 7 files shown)


n-eiling:
I noticed a lot of errors in the models due to proper override semantics not being enforced. For example most `clone` implementations were not overriding but only hiding. Same for some `log` implementations. 
I fixed all of what clang catched and also added a CI target to build using clang.

This also fixes a few minor issues with the VILLAS inteface and adds a draft for an FPGA interface.

Let's merge this quickly, to avoid having to touch even more files. @stv0g @m-mirz 

btw: We also lost the cppcheck CI. This was also catching a lot of the mistakes I had to fix here.

stv0g:
LGTM. I also fixed the uninitialized mOpened member in another PR. But let's go ahead and merge this here. 

@m-mirz I am in support of merging this. Can you pull the trigger?

stv0g (approved):

n-eiling:
Pybind is not working with clang because of some template error that are beyond me. I just deactivated pybind in the CI until somehow who has a better understanding of the attribute templates gets around to fixing it.

dpsim/include/dpsim/MNASolverFactory.h
```diff
@@ -59,14 +59,18 @@ class MnaSolverFactory {
...
+      String name, CPS::Domain domain = CPS::Domain::DP,
+      CPS::Logger::Level logLevel = CPS::Logger::Level::info,
+      DirectLinearSolverImpl implementation = mSupportedSolverImpls().back(),
+      String pluginName = "plugin.so") {
     //To avoid regression we use KLU in case of undefined implementation
     if (implementation == DirectLinearSolverImpl::Undef) {
+#ifdef WITH_KLU
       implementation = DirectLinearSolverImpl::KLU;
+#else
+      implementation = mSupportedSolverImpls().back();
```
leonardocarreras, line 72: Is still doing the same (KLU as default)? Maybe is just my bad and I am confused

leonardocarreras (commented):
Looks good to me, many of the changes are 1 to 1 with #257

m-mirz (approved):

dpsim/include/dpsim/DiakopticsSolver.h
```diff
@@ -94,7 +94,7 @@ class DiakopticsSolver : public Solver, public CPS::AttributeList {
   void initMatrices();
   void applyTearComponentStamp(UInt compIdx);
 
-  void log(Real time);
+  void log(Real time, Int timeStepCount);
```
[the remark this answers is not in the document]
m-mirz: We can continue this on #294

dpsim/include/dpsim/MNASolverFactory.h
m-mirz: We can continue this in #294

PR #294, opened 2024-05-13
Enable compilation in gcc 14 and clang 18 (code fixes)

leonardocarreras:
~~This integrates the changes in the #293 and #257 to have all in one place.~~
Now #257 is rebased on the master (#293 merged).

### DIFF
--- a/dpsim-models/include/dpsim-models/DP/DP_Ph3_SynchronGeneratorDQ.h
+++ b/dpsim-models/include/dpsim-models/DP/DP_Ph3_SynchronGeneratorDQ.h
@@ -73,8 +73,9 @@ public:
 
   // #### MNA Functions ####
   /// Initializes variables of component
-  virtual void mnaCompInitialize(Real omega, Real timeStep,
-                                 Attribute<Matrix>::Ptr) override = 0;
+  virtual void
+  mnaCompInitialize(Real omega, Real timeStep,
+                    Attribute<Matrix>::Ptr leftVector) override = 0;
   /// Performs with the model of a synchronous generator
   /// to calculate the flux and current from the voltage vector.
   void mnaStep(Matrix &systemMatrix, Matrix &rightVector, Matrix &leftVector,
--- a/dpsim-models/include/dpsim-models/DP/DP_Ph3_SynchronGeneratorDQODE.h
+++ b/dpsim-models/include/dpsim-models/DP/DP_Ph3_SynchronGeneratorDQODE.h
@@ -45,7 +45,7 @@ public:
 
   // #### MNA Section ####
   void mnaCompInitialize(Real omega, Real timeStep,
-                         Attribute<Matrix>::Ptr leftVector);
+                         Attribute<Matrix>::Ptr leftVector) override;
 
   /// Add MNA pre step dependencies
   void mnaCompPreStep(Real time, Int timeStepCount) override;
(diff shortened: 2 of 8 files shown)


leonardocarreras:
@n-eiling would you try if this integrated PR works for you?

leonardocarreras:
The advantages here:
- All the changes from the #294 and #257 are together
- No extra merging required
- Tested additionally with Fedora 40 (gcc 14.1.1 20240507 and clang version 18.1.1), so we have building setup that will be compiling reasonably okay in the future

What is still missing:
- Issue with `pybind` and `clang` was left as in #293 
- Any comments left in the PRs may need to be transfered/solved here

m-mirz (approved):

dpsim-models/include/dpsim-models/DP/DP_Ph3_SynchronGeneratorDQ.h
```diff
@@ -73,8 +73,8 @@ class SynchronGeneratorDQ : public MNASimPowerComp<Complex>,
 
   // #### MNA Functions ####
   /// Initializes variables of component
-  virtual void mnaCompInitialize(Real omega, Real timeStep,
-                                 Attribute<Matrix>::Ptr) override = 0;
+  void mnaCompInitialize(Real omega, Real timeStep,
+                         Attribute<Matrix>::Ptr leftVector) override = 0;
```
n-eiling, line 77: why did you remove the virtual keyword?

n-eiling (approved):

stv0g (dismissed):
I've briefly reviwed the changes. I agree with the comment from Niklas. Other than that it looks good.

dpsim-models/include/dpsim-models/DP/DP_Ph3_SynchronGeneratorDQ.h
leonardocarreras: **TL;DR** I did due to a suggestion of sonar that said the `virtual` should be dropped because is redundant. Should I keep it?

Some reasoning follows, just to argue on why/what... According to the standard, apparently it is not needed [and from here](https://en.cppreference.com/w/cpp/language/virtual)

> If some member function vf is declared as virtual in a class Base, and some class Derived, which is derived, directly or indirectly, from Base, has a declaration for member function with the same
> 
> - name
> - parameter type list (but not the return type)
> - cv-qualifiers
> - ref-qualifiers
> Then this function in the class Derived is also virtual (whether or not the keyword virtual is used in its declaration) and overrides Base::vf (whether or not the word override is used in its declaration).


_Conditions:_ 
- this class `DP_Ph3_SynchronGeneratorDQ` inherits from `MNASimPowerComp`, and is parent to `DP_Ph3_SynchronGeneratorDQODE` and `DP_Ph3_SynchronGeneratorDQTrapez`
- `MNASimPowerComp.cpp` provides an empty implementation of the `virtual void mnaCompInitialize` 
https://github.com/sogno-platform/dpsim/blob/f07066ab9941dfcb35e4acaf7086168f2b3462fb/dpsim-models/src/MNASimPowerComp.cpp#L120-L124
- and is not marked as pure virtual in `MNASimPowerComp.h`  and only virtual (done by design to be optional, according to the help)
https://github.com/sogno-platform/dpsim/blob/f07066ab9941dfcb35e4acaf7086168f2b3462fb/dpsim-models/include/dpsim-models/MNASimPowerComp.h#L75-L76

_Intention of the code:_
-  if we keep the implementation optional (then the childs will still inherit from `MNASimPowerComp`: consecuence is then that lines 76-77 need to be deleted (because we are not overriding neither implementing here) but we would be able to get instances of `DP_Ph3_SynchronGeneratorDQ` 
- if we force the implementation in the subclass (the intention seems more logical to me is that we only want to allow subclasses): the virtual keyword is not needed (according to sonar) because is inherited from the functon in `MNASimPowerComp`, and then we avoid the instantiation of this current class and forcing the implementation of it in the childs by marking it as `pure virtual` with the override

leonardocarreras: Update: I do not have a strong opinion to be in favour of sonar, I changed it.

leonardocarreras:
I adopted the proposed change, this should be good to go after passing the pipeline.

m-mirz (approved):

n-eiling (approved):

stv0g (approved):

PR #295, opened 2024-05-17
editorconfig: set C++ indentation size to two

n-eiling:
At some point the indentation size changed to two, but the editorconfig still  shows four. This fixes that.
We also seem to now use spaces rather than tabs.


stv0g (approved):

stv0g:
This makes more sense. LLVM code style uses two spaces for indentation.

@m-mirz This should be a no-brainer to merge :D

PR #296, opened 2024-05-17
Add new action using the build wrapper from sonar directly

leonardocarreras:
Use the `sonarsource/sonarcloud-github-c-cpp@v2`, with this we obtain less complexity in the action.

Additionally:

- Fixed deprecation in properties, now the cache is managed in a different way and some properties are different. The build process is also cached to improve the speed.
- Threading is managed automatically, now we get 4 threads and we were running only with 2 (speed should improve)


stv0g (approved):
LGTM

Issue #288, opened 2024-05-06
Reuse conductance stamp code


georgii-tishenin:
**What is the issue?**
Logic for stamping of conductance is the same (or very similar) for all EMT components that are represented by Norton Equivalent or conductance, however each component re-implements this logic. Because of that, `mnaCompApplySystemMatrixStamp()`:

-  increases a chance of copy-paste errors and complexity in implementing new components,

- requires to read more code, once you're reviewing/trying to understand the component.

**What is the possible solution?**

- Create static ("utility") functions that implement the stamping logic and put them to `MathUtils.`

- Reuse the utility functions in component classes to stamp conductance into system matrix.

To provide a better idea of what I mean by that, I implemented this in a branch: [reuse-conductance-stamp-code](https://github.com/sogno-platform/dpsim/tree/reuse-conductance-stamp-code).

**Describe alternatives you've considered**
An alternative solution could be to put the implementation of stamping in a base class. However, not all of the components are represented by Norton Equivalent or conductance.

I believe stamping conductance into matrix fits well into  `MathUtils`. That's why I suggest it as a first option.

**Additional context**

- In [reuse-conductance-stamp-code](https://github.com/sogno-platform/dpsim/tree/reuse-conductance-stamp-code) I am reusing the stamping code just for R, L, C components, however this can be extended to all components represented by Norton Equivalent or conductance, for example `EMT_Ph3_GeneratorVBR`.

- Similar utility functions can be added to reuse the code for stamping admittance in DP and SP domains.

- The logic can be extended to components that have more than 2 terminals (e.g., three-winding transformer).

- I did not check that, but I believe we can also reuse code for stamping impedance of components represented by Thevenin Equivalent.

georgii-tishenin:
Hi Team,
@m-mirz, @dinkelbachjan, @stv0g, @martinmoraga, @gnakti, @leonardocarreras, @pipeacosta, @MarvinTollnitschRWTH  let me know what you think.

m-mirz:
I think it is a good idea to remove this redundant code. Just two remarks:
Your new implementation seems to test the terminal grounded condition more often for the three phase case. I would try to avoid this since this function is called very often… better write a little bit more code.
I am wondering if we don’t have another place to put this code since it is already a little bit more specific than just „math“. Are there similar redundant pieces of code that we could pull out? You already suggested thevenin like stamps. Maybe even a new file is appropriate here with a more specific name.

martinmoraga:
great idea!

[...]

PR #297, opened 2024-05-18
Reuse code for MNA matrix stamp operations

georgii-tishenin:
### Summary
This PR introduces the utility class `MNAStampUtils` for MNA matrix stamping operations. It standardizes and reuses the stamping logic for 1-phase and 3-phase EMT, DP and SP R, L, C components. The motivation is to make the codebase more maintainable and reduce redundancy.

### Additional Context
- **Issue Reference**: These changes were discussed in issue [#288](https://github.com/sogno-platform/dpsim/issues/288) - Reuse conductance stamp code.
- **Future Plans**: There are plans for future PRs to extend this approach to other components deriving from the `MNASimPowerComp` class. This might involve adding more variations of stamping functions to `MNAStampUtils` to accommodate different parameter lists or optimized logic for specific cases, such as components with a single terminal or symmetric phase impedances.


dpsim-models/src/SP/SP_Ph1_Resistor.cpp
```diff
@@ -107,44 +107,9 @@ void SP::Ph1::Resistor::mnaCompApplySystemMatrixStamp(
...
-    if (terminalNotGrounded(0) && terminalNotGrounded(1)) {
-      SPDLOG_LOGGER_INFO(mSLog, "Add {:s} to system at ({:d},{:d})",
-                         Logger::complexToString(-conductance),
-                         matrixNodeIndex(0), matrixNodeIndex(1));
-      SPDLOG_LOGGER_INFO(mSLog, "Add {:s} to system at ({:d},{:d})",
-                         Logger::complexToString(-conductance),
-                         matrixNodeIndex(1), matrixNodeIndex(0));
-    }
-  }
+  MNAStampUtils::stampAdmittance(conductance, systemMatrix, matrixNodeIndex(0),
```
georgii-tishenin, line 110: I assumed here that iterating through mNumFreqs was a copy-paste error, because it is not done for other SP components.
@martinmoraga, @m-mirz

m-mirz: It could be... I don't remember testing the multi frequency approach for SP.

m-mirz (dismissed):

dpsim-models/include/dpsim-models/MNASimPowerComp.h
```diff
@@ -4,6 +4,7 @@
 
 #include <dpsim-models/SimPowerComp.h>
 #include <dpsim-models/Solver/MNAInterface.h>
+#include <dpsim-models/MNAStampUtils.h>
```
leonardocarreras, line 7: Did you use `git-clang-format`?

dpsim-models/src/DP/DP_Ph3_Resistor.cpp
```diff
@@ -56,147 +56,12 @@ void DP::Ph3::Resistor::mnaCompInitialize(Real omega, Real timeStep,
 
 void DP::Ph3::Resistor::mnaCompApplySystemMatrixStamp(
     SparseMatrixRow &systemMatrix) {
+  MatrixComp conductance = Matrix::Zero(3, 3);
```
leonardocarreras, line 59: Maybe think of using the template here
https://github.com/sogno-platform/dpsim/blob/f71d1a3e94d5e0ca2773acc5f4f0c53dac4ce9c8/dpsim-models/include/dpsim-models/Definitions.h#L103-L105

dpsim-models/src/MNAStampUtils.cpp
leonardocarreras, line 1: The functions in this file are not using the logger, it might be good to think if `SPDLOG_LOGGER_DEBUG` could be an interesting option

georgii-tishenin: Hey Leo, thanks for your comment! Yes, I left out logging, because stamp logging was implemented only for some of the components and also there's no logging in `MathUtils.`
But I think it's a good idea.
Since `MNAStampUtils `is a static class, would you then pass the `Logger::Log` object from a component, that calls a logging function? Or how would you do it?

dpsim-models/include/dpsim-models/MNASimPowerComp.h
georgii-tishenin: Ooops, I did not apply it to this file😅. How did you figure out? By the non-alphabetic order? 😀

leonardocarreras: Yes :grin: but is not a problem, is just trying to take advantage that we touch the file so we make it pretty

leonardocarreras (approved):
Now looks good to me :)

m-mirz (approved):

PR #298, opened 2024-05-18
Update actions in workflow and improve parallelisation and cache handling

leonardocarreras:
Changelog
===

In general, the versions of the actions were updated for checkout, cache, download and upload artifacts. Aditionally, the parallel build with a static value in an environment variable `-j2` was changed to the `--parallel $(nproc)` argument to the command to pick automatically the value.

In particular, other changes were done:
- `build_test_linux_fedora_minimal.yaml`
  - the build cache is updated with a new name to preserve for the same reference (branch, push) and not for the same commit
- `build_test_linux_fedora.yaml`
  - setup a build cache for dpsimpy & examples
    - archive in a different cache (not overwriting the others) per commit sha
  - changes to the archiving of build of cxx examples
    - input from the previous stage and recompile and save for the current compilation
  - jupyter notebook tests were parallelized 
    - the package pytest-xdist is installed
    - cache is changed to separate per commit sha
  - comparison of results
    - previous cached result is used
  - test of binaries
    - the output of the cxx compilation is used as cache
  - villas examples
    - the needs is changed to examples (dpsimpy probably worked due old overwrites in the cache, when chaining the compilation outputs it started failing)
- `build_test_linux_rocky.yaml`
  - added cache to the clang compilation
  - added cache to the profiling compilation
- `build_test_windows.yaml`
  - added cache to the build directory per reference
- `documentation-fein.yaml`
  - changed node to v20
  - updated hugo and gh-pages actions
- `run_and_profile_example.yaml`
  - changed the container to sogno/dpsim.dev-rocky (we were still on centos)
- `run_villas_example.yaml`
  - changed the cache to the cxx examples build

### DIFF
--- a/.github/workflows/build_test_linux_fedora.yaml
+++ b/.github/workflows/build_test_linux_fedora.yaml
@@ -118,7 +118,7 @@ jobs:
       uses: actions/upload-artifact@v4
       with:
         name: pytest-output-${{ github.sha }}
-        path: outputs/examples/Notebooks/
+        path: outputs/
 
   compare-notebook-results:
     name: Compare Notebook results
--- a/.github/workflows/build_test_linux_fedora.yaml
+++ b/.github/workflows/build_test_linux_fedora.yaml
@@ -7,17 +7,23 @@ on:
 
 jobs:
   linux-fedora-dpsimpy:
-    name: Build dpsimpy on Fedora Linux
+    name: Build dpsimpy and tests examples on Fedora Linux
     runs-on: ubuntu-latest
     container: sogno/dpsim:dev
     steps:
     - name: Checkout
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       with:
         submodules: recursive
 
     - name: Create Build Environment
       run: mkdir build
+
+    - name: Setup build directory cache
+      uses: actions/cache@v4
+      with:
+        path: ${{ github.workspace }}/build
+        key: build-fedora-dir-cache-${{ github.ref }}
 
     - name: Configure CMake
       shell: bash
@@ -27,15 +33,13 @@ jobs:
     - name: Build dpsimpy and test examples
       shell: bash
       working-directory: ${{ github.workspace }}/build
-      run: cmake --build . --target dpsimpy --target tests --target dpsimpyvillas
-      env:
-        MAKEFLAGS: "-j2"
+      run: cmake --build . --target dpsimpy --target tests --target dpsimpyvillas --parallel $(nproc)
 
     - name: Archive build directory
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
         path: ${{ github.workspace }}/build
-        name: build-cache
+        name: build-fedora-examples-cache-${{ github.sha }}
         retention-days: 1
 
   linux-fedora-examples:
@@ -45,28 +49,26 @@ jobs:
     container: sogno/dpsim:dev
     steps:
     - name: Checkout
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       with:
         submodules: recursive
 
     - name: Restore build archive
-      uses: actions/download-artifact@v3
+      uses: actions/download-artifact@v4
       with:
-       name: build-cache
+       name: build-fedora-examples-cache-${{ github.sha }}
        path: ${{ github.workspace }}/build
 
     - name: Build every target
       shell: bash
       working-directory: ${{ github.workspace }}/build
-      run: cmake --build .
-      env:
-        MAKEFLAGS: "-j2"
+      run: cmake --build . --parallel $(nproc)
 
     - name: Archive build directory
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
         path: ${{ github.workspace }}/build
-        name: build-cache
+        name: build-cache-examples-cpp-${{ github.sha }}
         retention-days: 1
 
 ## Tests ##
@@ -77,12 +79,12 @@ jobs:
     container: sogno/dpsim:dev
     steps:
     - name: Checkout
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
 
     - name: Restore build archive
-      uses: actions/download-artifact@v3
+      uses: actions/download-artifact@v4
       with:
-       name: build-cache
+       name: build-fedora-examples-cache-${{ github.sha }}
        path: ${{ github.workspace }}/build
 
     - name: Prepare binary permissions
@@ -99,6 +101,10 @@ jobs:
       working-directory: ${{ github.workspace }}
       run: ls -l ./build/dpsim/examples/cxx/
 
+    - name: Install parallel testing
+      shell: bash
+      run: pip install pytest-xdist
+
     - name: Run pytest
       shell: bash
       working-directory: ${{ github.workspace }}
@@ -106,12 +112,12 @@ jobs:
         PYTHONPATH: "${{ github.workspace }}/build"
       run: |
         cp -r python/src/dpsim build/
-        pytest -v examples/Notebooks
+        pytest -v examples/Notebooks -n auto
 
     - name: Archive notebook outputs
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
-        name: pytest-output
+        name: pytest-output-${{ github.sha }}
         path: outputs/examples/Notebooks/
 
   compare-notebook-results:
@@ -120,13 +126,13 @@ jobs:
     needs: [test-jupyter-notebooks]
     steps:
     - name: Download new notebook results
-      uses: actions/download-artifact@v3
+      uses: actions/download-artifact@v4
       with:
-       name: pytest-output
+       name: pytest-output-${{ github.sha }}
        path: ${{ github.workspace }}/notebooks-new
 
     - name: Download master notebook results
-      uses: actions/cache@v3
+      uses: actions/cache@v4
       with:
         path: ${{ github.workspace }}/notebooks-master
         key: notebook-output-cache-master-${{ github.sha }}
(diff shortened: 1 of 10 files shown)


.github/workflows/build_test_linux_fedora.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
     - name: Build dpsimpy and test examples
       shell: bash
       working-directory: ${{ github.workspace }}/build
-      run: cmake --build . --target dpsimpy --target tests --target dpsimpyvillas -j`nproc`
+      run: cmake --build . --target dpsimpy --target tests --target dpsimpyvillas -j$(cat /proc/cpuinfo|grep -w processor|wc -l)
```
stv0g, line 30: Hey @leonardocarreras,

could you briefly describe the motivation behind using `/proc/cpuinfo` vs `nproc` here?

I believe, if we simply pass -j, we should already use maximum parallelism. Because of:

CMake ([docs(https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-build-j)):
> If <jobs> is omitted the native build tool's default number is used.

GNU Make:
> If the -j option is given without an argument, make will not limit the number of jobs that can run simultaneously.

Reference: https://www.flamingspork.com/blog/2020/11/25/why-you-should-use-nproc-and-not-grep-proc-cpuinfo/

A little side note: I would prefer it if we always use long options (`--parallel` vs `-j` in scripts as its more expressive. Short options are great if you use them on the CLI. In scripts we want the code to be self-documenting as much as possible. Hence using a long name is easier to understand and review..

leonardocarreras: Thanks @stv0g for having a look!

This is WIP, I just had some errors in some runs and I was trying options. The errors were due to other factors.

I will take then the `--parallel` option to make it easier to read.

stv0g (dismissed):
LGTM :)

.github/workflows/build_test_linux_fedora.yaml
```diff
@@ -27,15 +33,13 @@ jobs:
     - name: Build dpsimpy and test examples
       shell: bash
       working-directory: ${{ github.workspace }}/build
-      run: cmake --build . --target dpsimpy --target tests --target dpsimpyvillas
-      env:
-        MAKEFLAGS: "-j2"
+      run: cmake --build . --target dpsimpy --target tests --target dpsimpyvillas --parallel $(nproc)
```
n-eiling, line 36: I'm not sure if this is still the case but it used to be that `nproc` within a container would report the cores of the host rather than the cores assigned to the container. The runner is executed in a container, so `nproc` might not report the correct number.
Maybe we should parse `/sys/fs/cgroup/cpuset/cpuset.cpus` instead? Or set a fixed number of threads.

Now that I think about it, which runner do we actually use? The default github actions runner? Then this might not be an issue.

.github/workflows/build_test_linux_fedora.yaml
```diff
@@ -99,19 +101,23 @@ jobs:
...
+
     - name: Run pytest
       shell: bash
       working-directory: ${{ github.workspace }}
       env:
         PYTHONPATH: "${{ github.workspace }}/build"
       run: |
         cp -r python/src/dpsim build/
-        pytest -v examples/Notebooks
+        pytest -v examples/Notebooks -n auto
```
n-eiling, line 115: same as above. Does this create the correct number of threads?

.github/workflows/build_test_linux_fedora.yaml
```diff
@@ -194,15 +200,15 @@ jobs:
...
+      uses: actions/checkout@v4
 
     - name: Run cppcheck
       working-directory: ${{ github.workspace }}
       shell: bash
       id: cppcheck
       run: |
         set -o pipefail
-        cppcheck --max-configs=32 -j 32 --inline-suppr --error-exitcode=1 -q --enable=warning,performance,portability,information,missingInclude --std=c++17 -I dpsim/include/ -I dpsim-models/include/ -I dpsim-villas/include dpsim/src/ dpsim-models/src/ dpsim-villas/src/ 2>&1 | tee cppcheck-output.log
+        cppcheck --max-configs=32 -j $(nproc) --inline-suppr --error-exitcode=1 -q --enable=warning,performance,portability,information,missingInclude --std=c++17 -I dpsim/include/ -I dpsim-models/include/ -I dpsim-villas/include dpsim/src/ dpsim-models/src/ dpsim-villas/src/ 2>&1 | tee cppcheck-output.log
```
n-eiling, line 211: same as above

.github/workflows/build_test_linux_fedora_minimal.yaml
```diff
@@ -27,14 +33,7 @@ jobs:
     - name: Build every target
       shell: bash
       working-directory: ${{ github.workspace }}/build
-      run: cmake --build .
-      env:
-        MAKEFLAGS: "-j2"
+      run: cmake --build . --parallel $(nproc)
```
n-eiling, line 36: same

n-eiling (changes requested):
I'm not sure we want caching for the profiling actions, because the point of this action was to have performance figures for each commit. However, we don't really use them, so I would also be fine with removing them.

.github/workflows/build_test_linux_fedora.yaml
leonardocarreras: Hi, we are using the github actions runner... For this case seems to give the correct data

.github/workflows/build_test_linux_fedora.yaml
leonardocarreras: For this one, we are getting a lower number (it starts 2 not 4 processes)

.github/workflows/container.yaml
```diff
@@ -68,17 +68,17 @@ jobs:
...
-        uses: docker/login-action@v2
+        uses: docker/login-action@v3
         with:
           username: ${{ secrets.DOCKERHUB_USER }}
           password: ${{ secrets.DOCKERHUB_SECRET }}
 
       - name: Build and push
         id: docker_build_dev
-        uses: docker/build-push-action@v3
+        uses: docker/build-push-action@v5
```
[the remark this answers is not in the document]
leonardocarreras: The reason to change v3 (Jan 2023) to v5 (current) is mainly bumping to newer Node 20 in all actions, as Node 16 actions will get phased out (currently deprecated). This change was integrated in v5.3.0
https://github.com/docker/build-push-action/commit/0c20fff10d747c60a2ea9bbfd1c67d4810ffa53a

.github/workflows/container.yaml
```diff
@@ -88,17 +88,17 @@ jobs:
...
```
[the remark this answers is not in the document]
leonardocarreras: > The reason to change v3 (Jan 2023) to v5 (current) is mainly bumping to newer Node 20 in all actions, as Node 16 actions will get phased out (currently deprecated). This change was integrated in v5.3.0 [docker/build-push-action@0c20fff](https://github.com/docker/build-push-action/commit/0c20fff10d747c60a2ea9bbfd1c67d4810ffa53a)

.github/workflows/build_test_linux_fedora.yaml
leonardocarreras: Another option would be to use this action [here](https://github.com/SimenB/github-actions-cpu-cores)

```yaml
    steps:
      - name: Get number of CPU cores
        uses: SimenB/github-actions-cpu-cores@v2
        id: cpu-cores
      - name: run tests
        run: npx jest --max-workers ${{ steps.cpu-cores.outputs.count }}
```

leonardocarreras:
> I'm not sure we want caching for the profiling actions, because the point of this action was to have performance figures for each commit. However, we don't really use them, so I would also be fine with removing them.

I think if we do not plan to use them it would be okay to delete the cache or the action itself. I listen to opinions.

.github/workflows/build_test_linux_fedora.yaml
leonardocarreras: Well, in theory it should do it automatically, but there are mixed opinions about the actual performance across the internet for the case of a github windows runner.
We could use something like `NumberOfLogicalProcessors` for consistency, because `nproc` is in theory not implemented in `pwsh` or use the action `SimenB/github-actions-cpu-cores@v2`

.github/workflows/build_test_windows.yaml
```diff
@@ -11,13 +11,23 @@ jobs:
...
+        key: build-windows-cache-${{ github.ref }}
+
+    - name: Build DPsim
+      run: |
+        cd ${{ github.workspace }}
         cd build
         git describe --tags --abbrev=0 --match "v*"
         cmake -DWITH_PYBIND=OFF ..
-        cmake --build . --target dpsim
+        cmake --build . --target dpsim --parallel
```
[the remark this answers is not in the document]
leonardocarreras: Well, in theory it should do it automatically, but there are mixed opinions about the actual performance across the internet for the case of a github windows runner.
We could use something like `NumberOfLogicalProcessors` for consistency, because `nproc` is in theory not implemented in `pwsh` or use the action `SimenB/github-actions-cpu-cores@v2`

n-eiling:
OK thanks for the additional research. Looks like our runners get 4 CPUs according to the github documentation.
I don't need the profiling anymore, so I would be fine with removing them to make the CI run faster.

n-eiling (dismissed):

leonardocarreras:
There is now a villas problem that needs some investigation

leonardocarreras:
Hi @n-eiling! I see that you are working on #299 and the updated container with a new villas version, thanks for this... Do you think that it will be fixing the new errors that we get? In your opinion could this go first? Should I retrigger the workflow with the older container to make the workflow to pass again?

n-eiling:
This is due to an error in VILLAS. I already fixed it. Once it's merged the container should work again. Sorry for breaking this.

leonardocarreras:
@n-eiling thank you for fixing, maybe with luck the merge does not take long :)

m-mirz:
@leonardocarreras instead of removing the profiling we could also trigger this manually. In general, I think it is valuable to have the profiling in such a convenient way.

leonardocarreras:
@m-mirz  Now it works as suggested (on-demand)

m-mirz (approved):

PR #299, opened 2024-06-06
Villas Interface: Improve FpgaExample

n-eiling:
This makes the FpgaExample a closed loop test of the communication via VILLAS.
I also added the signal names configured in the example to the actual signals and added a bit more debug output.


n-eiling:
I don't get why the two tests fail. It seems like VILLAS segfaults. However, my changes did not touch anything related to this and I cannot reproduce the errors outside the CI.
@stv0g do you have an idea how to debug this, or what I might have done wrong?

n-eiling:
There was some issue with the new VILLAS version. It seems to be fixed now, so this is ready to be merged.

m-mirz (approved):

PR #316, opened 2024-07-29
New queueless VILLAS interface / improve real-time performance

n-eiling:
villas interface: improve real-time capability

The original villas interface was not real-time capable because of the queue it uses. This commit creates an interface for villas interfaces, moves the old implementation to InterfaceQueued and adds a new queueless, threadless implementation InterfaceQueueless.

models: Add DC voltage source

The DP_Ph1_VoltageSource always calculates a sinusoidal signal, even if we only want to set a DC value. This commit adds a DC generator for the model so it can be significantly faster if we do not need the sinusoidal signal.

 FpgaExample: use new villas interface and make logging optional.

### DIFF
--- a/dpsim-models/include/dpsim-models/Attribute.h
+++ b/dpsim-models/include/dpsim-models/Attribute.h
@@ -353,6 +353,7 @@ public:
   AttributeBase::Ptr cloneValueOntoNewAttribute() override {
     return AttributePointer<AttributeBase>(
         AttributeStatic<T>::make(this->get()));
+    //TODO: This is in the real time path. We should not use heap here.
   };
 
   /**
--- a/dpsim-models/include/dpsim-models/Attribute.h
+++ b/dpsim-models/include/dpsim-models/Attribute.h
@@ -353,7 +353,7 @@ public:
   AttributeBase::Ptr cloneValueOntoNewAttribute() override {
     return AttributePointer<AttributeBase>(
         AttributeStatic<T>::make(this->get()));
-    //TODO: This is in the real time path. We should not use heap here.
+    // TODO: This is in the real time path. We should not use heap here.
   };
 
   /**
--- a/dpsim-models/include/dpsim-models/DP/DP_Ph1_VoltageSource.h
+++ b/dpsim-models/include/dpsim-models/DP/DP_Ph1_VoltageSource.h
@@ -10,6 +10,7 @@
 
 #include <dpsim-models/MNASimPowerComp.h>
 #include <dpsim-models/Signal/CosineFMGenerator.h>
+#include <dpsim-models/Signal/DCGenerator.h>
 #include <dpsim-models/Signal/FrequencyRampGenerator.h>
 #include <dpsim-models/Signal/SignalGenerator.h>
 #include <dpsim-models/Signal/SineWaveGenerator.h>
--- a/dpsim-models/include/dpsim-models/DP/DP_Ph1_VoltageSource.h
+++ b/dpsim-models/include/dpsim-models/DP/DP_Ph1_VoltageSource.h
@@ -80,6 +80,10 @@ public:
   void setParameters(Complex initialPhasor, Real modulationFrequency,
                      Real modulationAmplitude, Real baseFrequency = 0.0,
                      bool zigzag = false);
+  /// Setter for reference voltage with a real valued generator
+  /// This will initialize the values of mVoltageRef to match the given parameters
+  /// However, the attributes can be modified during the simulation to dynamically change the value of the output voltage.
+  void setParameters(Real voltageRef);
 
   // #### MNA Section ####
   /// Initializes internal variables of the component
--- a/dpsim-models/src/DP/DP_Ph1_VoltageSource.cpp
+++ b/dpsim-models/src/DP/DP_Ph1_VoltageSource.cpp
@@ -31,12 +31,20 @@ SimPowerComp<Complex>::Ptr DP::Ph1::VoltageSource::clone(String name) {
 }
 
 void DP::Ph1::VoltageSource::setParameters(Complex voltageRef, Real srcFreq) {
-  auto srcSigSine = Signal::SineWaveGenerator::make(**mName + "_sw");
-  srcSigSine->mVoltageRef->setReference(mVoltageRef);
-  srcSigSine->mFreq->setReference(mSrcFreq);
-  srcSigSine->setParameters(voltageRef, srcFreq);
+  if (srcFreq == 0.0) {
+    auto srcDC = Signal::DCGenerator::make(**mName + "_dc");
+    srcDC->mVoltageRef->setReference(mVoltageRef);
+    srcDC->setParameters(voltageRef.real());
 
-  mSrcSig = srcSigSine;
+    mSrcSig = srcDC;
+  } else {
+    auto srcSigSine = Signal::SineWaveGenerator::make(**mName + "_sw");
+    srcSigSine->mVoltageRef->setReference(mVoltageRef);
+    srcSigSine->mFreq->setReference(mSrcFreq);
+    srcSigSine->setParameters(voltageRef, srcFreq);
+
+    mSrcSig = srcSigSine;
+  }
   mParametersSet = true;
 }
 
--- a/dpsim-models/src/DP/DP_Ph1_VoltageSource.cpp
+++ b/dpsim-models/src/DP/DP_Ph1_VoltageSource.cpp
@@ -30,21 +30,23 @@ SimPowerComp<Complex>::Ptr DP::Ph1::VoltageSource::clone(String name) {
   return copy;
 }
 
+void DP::Ph1::VoltageSource::setParameters(Real voltageRef) {
+  auto srcDC = Signal::DCGenerator::make(**mName + "_dc");
+  srcDC->mVoltageRef->setReference(mVoltageRef);
+  srcDC->setParameters(voltageRef);
+
+  mSrcSig = srcDC;
+  mParametersSet = true;
+}
+
 void DP::Ph1::VoltageSource::setParameters(Complex voltageRef, Real srcFreq) {
-  if (srcFreq == 0.0) {
-    auto srcDC = Signal::DCGenerator::make(**mName + "_dc");
-    srcDC->mVoltageRef->setReference(mVoltageRef);
-    srcDC->setParameters(voltageRef.real());
+  auto srcSigSine = Signal::SineWaveGenerator::make(**mName + "_sw");
+  srcSigSine->mVoltageRef->setReference(mVoltageRef);
+  srcSigSine->mFreq->setReference(mSrcFreq);
+  srcSigSine->setParameters(voltageRef, srcFreq);
 
-    mSrcSig = srcDC;
-  } else {
-    auto srcSigSine = Signal::SineWaveGenerator::make(**mName + "_sw");
-    srcSigSine->mVoltageRef->setReference(mVoltageRef);
-    srcSigSine->mFreq->setReference(mSrcFreq);
-    srcSigSine->setParameters(voltageRef, srcFreq);
+  mSrcSig = srcSigSine;
 
-    mSrcSig = srcSigSine;
-  }
   mParametersSet = true;
 }
 
--- a/dpsim-models/src/Signal/DCGenerator.cpp
+++ b/dpsim-models/src/Signal/DCGenerator.cpp
@@ -1,0 +1,25 @@
+/* Copyright 2024 Institute for Automation of Complex Power Systems,
+ *                EONERC, RWTH Aachen University
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *********************************************************************************/
+
+#include <dpsim-models/Signal/DCGenerator.h>
+
+using namespace CPS;
+
+Signal::DCGenerator::DCGenerator(String name, Logger::Level logLevel)
+    : SignalGenerator(name, logLevel),
+      mVoltageRef(mAttributes->createDynamic<Complex>("V_ref")) {
+  **mFreq = 0.0;
+  SPDLOG_LOGGER_INFO(mSLog, "Create {} {}", type(), name);
+}
+
+void Signal::DCGenerator::setParameters(Real initialReference) {
+  **mVoltageRef = initialReference;
+  **mSigOut = initialReference;
+}
+
+void Signal::DCGenerator::step(Real time) { **mSigOut = **mVoltageRef; }
--- a/dpsim-models/src/Signal/DCGenerator.cpp
+++ b/dpsim-models/src/Signal/DCGenerator.cpp
@@ -1,10 +1,7 @@
-/* Copyright 2024 Institute for Automation of Complex Power Systems,
- *                EONERC, RWTH Aachen University
- *
- * This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at https://mozilla.org/MPL/2.0/.
- *********************************************************************************/
+/* Author: Niklas Eiling <niklas.eiling@eonerc.rwth-aachen.de>
+ * SPDX-FileCopyrightText: 2023-2024 Niklas Eiling <niklas.eiling@eonerc.rwth-aachen.de>
+ * SPDX-License-Identifier: MPL-2.0
+ */
 
 #include <dpsim-models/Signal/DCGenerator.h>
 
--- a/dpsim-villas/examples/cxx/FpgaExample.cpp
+++ b/dpsim-villas/examples/cxx/FpgaExample.cpp
@@ -9,9 +9,11 @@
 #include <fstream>
 
 #include <DPsim.h>
+#include <dpsim-models/Attribute.h>
 #include <dpsim-models/DP/DP_Ph1_CurrentSource.h>
 #include <dpsim-models/SimNode.h>
 #include <dpsim-villas/InterfaceVillas.h>
+#include <dpsim-villas/InterfaceVillasQueueless.h>
 #include <dpsim/Utils.h>
 
 using namespace DPsim;
@@ -21,11 +23,17 @@ using namespace CPS::DP::Ph1;
 const std::string buildFpgaConfig(CommandLineArgs &args) {
   std::filesystem::path fpgaIpPath =
       "/usr/local/etc/villas/node/etc/fpga/vc707-xbar-pcie-dino/"
-      "vc707-xbar-pcie-dino.json";
+      "vc707-xbar-pcie-dino-v2.json";
 
   if (args.options.find("ips") != args.options.end()) {
     fpgaIpPath = std::filesystem::path(args.getOptionString("ips"));
   }
+  std::string loopbackConfig = fmt::format(
+      R"STRING(
+      "queuelen": 1024,
+      "samplelen": 1024,
+      "mode": "polling"
+    )STRING");
   std::string cardConfig = fmt::format(
       R"STRING("card": {{
       "interface": "pcie",
@@ -48,7 +56,8 @@ const std::string buildFpgaConfig(CommandLineArgs &args) {
       "signals": [{{
         "name": "from_dpsim",
         "type": "complex",
-        "unit": "V"
+        "unit": "V",
+        "builtin": false
       }}],
       "hooks": [{{
         "type": "dp",
@@ -63,12 +72,18 @@ const std::string buildFpgaConfig(CommandLineArgs &args) {
   std::string signalInConfig = fmt::format(
       R"STRING("in": {{
       "signals": [{{
+        "name": "seqnum",
+        "type": "integer",
+        "unit": "",
+        "builtin": false
+      }},
+      {{
         "name": "to_dpsim",
         "type": "float",
         "unit": "V",
         "builtin": false
       }}],
-      "hooks": ["print", {{
+      "hooks": [{{
         "type": "dp",
         "signal": "to_dpsim",
         "f0": {},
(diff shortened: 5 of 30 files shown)


m-mirz:
@n-eiling this reduces the latency by removing the intermediate queue, hoping that the sync call to villas returns very quickly and does not block dpsim for too long?
Do I remember correctly that villas also has internal queues for some node types to avoid blocking?

It seems like a good idea if you do not only care about real-time in the simulation but also across the interface.

stv0g:
> Do I remember correctly that villas also has internal queues for some node types to avoid blocking?

Yes there are some, but not for the reason you are thinking. Some nodes have internal queues as the underlying libraries have callback based APIs or work in a background thread which require a queue to exchange samples with the main VILLASnode thread.

> @n-eiling this reduces the latency by removing the intermediate queue, hoping that the sync call to villas returns very quickly and does not block dpsim for too long?

I agree here. I think both use cases are valid. But most people probably will use DPsim with the non hard real-time VILLASndoe node-types which use syscalls / network I/O anyway. In that cases, I would still recommend to use the queue-based interface

@n-eiling Does your new interface also support non-blocking operation?

dpsim-models/include/dpsim-models/Attribute.h
```diff
@@ -353,6 +353,7 @@ class Attribute : public AttributeBase,
   AttributeBase::Ptr cloneValueOntoNewAttribute() override {
     return AttributePointer<AttributeBase>(
         AttributeStatic<T>::make(this->get()));
+    //TODO: This is in the real time path. We should not use heap here.
```
stv0g, line 356:
```suggestion
    // TODO: This is in the real time path. We should not use heap here.
```

dpsim-models/include/dpsim-models/DP/DP_Ph1_VoltageSource.h
```diff
@@ -10,6 +10,7 @@
 
 #include <dpsim-models/MNASimPowerComp.h>
 #include <dpsim-models/Signal/CosineFMGenerator.h>
+#include <dpsim-models/Signal/DCGenerator.h>
```
stv0g, line 13: I find the name "DCGenerator" pretty confusing. Isnt it just a constant value? Can we call it "Constant"?

@m-mirz Whats your opinion on this?

dpsim-villas/examples/cxx/FpgaExample.cpp
```diff
@@ -21,11 +23,17 @@ using namespace CPS::DP::Ph1;
 const std::string buildFpgaConfig(CommandLineArgs &args) {
   std::filesystem::path fpgaIpPath =
       "/usr/local/etc/villas/node/etc/fpga/vc707-xbar-pcie-dino/"
-      "vc707-xbar-pcie-dino.json";
+      "vc707-xbar-pcie-dino-v2.json";
 
   if (args.options.find("ips") != args.options.end()) {
     fpgaIpPath = std::filesystem::path(args.getOptionString("ips"));
   }
+  std::string loopbackConfig = fmt::format(
```
stv0g, line 31: DPsim already uses [nlohmann/json](https://github.com/nlohmann/json).

I think it would nicer here to use the JSON library to construct the VILLASnode config.

dpsim-villas/examples/cxx/FpgaExample.cpp
```diff
@@ -174,17 +190,21 @@ int main(int argc, char *argv[]) {
...
 
-  auto sys = getTopology(args, intf, logger);
+  auto sys = getTopology(args, intf, nullptr);
 
-  RealTimeSimulation sim(args.name, args);
+  Simulation sim(args.name, args);
   sim.setSystem(sys);
   sim.addInterface(intf);
-  sim.addLogger(logger);
+  //sim.addLogger(logger);
```
stv0g, line 202: Please remove dead-code or make it configurable.

dpsim-villas/examples/cxx/FpgaExample.cpp
```diff
@@ -174,17 +190,21 @@ int main(int argc, char *argv[]) {
...
   sim.run();
 
+  CPS::Logger::get("FpgaExample")->info("Simulation finished.");
+  sim.logStepTimes("FpgaExample");
+
   //std::ofstream of("task_dependencies.svg");
```
stv0g, line 208: Please remove dead-code or make it configurable

dpsim-villas/include/dpsim-villas/InterfaceVillasQueueless.h
```diff
@@ -0,0 +1,104 @@
+// SPDX-License-Identifier: Apache-2.0
```
stv0g, line 1: What about the copyright here?

dpsim-villas/src/InterfaceVillasQueueless.cpp
```diff
@@ -0,0 +1,374 @@
+// SPDX-License-Identifier: Apache-2.0
```
stv0g, line 1: What about the copyright here?

dpsim-models/src/Signal/DCGenerator.cpp
```diff
@@ -0,0 +1,25 @@
+/* Copyright 2024 Institute for Automation of Complex Power Systems,
```
stv0g, line 1: Dont we want to have SPDX headers here as well?

@m-mirz

dpsim-villas/examples/cxx/FpgaExample.cpp
stv0g: You already do a similar check in line 165

dpsim-villas/src/InterfaceVillasQueueless.cpp
```diff
@@ -0,0 +1,374 @@
...
+#include <spdlog/spdlog.h>
+#include <typeinfo>
+#include <villas/node.h>
+
+using namespace CPS;
+using namespace villas;
+
+namespace DPsim {
+
+InterfaceVillasQueueless::InterfaceVillasQueueless(
```
stv0g, line 19: What about naming the interface directly about what they do?

E.g.

- VillasInterfaceBuffered
- VillasInterfaceQueued

"Queueless" is not descriptive here.

dpsim-villas/src/InterfaceVillasQueueless.cpp
```diff
@@ -0,0 +1,374 @@
...
+  }
+  ret = mNode->check();
+  if (ret < 0) {
+    SPDLOG_LOGGER_ERROR(
+        mLog,
+        "Error: Node in InterfaceVillas failed check. Check returned code {}",
+        ret);
+    std::exit(1);
+  }
+  struct villas::node::memory::Type *pool_mt = &villas::node::memory::heap;
```
stv0g, line 64: Please try do add some more empty lines. The code looks quite dense is hard to read in my opinion. Especially after closing brackets, I think an empty line would do better.

stv0g: You coding style is also not consistent in this regard. Sometimes you have empty lines after a `}`, and sometimes not.

dpsim-villas/src/InterfaceVillasQueueless.cpp
```diff
@@ -0,0 +1,374 @@
...
+                          "Error: Received Sample length ({}) does not match "
+                          "configured attributes length ({})",
+                          sample->length, mImportAttrsDpsim.size());
+      throw RuntimeError(
+          "Received Sample length does not match configured attributes length");
+    }
+
+    for (size_t i = 0; i < mImportAttrsDpsim.size(); i++) {
+      auto attr = std::get<0>(mImportAttrsDpsim[i]);
+      if (attr->getType() == typeid(Real)) {
```
stv0g, line 241: Why no `switch` / `case` here?

dpsim-villas/src/InterfaceVillasQueueless.cpp
```diff
@@ -0,0 +1,374 @@
...
+      SPDLOG_LOGGER_ERROR(mLog, "InterfaceVillas could not allocate a new "
+                                "sample! Not sending any data!");
+      return;
+    }
+
+    sample->signals = mNode->getOutputSignals(false);
+
+    for (size_t i = 0; i < mExportAttrsDpsim.size(); i++) {
+      auto attr = std::get<0>(mExportAttrsDpsim[i]);
+      if (attr->getType() == typeid(Real)) {
```
stv0g, line 298: `switch` / `case`

dpsim-villas/src/InterfaceVillasQueueless.cpp
```diff
@@ -0,0 +1,374 @@
...
+  } catch (const std::exception &) {
+    sample_decref(sample);
+
+    if (ret < 0)
+      SPDLOG_LOGGER_ERROR(
+          mLog,
+          "Failed to write samples to InterfaceVillas. Write returned code {}",
+          ret);
+
+    /* Don't throw here, because we managed to send something */
```
stv0g, line 346: Please use C++ style comments
```suggestion
    // Don't throw here, because we managed to send something.
```

dpsim-villas/src/InterfaceVillasQueueless.cpp
```diff
@@ -0,0 +1,374 @@
...
+  }
+}
+
+void InterfaceVillasQueueless::syncImports() {
+  //Block on read until all attributes with syncOnSimulationStart are read
+  mSequenceToDpsim = this->readFromVillas();
+}
+
+void InterfaceVillasQueueless::syncExports() {
+  //Just push all the attributes
```
stv0g, line 356:
```suggestion
  // Just push all the attributes.
```

dpsim-villas/src/InterfaceVillasQueueless.cpp
```diff
@@ -0,0 +1,374 @@
...
+          mLog,
+          "Failed to write samples to InterfaceVillas. Write returned code {}",
+          ret);
+
+    /* Don't throw here, because we managed to send something */
+  }
+}
+
+void InterfaceVillasQueueless::syncImports() {
+  //Block on read until all attributes with syncOnSimulationStart are read
```
stv0g, line 351:
```suggestion
  // Block on read until all attributes with syncOnSimulationStart are read.
```

dpsim-villas/src/InterfaceWorkerVillas.cpp
```diff
@@ -284,7 +288,7 @@ void InterfaceWorkerVillas::readValuesFromEnv(
 }
 
 void InterfaceWorkerVillas::writeValuesToEnv(
-    std::vector<Interface::AttributePacket> &updatedAttrs) {
+    std::vector<InterfaceQueued::AttributePacket> &updatedAttrs) {
   //Update export sequence IDs
```
stv0g, line 292:
```suggestion
  // Update export sequence IDs.
```

dpsim-villas/src/InterfaceVillasQueueless.cpp
```diff
@@ -0,0 +1,374 @@
...
+    idx++;
+  }
+}
+
+void InterfaceVillasQueueless::open() {
+  createNode();
+  createSignals();
+
+  mNode->getFactory()->start(
+      nullptr); //We have no SuperNode, so just hope type_start doesnt use it...
```
stv0g, line 142:
```suggestion
      nullptr); // We have no SuperNode, so just hope type_start doesn't use it.
```

stv0g (changes requested):
I did a first pass of review. But I think also @m-mirz or @dinkelbachjan should have a closer look.

n-eiling:
> @n-eiling this reduces the latency by removing the intermediate queue, hoping that the sync call to villas returns very quickly and does not block dpsim for too long? Do I remember correctly that villas also has internal queues for some node types to avoid blocking?
> 
> It seems like a good idea if you do not only care about real-time in the simulation but also across the interface.

The FpgaExample uses no queue, no VILLAS path and consequently is purely single-threaded. To achieve time steps in the microsecond range, even simple synchronization primitives create too much delay - even polling would create jitter.
In the example the simulation is hardware synchronized - so the time step is created on the FPGA, making it more accurate. This is achieved by disabling real-time in DPsim and having the VILLASnode read call block until the next time step starts. The FPGA also creates a sequence number that makes it possible for the VILLAS interface to detect time step overruns.
With some quite heavy OS optimizations (e.g., PREEMPT_RT, isolcpu, ...) and custom compiler flags for VILLAS and DPsim (`-flto -Ofast`), I was able to achieve time steps around 10 us.

dpsim-villas/examples/cxx/FpgaExample.cpp
n-eiling: This is there to stay consistent with the other examples.

dpsim-villas/include/dpsim-villas/InterfaceVillasQueueless.h
n-eiling: This is consistent with the other files

dpsim-models/src/Signal/DCGenerator.cpp
n-eiling: This is consistent with the other files

dpsim-villas/src/InterfaceVillasQueueless.cpp
n-eiling: It's not buffered though. I think something like "direct" is not descriptive either. My reasoning is that the main property of the interface is that it does not use a queue and therefore needs to be used with quite a bit more caution.

dpsim-villas/src/InterfaceVillasQueueless.cpp
n-eiling: Hm I use clang-format and editorconf. I think if its not specified in clang-format then it should be fine, no?

dpsim-villas/examples/cxx/FpgaExample.cpp
n-eiling: I actually use that library to create json objects. I don't think there is much difference if I use strings or initializer lists. This way makes it more similar to a json file that would be used with VILLASnode.

dpsim-villas/src/InterfaceVillasQueueless.cpp
n-eiling: what is the advantage?

dpsim/src/InterfaceQueued.cpp
```diff
@@ -0,0 +1,176 @@
...
+  //Just push all the attributes
+  this->pushDpsimAttrsToQueue();
+}
+
+void InterfaceQueued::popDpsimAttrsFromQueue(bool isSync) {
+  AttributePacket receivedPacket = {nullptr, 0, 0,
+                                    AttributePacketFlags::PACKET_NO_FLAGS};
+  UInt currentSequenceId = mNextSequenceInterfaceToDpsim;
+
+  //Wait for and dequeue all attributes that read should block on
```
stv0g, line 82: Comment style

dpsim/include/dpsim/Interface.h
```diff
@@ -23,53 +21,25 @@ class Interface : public SharedFactory<Interface> {
...
-  // Function used in the interface's simulation task to write all exported attributes to the queue
-  // Called once after every simulation timestep
-  virtual void popDpsimAttrsFromQueue(bool isSync = false);
+  virtual void open() = 0;
+  virtual void close() = 0;
 
   // Function called by the Simulation to perform interface synchronization
-  virtual void syncExports();
+  virtual void syncExports() = 0;
   /// Function called by the Simulation to perform interface synchronization
```
stv0g, line 35: Commeny style

dpsim/include/dpsim/InterfaceQueued.h
```diff
@@ -0,0 +1,151 @@
...
+class InterfaceQueued : public Interface,
+                        public SharedFactory<InterfaceQueued> {
+
+public:
+  typedef std::shared_ptr<InterfaceQueued> Ptr;
+
+  using AttributePacket = struct AttributePacket {
+    CPS::AttributeBase::Ptr value;
+    UInt
+        attributeId; //Used to identify the attribute. Defined by the position in the `mExportAttrsDpsim` and `mImportAttrsDpsim` lists
```
stv0g, line 30: Comment style

dpsim/include/dpsim/InterfaceQueued.h
```diff
@@ -0,0 +1,151 @@
...
+    UInt
+        sequenceId; //Increasing ID used to discern multiple consecutive updates of a single attribute
```
stv0g, line 32: Comment style

dpsim/include/dpsim/InterfaceQueued.h
```diff
@@ -0,0 +1,151 @@
...
+    unsigned char flags; //Bit 0 set: Close interface
```
stv0g, line 33: Comment style

dpsim/src/InterfaceQueued.cpp
```diff
@@ -0,0 +1,176 @@
...
+  }
+}
+
+void InterfaceQueued::syncImports() {
+  //Block on read until all attributes with syncOnSimulationStart are read
+  this->popDpsimAttrsFromQueue(true);
+}
+
+void InterfaceQueued::syncExports() {
+  //Just push all the attributes
```
stv0g, line 73: Comment style

dpsim/src/InterfaceQueued.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+void InterfaceQueued::setLogger(CPS::Logger::Log log) {
+  Interface::setLogger(log);
+  if (mInterfaceWorker != nullptr) {
+    mInterfaceWorker->mLog = log;
+  }
+}
+
+void InterfaceQueued::syncImports() {
+  //Block on read until all attributes with syncOnSimulationStart are read
```
stv0g, line 68: Comment style

dpsim/src/InterfaceQueued.cpp
```diff
@@ -0,0 +1,176 @@
...
+  this->pushDpsimAttrsToQueue();
+}
+
+void InterfaceQueued::popDpsimAttrsFromQueue(bool isSync) {
+  AttributePacket receivedPacket = {nullptr, 0, 0,
+                                    AttributePacketFlags::PACKET_NO_FLAGS};
+  UInt currentSequenceId = mNextSequenceInterfaceToDpsim;
+
+  //Wait for and dequeue all attributes that read should block on
+  //The std::find_if will look for all attributes that have not been updated in the current while loop (i. e. whose sequence ID is lower than the next expected sequence ID)
```
stv0g, line 83: Comment style

dpsim/src/InterfaceQueued.cpp
```diff
@@ -0,0 +1,176 @@
...
+             ->copyValue(receivedPacket.value)) {
+      SPDLOG_LOGGER_WARN(
+          mLog, "Failed to copy received value onto attribute in Interface!");
+    }
+    std::get<1>(mImportAttrsDpsim[receivedPacket.attributeId]) =
+        receivedPacket.sequenceId;
+    mNextSequenceInterfaceToDpsim = receivedPacket.sequenceId + 1;
+  }
+
+  //Fetch all remaining queue packets
```
stv0g, line 114: Comment style

dpsim/src/InterfaceQueued.cpp
```diff
@@ -0,0 +1,176 @@
...
+}
+
+void InterfaceQueued::WriterThread::operator()() const {
+  bool interfaceClosed = false;
+  std::vector<InterfaceQueued::AttributePacket> attrsToWrite;
+  while (!interfaceClosed) {
+    AttributePacket nextPacket = {nullptr, 0, 0,
+                                  AttributePacketFlags::PACKET_NO_FLAGS};
+
+    //Wait for at least one packet
```
stv0g, line 145: Comment style

dpsim/src/InterfaceQueued.cpp
```diff
@@ -0,0 +1,176 @@
...
+    mQueueDpsimToInterface->wait_dequeue(nextPacket);
+    if (nextPacket.flags & AttributePacketFlags::PACKET_CLOSE_INTERFACE) {
+      interfaceClosed = true;
+    } else {
+      attrsToWrite.push_back(nextPacket);
+    }
+
+    //See if there are more packets
```
stv0g, line 153: Comment style

dpsim-villas/src/InterfaceVillasQueueless.cpp
stv0g: Sorry, I have to insist. Copyright info is missing here. This is a new contribution and as such should have a copyright header.

If the other VILLAS related code here does not have it, its a mistake. We should add it lateron too.

We should also try to harmonize the header style with the remaining DPsim code.

stv0g (changes requested):

dpsim-villas/src/InterfaceVillasQueueless.cpp
```diff
@@ -0,0 +1,374 @@
+// SPDX-License-Identifier: Apache-2.0
+
+#include "villas/node/memory.hpp"
```
stv0g, line 3: Please use `<` & `>` for the includes:

```suggestion
#include <villas/node/memory.hpp>
```

dpsim-models/include/dpsim-models/DP/DP_Ph1_VoltageSource.h
m-mirz: Yes, constant is better. Generator is already used for the equipment type.

n-eiling: I think not naming it Generator would be quite confusing, because all other classes used for generating the signal of a voltage source are called generator. Constant is also a bad name, because the value generated is not constant at all. Also ConstantGenerator is a pretty weird name.
I'm open to another name if you can think of a good one.

dpsim-villas/src/InterfaceVillasQueueless.cpp
n-eiling: I'm not sure what to do here. Some files have the SPDX note above but without copyright notice and some files only have a copyright notice with a different license and no SPDX at all. If I copy the SPDX header from VILLAS, I will introduce a third type of header.

n-eiling: I'll do that anyways, because it's the most correct of all options.

n-eiling:
I changed the clang format file so it enforces @stv0g  comment style. If this does not work for you I would prefer that we fix the clang format file in a separate PR, because this gets really messy here.

dpsim-models/src/DP/DP_Ph1_VoltageSource.cpp
```diff
@@ -31,12 +31,20 @@ SimPowerComp<Complex>::Ptr DP::Ph1::VoltageSource::clone(String name) {
 }
 
 void DP::Ph1::VoltageSource::setParameters(Complex voltageRef, Real srcFreq) {
-  auto srcSigSine = Signal::SineWaveGenerator::make(**mName + "_sw");
-  srcSigSine->mVoltageRef->setReference(mVoltageRef);
-  srcSigSine->mFreq->setReference(mSrcFreq);
-  srcSigSine->setParameters(voltageRef, srcFreq);
+  if (srcFreq == 0.0) {
```
m-mirz, line 34: Maybe it would be better to add another overload of setParameters without frequency. Then, we could avoid the comparison with 0.0. Comparison with a float looks dangerous if it involves equal.

m-mirz: Frequency set to 0.0 is also a default parameter of this method. I am afraid that this affects existing examples. I need to check this more thoroughly.

dpsim-models/include/dpsim-models/DP/DP_Ph1_VoltageSource.h
stv0g: > Constant is also a bad name, because the value generated is not constant at all.

Can you elaborate? I thought the purpose is to produce a constant (non oscillating) signal?
Maybe we can simply get rid of it and just use a Sine wave generator using a frequency of 0?

n-eiling: no that's too slow. I'm have changed things around a bit so there is no regression for old code.

n-eiling: The sine wave generator is too slow.

dpsim-models/src/DP/DP_Ph1_VoltageSource.cpp
n-eiling: I changed this so old code is not affected by my changes.

n-eiling:
Some of the CI errors seem to be not caused by this PR. The tests are missing `docker-compose`.

dpsim-models/src/DP/DP_Ph1_VoltageSource.cpp
m-mirz: Thanks!

m-mirz (approved):

m-mirz:
@stv0g I think that the critical points are taken care of. Would you agree to merge?

n-eiling:
@stv0g ping :)

n-eiling:
@stv0g ping!

n-eiling:
@stv0g pong!

n-eiling:
@m-mirz can you just merge it without steffens approval? I addressed the most important points of his review and the remaining points are not really critical. 

I have quite a bit of changes depending on this that have been waiting for a PR for months now.

PR #322, opened 2024-10-21
Update WSCC_9bus_mult examples and implement splitting examples in DP and EMT

pipeacosta:
Currently, the examples `WSCC_9bus_mult_coupled.cpp` and `WSCC_9bus_mult_decoupled.cpp` don't work when the `copies` option is greater than zero, due to `clone()` methods missing in some components. Additionally, the `WSCC_9bus_parallel.ipynb` Notebook must be fixed because it does not find the CIM files of the scenario. Finally, we want to implement a similar Noteboook, based on the python script [decoupling_9bus_existing.py](https://github.com/sogno-platform/dpsim/blob/master/examples/Python/RuntimeMeas/decoupling_9bus_existing.py) for testing the process for splitting the WSCC 9 bus system using a decoupling line in both the DP and the EMT domains.


.github/workflows/build_test_linux_fedora.yaml
```diff
@@ -200,6 +200,12 @@ jobs:
     - name: Run Binaries 2/2
       run: ./build/dpsim/examples/cxx/WSCC_9bus_mult_decoupled
 
+    - name: Run Binaries 2/2
+      run: ./build/dpsim/examples/cxx/WSCC_9bus_split_decoupled
+
+    - name: Run Binaries 2/2
+      run: ./build/dpsim/examples/cxx/WSCC_9bus_split_decoupled_EMT
+
```
gnakti, line 208: @leonardocarreras should we make it also part of [build_test_linux_rocky_profiling](https://github.com/sogno-platform/dpsim/blob/master/.github/workflows/build_test_linux_rocky_profiling.yaml)?

dpsim-models/include/dpsim-models/SP/SP_Ph1_CurrentSource.h
gnakti, line 1: We should move this to a separate PR. The addition of an SP current source component is not related to this PR and the component is not used/tested in any of the examples.

dpsim-models/include/dpsim-models/Components.h
gnakti, line 1: to be moved to a separate PR

dpsim-models/src/CMakeLists.txt
gnakti, line 1: to be moved to a separate PR

dpsim-models/src/SP/SP_Ph1_CurrentSource.cpp
gnakti, line 1: to be moved to a separate PR

dpsim/examples/cxx/CIM/WSCC_9bus_mult_diakoptics.cpp
```diff
@@ -113,7 +113,7 @@ int main(int argc, char *argv[]) {
   filenames = DPsim::Utils::findFiles(
       {"WSCC-09_RX_DI.xml", "WSCC-09_RX_EQ.xml", "WSCC-09_RX_SV.xml",
        "WSCC-09_RX_TP.xml"},
-      "build/_deps/cim-data-src/WSCC-09/WSCC-09_RX", "CIMPATH");
+      "../../../build/_deps/cim-data-src/WSCC-09/WSCC-09_RX", "CIMPATH");
```
gnakti, line 116: I get an error when I run the simulation because of incorrect search path here. I think the path should be kept the same as before. Please update

dpsim/examples/cxx/CIM/WSCC_9bus_split_decoupled.cpp
gnakti, line 1: Rename example to add the domain at the beginning. Make sure also to update the worflows and CMakeLists

gnakti (changes requested):

.github/workflows/build_test_linux_fedora.yaml
leonardocarreras: It could be included here too (sounds like a nice suggestion), I would not consider it as blocking issue as we are not running it on every push request https://github.com/sogno-platform/dpsim/blob/d66f8267f1abed95a379ec2baf5df8e7131f5d98/.github/workflows/build_test_linux_rocky_profiling.yaml#L40-L54

pipeacosta: Added examples to the workflow

gnakti (approved):

fwege:
I would like to use the `copies` function for some testing. Is there anything against merging this, except missing code owner reviews?

PR #325, opened 2024-11-11
VILLASfpga cosimulation development

n-eiling:
- add a real time data logger
   This writes simulation results to memory and only dumps to a file when the simulation ends. It allocates a buffer that is attributes*simulation duration/time step long - so this only works for relatively short simulations.
- adds `EMT::Ph3::ControlledVoltageSource` and `EMT::Ph3::ControlledCurrentSource`. The existing sources `EMT::Ph3::VoltageSource` and `EMT::Ph3::CurrentSource` only allow setting RMS voltages and derives the individual phase voltages from that. This cannot be used for PHIL or Co-Simulations where we need the set each phase voltage individually. The existing voltage source can also not be easily adapted to support both, so I created a new model (RTDS and Simulink also have a separate model for this so it makes sense).
- modifies `EMT::Ph3::RXLoad` to be able to implement a series R-L load instead of only a parallel R-L load. The series load is needed for the WSCC 9 bus model.
- adds `FPGACosimulation`, an example of using dpsim-villas to build a co-simulation where DPsim simulates load connected to bus 5 in a WSCC 9 bus model.
- add `FPGACosim3PhInfiniteBus`, an example that implements an infinite bus to be connected with a load in another simulator.
- make logging step times optional, as this slows down the simulation quite a bit when running in real time.
- None of the changes should break existing code, because everything is again optional and the default behavior is the old behavior.


n-eiling:
VILLASfpga does not exist anymore. I thought about VILLASnode nodetype fpga :)

n-eiling:
The CI is failing because of missing pybind interfaces. I'll try to fix it in the next days - but the rest should not change anymore.

n-eiling:
I fixed dpsimpy. 
@m-mirz  Let's merge this, because others are waiting for this.

m-mirz (approved):

PR #329, opened 2024-12-10
Trigger actions for PRs and pin the version of libcimpp

gnakti:
- Since we work now mainly from forks, I extended the actions triggered `on: push` to be additionally triggered `on: pull_request`. Activity types set to `types: [opened, synchronize, reopened]`. For more details see [here](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request). An optimised version of this is to additionally prevent feature branches to trigger on push when they already have a PR: `on: [push, pull_request]`
- Pin the version of libcimpp to avoid failure of windows workflow caused by [sogno-platform/libcimpp#29](https://github.com/sogno-platform/libcimpp/pull/29).

### DIFF
--- a/.github/workflows/build_test_linux_fedora.yaml
+++ b/.github/workflows/build_test_linux_fedora.yaml
@@ -2,6 +2,8 @@ name: Build & Test Fedora
 
 on:
   push:
+  pull_request:
+    types: [opened, synchronize, reopened]
 
 ## Build ##
 
--- a/.github/workflows/build_test_linux_fedora_minimal.yaml
+++ b/.github/workflows/build_test_linux_fedora_minimal.yaml
@@ -2,6 +2,8 @@ name: Build & Test Fedora Minimal
 
 on:
   push:
+  pull_request:
+    types: [opened, synchronize, reopened]
 
 ## Build ##
 
--- a/.github/workflows/build_test_linux_rocky.yaml
+++ b/.github/workflows/build_test_linux_rocky.yaml
@@ -2,6 +2,8 @@ name: Build & Test RockyLinux
 
 on:
   push:
+  pull_request:
+    types: [opened, synchronize, reopened]
 
 ## Build ##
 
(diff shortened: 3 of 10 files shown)


leonardocarreras:
It looks like there is a non-related problem (the runner changed) that maybe is also possible to solve in the PR or in a different one, by explicitly installing this library just after the checkout process... In theory, the solution could be something like this...

```yaml
    - name: Install dependencies (LibXml2)
      uses: microsoft/vcpkg-action@v2
      with:
        vcpkgArgs: install libxml2
```
and later, adding it to the path for cmake

```yaml
        cmake -DWITH_PYBIND=OFF -DCMAKE_TOOLCHAIN_FILE=$VCPKG_ROOT/scripts/buildsystems/vcpkg.cmake ..
```

gnakti:
Thanks @leonardocarreras!
Better solve it in a new PR with a new branch.

stv0g:
Maybe the failing pipeline is caused by https://github.com/sogno-platform/libcimpp/pull/29 ?

If thats the case, we should really pin the version of libcimpp which is used to avoid such breakage in the future.

/cc @m-mirz

m-mirz:
> Maybe the failing pipeline is caused by [sogno-platform/libcimpp#29](https://github.com/sogno-platform/libcimpp/pull/29) ?
> 
> If thats the case, we should really pin the version of libcimpp which is used to avoid such breakage in the future.
> 
> /cc @m-mirz

@stv0g yes, the version should be pinned. I thought this was already the case. In this libcimpp PR we dropped support for every parser except libxml. So windows does not work out of the box but with @leonardocarreras suggestion it might work. Let me know if you think this change causes too many issues. Basically, I would like to get rid of the arabica library.

gnakti:
I edited the PR to pin the version of libcimpp. Let's do that first and the test the proposition of Leo in a separate PR.

leonardocarreras:
It runs in Windows too, I think we can consider it ready for review again @m-mirz @stv0g 
Long term we will need to think when/if to implement the breaking changes

m-mirz (approved):

PR #331, opened 2025-01-22
Fix submodule initialization for CIMpp in Dockerfiles.

pipeacosta:
This PR fixes the building process of CIMpp in the Docker files. Since we are now checking out a specific commit of this library, we need to make sure that the git submodules are updated after the checkout.

### DIFF
--- a/packaging/Docker/Dockerfile.dev
+++ b/packaging/Docker/Dockerfile.dev
@@ -77,9 +77,10 @@ RUN pip3 install -r requirements.txt
 
 # Install CIMpp from source
 RUN cd /tmp && \
-	git clone --recurse-submodules https://github.com/cim-iec/libcimpp.git && \
+	git clone https://github.com/cim-iec/libcimpp.git && \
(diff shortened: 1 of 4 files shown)


n-eiling (dismissed):

m-mirz:
@pipeacosta can you also update the repo url? Libcimpp has moved to sogno a long time ago but I forgot to update the path.

n-eiling:
The CI is failing because of some caching issue. If you delete the cache and retry it usually works. Maybe rebasing the branch also helps.

leonardocarreras:
I cleared the cache now, let's see if it works

pipeacosta:
> @pipeacosta can you also update the repo url? Libcimpp has moved to sogno a long time ago but I forgot to update the path.

done

leonardocarreras (approved):
Looks good! We probably want to update some other parts where this is (cim-iec) in a separate PR too (scripts for example)

PR #333, opened 2025-02-06
Fix docu pipeline

leonardocarreras:
This continues #332, as it fails from a fork


leonardocarreras:
This works now, see https://github.com/sogno-platform/dpsim/deployments/github-pages

m-mirz (approved):

PR #335, opened 2025-02-13
chore: update docs on real-time and VILLASnode interfaces

n-eiling:
The VILLASnode interface has changed quite a bit, so I updated to docs.
I also updated the docs on real-time because likewise, there were quite outdated.


stv0g (approved):
LGTM

n-eiling:
@m-mirz

PR #337, opened 2025-03-25
fix(docs): Fix typo and wrong diff

stv0g:


n-eiling (approved):

stv0g:
@m-mirz & @dinkelbachjan Could you also review & approve this PR?

You approval is required as a code owner here.
Can we possibly relax this requirement?

PR #338, opened 2025-03-25
fix(examples): Only build Fpga9BusHil example if build with CIMpp

stv0g:


dpsim-villas/examples/cxx/CMakeLists.txt
```diff
@@ -33,6 +32,7 @@ set(SHMEM_SOURCES
 )
 
 if(WITH_CIM)
+	list(APPEND SHMEM_SOURCES Fpga9BusHil.cpp)
```
n-eiling, line 35: Why is the variable called 'SHMEM_SOURCES'? Hasn't the name changed since the shared memory interface?

stv0g: Yes, thats a relict from early times. We should rename this variable..

I will push another commit later..

n-eiling (approved):

n-eiling (approved):

PR #341, opened 2025-03-25
feat(cmake): Allow disable LTO builds

stv0g:


CMakeLists.txt
```diff
@@ -79,7 +82,11 @@ include(CheckSymbolExists)
 check_symbol_exists(timerfd_create sys/timerfd.h HAVE_TIMERFD)
 check_symbol_exists(getopt_long getopt.h HAVE_GETOPT)
 if(CMAKE_BUILD_TYPE STREQUAL "Release" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
-	add_compile_options(-flto -march=native -Ofast)
+	add_compile_options(-march=native -Ofast)
```
n-eiling, line 85: I wouldn't do -march=native by default because binaries will not be portable. Maybe we can move that also to a flag or combine it with LTO?

n-eiling (changes requested):
Sorry, I didn't want lto to become the default. It must have slipped into master with one of my PRs.

stv0g:
> Sorry, I didn't want lto to become the default. It must have slipped into master with one of my PRs.

It only becomes enabled in Release builds at the moments. But I am working on packaging DPsim with Nix which uses a release build by default.

I would add another option for native arch.

n-eiling (approved):

PR #342, opened 2025-03-25
chore(cmake): Remove deprecated options

stv0g:


n-eiling (dismissed):

n-eiling:
The PR look sensible to me, but I don't know if this breaks anything. The CI is also failing.
In general, I'd prefer to not remove the options but keep the old ones and maybe show a warning if anything uses them.
Maybe we just switch to having `FETCH_` be the option and `WITH_` use the variable?
Not having strong feelings on this, but I just want to bring up the point that I think nobody has a good overview of the implications of this change.

stv0g:
I have deprecated to the old option naming already 2 years ago. So, I thought its now time to finally remove them :D

> Maybe we just switch to having FETCH_ be the option and WITH_ use the variable?

There is a difference between them:

- `FETCH_` enables the download of external dependencies via CMake External Project feature
- `WITH_` enables or disables a feature altogether, even if the required dependencies are available

stv0g:
I've checked all the other deprecated options. They are not used anymore and are now safe to remove.

Ready for review, approval & merge :)

stv0g:
@leonardocarreras I think this is now ready for merging :)

leonardocarreras (approved):
Looks good

PR #349, opened 2025-04-01
python: Harmonize comment style

stv0g:


examples/Python/RuntimeMeas/decoupling_9bus_comp.py
```diff
@@ -90,8 +90,8 @@ def do_sim(name, system):
 
 for i in range(0, 6):
     varname = 'BUS' + str(i+4) + '.v'
-#varname = 'BUS4.v'
-#i = 0
+    # varname = 'BUS4.v'
+    i = 0
```
fwege, line 94: Did you comment-in `i = 0` on purpose? With that, we will plot the same time series five times.

stv0g: Addressed. Thanks for the thorough review :)

fwege (approved):

PR #352, opened 2025-04-01
Add pre-commit

stv0g:


stv0g:
To avoid blowing up this PR, I started by enabling only the trailing whitespaces linter initially.

The remaining linters are already configured in the `.pre-commit-config.yaml`, but still disabled.

I will use follow-up PRs to gradually enable them and make the repo pass those new checks.

PR #355, opened 2025-04-01
Fix GitHub actions workflow for publishing to PyPi

stv0g:


leonardocarreras (approved):
Looks good, we probably want to update VILLAS in all the dockerfiles/scripts in a future PR.

PR #358, opened 2025-04-01
Fix Rocky builds with Clang due to -Wnan-infinity-disabled and -ffast-math

### LINKED ISSUE #344: CI random errors when compiling with clang in Rocky
**Describe the bug**
The pipeline fails for Rocky Linux compiling with clang, when is not the first time

**To Reproduce**
1. Make a PR
2. Push additional commits
3. Compilation starts to fail
4. Cleaning github actions cache solves
5. After deleting the cache, it pass again

**Expected behavior**
No problems when running the CI.

**Additional context**
It also has the same behaviour for the master branch, cache stops being useful randomly for this specifics.

stv0g:
This is a fix for failing Clang builds as this one:

https://github.com/sogno-platform/dpsim/actions/runs/14202423530/job/39792222373?pr=348

Reason behind this failure is:

On Clang -Ofast enables -ffast-math which in turn enables -ffinite-math-only.
This option causes all calls to functions checking for infinity or NaN to raise a warning (e.g. std::isnan).
As we also set -Werror, these warnings are causing the build to fail.
With this compiler option, we excplicity silence this warning.

Possibly fixes #344

### DIFF
--- a/.github/workflows/build_test_windows.yaml
+++ b/.github/workflows/build_test_windows.yaml
@@ -31,5 +31,6 @@ jobs:
         cd ${{ github.workspace }}
         cd build
         git describe --tags --abbrev=0 --match "v*"
-        cmake -DWITH_PYBIND=OFF ..
+        # TODO: Remove '-DCMAKE_POLICY_VERSION_MINIMUM=3.5' once spdlog version has been bumped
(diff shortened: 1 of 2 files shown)


stv0g:
The remaining build failure of the Windows workflow is fixed by #359

leonardocarreras (approved):
Lets merge it

PR #359, opened 2025-04-01
Fix Windows builds with newer CMake versions

stv0g:
This PR fixes Windows builds in the GitHub CI like this one:
https://github.com/sogno-platform/dpsim/actions/runs/14202958450/job/39793963971?pr=358

Windows builds are failing due to GitHubs Windows CI runners shipping a newer CMake version which is not anymore compatible with the old CMake version requiremenst of our ancient spdlog version.

This can probably reverted once #340 has been merged.

PR #360, opened 2025-04-01
Fix Windows and Rocky/Clang builds

stv0g:
This is a combination of #358 and #359 on order to resolve their cross-dependency on each other.


leonardocarreras (approved):
+1

PR #362, opened 2025-04-02
Allow builds without nlohmann/json library

stv0g:
We actually depend in a very few places on this library.

Lets make it optional.

Depends on #342 to be merged first.


fwege (approved):
Looks reasonable to me.
With `FETCH_JSON=ON` being the default, the usage of this library becomes opt-out and the CI still includes the scenarios relying on it.

PR #376, opened 2025-04-04
Fix CMake typos and style

stv0g:


iripiri (approved):

dpsim/include/dpsim/Simulation.h
```diff
@@ -263,7 +263,7 @@ class Simulation : public CPS::AttributeList {
     mInterfaces.push_back(eint);
   }
 
-#ifdef WITH_GRAPHVIZ
+#ifdef js
```
leonardocarreras: https://github.com/sogno-platform/dpsim/blob/816f697789bc6c36b638f20cf54247d1483ea30a/CMakeLists.txt#L226 
I think this is for the graph, does this work? I cannot find the definition of `js`

leonardocarreras (changes requested):
I think we can skip one of the commits, the other looks reasonable

leonardocarreras (approved):

PR #380, opened 2025-04-06
Add support for CMake `find package` and rename "KLU" dependency to name of actual library "SuiteSparse"

stv0g:
This PR contains two changes:

#### Add support for CMake find package

We dont want to necessarily solely depend on CMake's `FetchModule`.

When building with Nix, we even have to declare all dependencies externally.

#### Rename "KLU" dependency to name of actual library "SuiteSparse"

KLU is a feature, SuiteSparse is the dependency name.


leonardocarreras (approved):
Looks fine

PR #381, opened 2025-04-06
feat(cmake): Add install target

stv0g:
Lets support a `make install`, to install the DPsim shared library, Python extension modules and headers to the system.


m-mirz (approved):

stv0g:
Hi @m-mirz,

thanks for approving. Unfortunately, the current branch protection rules require an approval on the most recent HEAD of the PR. Since, I've merged #377 before and now rebased this PR, we need your approval again :(

Can we maybe relax this rule in a way a single approval even not on the HEAD of the PR is sufficient?

PR #386, opened 2025-04-08
Add pre-commit /8

stv0g:
Add markdownlint hook and fix its warnings in the documentation


stv0g:
This is also ready for merge

leonardocarreras (approved):

Haifischbecken (approved):
If you are certain that the doc files should not have a top level header at the start, this seems like a great improvement.

Issue #389, opened 2025-04-18
Remove commented out code


georgii-tishenin:
Hi Team,

While browsing through the codebase, I noticed that there are several blocks of code that have been commented out for many years. Thanks to @stv0g, this code is now commented out using compile-time directives, like:

```
#if 0
// some code here
#endif
```
which is already much nicer than simply commented out code.

Anyway, I'd suggest to remove such code.  If required, git history has this code. I'd start with the solver and model files. I think, it's still acceptable to have commented out code in the "examples".

If this sounds okay for you, I can do it.

georgii-tishenin:
@stv0g, @leonardocarreras what do you think?

PR #390, opened 2025-04-18
Remove commented out code

georgii-tishenin:
This PR removes commented out code, as discussed in #389


stv0g (approved):
I generally agree. Especially if there is no context why the code is disabled. With context given, o think we can accept dead code if it will be helpful for re-enabling or completing a feature.

PR #392, opened 2025-05-09
Windows: build with pybind and add workflow for wheels

### LINKED ISSUE #368: Start building wheels for Windows

leonardocarreras:
This is an initial test to build windows wheels, following the idea in #368

This PR does:
- Create a Pybind compilation test for windows
- Create a CI workflow to publish python c313 artifacts for windows

Additionally, to make this possible:
- Fixes some of the bindings due to sonarcloud errors
- Changes CMakefiles to allow the multi-platform (and multi-compiler) use in the cmake dependencies
- Fixes some missing destructors

### DIFF
--- a/.github/workflows/build_test_windows.yaml
+++ b/.github/workflows/build_test_windows.yaml
@@ -52,7 +52,7 @@ jobs:
       uses: actions/cache@v4
       with:
         path: ${{ github.workspace }}/build
-        key: build-windows-cache-${{ github.ref }}
+        key: build-windows-bindings-cache-${{ github.ref }}
 
     - name: Build DPsim
       run: |
@@ -62,4 +62,4 @@ jobs:
         git describe --tags --abbrev=0 --match "v*"
         # TODO: Remove '-DCMAKE_POLICY_VERSION_MINIMUM=3.5' once spdlog version has been bumped
         cmake -DWITH_PYBIND=ON -DCMAKE_POLICY_VERSION_MINIMUM="3.5" ..
-        cmake --build . --target dpsim --parallel
+        cmake --build . --target dpsimpy --parallel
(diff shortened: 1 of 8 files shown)


stv0g:
Hi @leonardocarreras,

Thanks for this great initiative :)

Is there a motivation why you dont you [cibuildwheel](http://cibuildwheel.pypa.io/) as we already use it for the Linux wheels:

https://github.com/sogno-platform/dpsim/blob/3cab0c2898a9f14be13e2f7bbd17c5665e7f19fa/.github/workflows/publish_to_pypi.yaml#L46

I am strong for using it, as this will limit our maintenance burden in the long run as cibuildwheel is an officil project of the Python Packaging Authority (PyPA).

Also, it would allow us to build Wheels for more platforms like MacOS and automatically for more Python versions & interpreters.

stv0g:
Ideally, we use the existing GitHub actions workflow and extend it with a build matrix for more platforms.

leonardocarreras:
Hi @stv0g 

Maybe there are two different things in this PR, they can be split if needed. One is just about being able to compile in Windows with pybind, which is now possible after some changes.

The other is the cibuildwheel, that I actually did it in a new file
https://github.com/sogno-platform/dpsim/blob/f1b398948955bc49d00d1236eb10246d02b092ac/.github/workflows/publish_to_pypi_win.yaml#L69-L77

I did not know exactly how to join the linux and windows ones together (the way to create the matrix is a bit more convoluted), probably something like this is needed
https://github.com/matplotlib/matplotlib/blob/main/.github/workflows/cibuildwheel.yml

leonardocarreras:
Lets say that I can try to find some more time (or some more help) to make it cleaner in the next weeks, but for the moment the good news is that it works on windows (I tested locally with VS2022 too).

stv0g:
Hi @leonardocarreras,

Sorry I missed the CI build wheel part. I will have another and closer look.

leonardocarreras:
Superseded by #633

Issue #394, opened 2025-06-20
Add Code Coverage Report to CI/CD


AnasBahr:
Description
We should add automatic **code coverage reporting** to the CI/CD pipeline. This will help monitor test quality and ensure that our codebase maintains sufficient test coverage over time.

---

#### Acceptance Criteria
- [ ] Code coverage is measured during automated test runs
- [ ] Coverage report is generated and stored as an artifact or shown in CI output
- [ ] Integrates with external coverage tracking services (Codecov)

---

#### Suggested Tools and Frameworks
- **For Python**: `pytest`, `pytest-cov`, ` nbformat `, ` nbconvert`
- **CI System**: GitHub Actions (`.github/workflows/ci.yml`)
- **For C++**: `lcov`
- **For analyzing coverage report**:  `Codecov`

---

#### Affected Areas
- CI configuration (`.github/workflows/build_test_linux_fedora.yaml`)
- CMake configuration  (`CMakeLists.txt`)
- Possibly `.gitignore` (to exclude `.coverage`, `htmlcov/`, etc.)

---

PR #395, opened 2025-06-22
Add Code Coverage Report

georgii-tishenin:
This PR implements code coverage reporting as discussed in #394:

- Adds coverage option to CMakeLists.txt

- Updates CI workflow to generate and upload coverage reports

- Integrates with Codecov for external coverage tracking


.github/workflows/build_test_linux_fedora.yaml
```diff
@@ -113,23 +113,49 @@ jobs:
...
 
     - name: Run pytest
       shell: bash
       working-directory: ${{ github.workspace }}
       env:
         PYTHONPATH: "${{ github.workspace }}/build"
-      run: |
-        cp -r python/src/dpsim build/
-        pytest -v examples/Notebooks -n auto
+      run: pytest --cov --cov-branch --cov-report=xml examples/Notebooks/Circuits/
```
leonardocarreras, line 123: Hi! Some questions here:
- any specific reason to avoid testing the other Notebooks? There are other tests outside Circuits (notebooks with "assert")
- any specific reason to drop -n? if the testing does not run in parallel will be slow, does this way of executing the tests has a default to parallel?

.github/workflows/build_test_linux_fedora.yaml
```diff
@@ -113,23 +113,49 @@ jobs:
...
+      run: |
+        lcov --capture --directory . --output-file coverage.info --rc lcov_branch_coverage=1 --ignore-errors mismatch
+        lcov --remove coverage.info '/usr/*' '*/test/*' --output-file coverage.cleaned.info
+        lcov --list coverage.cleaned.info
+      continue-on-error: true
+
+    - name: Upload coverage artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: coverage-rocky
```
leonardocarreras, line 145: is this rocky just because? I think we are running in fedora, might be good to adjust the name

leonardocarreras (changes requested):
Hi! Thanks for the PR! Looks good in general, but I have some questions about the changes

stv0g:
Could you rewrite the commit messages to follow a common style? I see lower- & upper-case beginnings here.

stv0g:
I also believe, we could stash the last 5 commits into one. Having one for the CMakeLists and another one for the GitHub actions workflow should be sufficient :)

.github/workflows/build_test_linux_fedora.yaml
```diff
@@ -113,23 +113,49 @@ jobs:
...
+      run: |
+        lcov --capture --directory . --output-file coverage.info --rc lcov_branch_coverage=1 --ignore-errors mismatch
+        lcov --remove coverage.info '/usr/*' '*/test/*' --output-file coverage.cleaned.info
+        lcov --list coverage.cleaned.info
+      continue-on-error: true
+
+    - name: Upload coverage artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: coverage-fedora-${{ github.ref_name }}-${{ github.run_id }}
```
leonardocarreras, line 145:
```suggestion
name: coverage-fedora-${{ github.ref_name.replace('/', '_') }}-${{ github.run_id }}

leonardocarreras (dismissed):

stv0g (dismissed):

PR #397, opened 2025-07-16
Fix Eigen matrix initialization in examples

leonardocarreras:
There is an error with Eigen when creating the matrixes


leonardocarreras:
Fixed in a cleaner way in #401

PR #400, opened 2025-08-29
Fix realtime datalogger

leonardocarreras:
There is a segmentation fault coming from the way the attributes are accessed for some cases.
    
 Additionally, now the model is added to DPsim.h and a more verbose behaviour is implemented when saving the memory dump.


leonardocarreras:
In this case, even when using the provided test notebook, the tests do not seem to log the execution of the methods. I suggest we can fix the code checking in a separate PR.

PR #402, opened 2025-08-29
Feature: New exciter models

leonardocarreras:
This is based on various PRs from @martinmoraga at the times of his work on ACS.

Adaptations are mine.

Ongoing work.

New models:
- [x] Add Model: Exciter DC1
- [x] Add Model: Exciter DC1 Simplified > Replaces old Exciter (.h & .cpp)
- [x] Add Model: Exciter DC1 Simplified for sT1
- [x] Add Model: Exciter DC1 Static

Updates
- [x] Add Base Exciter model
- [x] Add models to Factory
- [x] Add models to Component list 
- [x] Update Generator models to the new Exciters 
- [x] Keep a deprecated interface for backwards compatibility in generators 
- [x] Update existing C++ examples 
- [x] Update python bindings
- [x] Update notebook examples
- [x] Add documentation

Other
- [x] Fix logger (fmt related) for handling new exciter


leonardocarreras:
Hi @georgii-tishenin @pipeacosta, I know this is not a small PR. Let me know if it is possible to have a look, or if we need to do it differently. Thanks in advance.

dpsim-models/src/Base/Base_ReducedOrderSynchronGenerator.cpp
```diff
@@ -579,22 +580,75 @@ void Base::ReducedOrderSynchronGenerator<VarType>::mnaCompPostStep(
...
+  }
+  mExciter->setParameters(params);
 }
 
 template <typename VarType>
 void Base::ReducedOrderSynchronGenerator<VarType>::addExciter(
-    std::shared_ptr<Signal::Exciter> exciter) {
+    std::shared_ptr<Base::Exciter> exciter) {
   mExciter = exciter;
   mHasExciter = true;
```
georgii-tishenin: would it make sense to do the null check for passed `exciter` as in `addExciter(std::shared_ptr<Base::Exciter> exciter, std::shared_ptr<Base::ExciterParameters> params)`?

dpsim-models/src/Base/Base_ReducedOrderSynchronGenerator.cpp
```diff
@@ -579,22 +580,75 @@ void Base::ReducedOrderSynchronGenerator<VarType>::mnaCompPostStep(
...
 }
 
+// Deprecated: only for compatibility
```
georgii-tishenin, line 614: just using opportunity to ask. When/how often would you remove deprecated things?

dpsim-models/src/Signal/ExciterDC1.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+#include <dpsim-models/Signal/ExciterDC1.h>
+
+using namespace CPS;
+
+Signal::ExciterDC1::ExciterDC1(const String &name, CPS::Logger::Level logLevel)
+    : SimSignalComp(name, name, logLevel) {}
+
+void Signal::ExciterDC1::setParameters(
+    std::shared_ptr<Base::ExciterParameters> parameters) {
```
georgii-tishenin, line 12: I know is not usually done in DPsim, so just pointing your attention to this (does not have to be addressed in this PR).

Several parameters are later used as divisors, for example `Tr`, `Tb`, `Ta`, `Tf`, `Tef`, and `Ka`. If one of them is zero or missing, the simulation may continue with `inf`/`nan` values instead of failing clearly during setup. A simple parameter validation would help. This is probably easier to catch in `setParameters()` than later in `initialize()` or `step()`.

I think this also can be true for other Exciters.

dpsim-models/src/Signal/ExciterStatic.cpp
```diff
@@ -0,0 +1,129 @@
...
+  mXb_prev = mXb;
+
+  /// Compute Vr at time k
+  if (mParameters->Tr > 0)
+    mVr = mVr_prev + dt / mParameters->Tr * (mVh - mVr_prev);
+  else
+    mVr = mVh;
+
+  /// Compute Vin at time k
+  mVin = mVref - mVr_prev + Vpss;
```
georgii-tishenin: Could we check whether this should use `mVr `instead of `mVr_prev` in this equation?

`mVr` is computed just above, while the documentation equation uses the transducer output for the current step. Using `mVr_prev` adds one extra sample delay in the static exciter.

If this is intentional, a short comment would make it clearer.

dpsim-models/src/Signal/ExciterStatic.cpp
```diff
@@ -0,0 +1,129 @@
...
+          parameters)) {
+    mParameters = params;
+
+    if (mParameters->Ta == mParameters->Tb) {
+      SPDLOG_LOGGER_INFO(
+          mSLog,
+          "\n if Ta=Tb  the auxilary state variable Xb can be ignored, as Xb=0",
+          this->name());
+      throw CPS::TypeException();
+    }
```
georgii-tishenin, line 35: Could we check this case?

The log message says that for` Ta == Tb`, the auxiliary state `Xb` can be ignored, but the code then throws an exception. If this parameter set is unsupported, a clearer error message would help. If it is valid, this may reject a simplified lead-lag case unnecessarily.

docs/hugo/content/en/docs/Models/Synchronous Generator Regulators/index.md
```diff
@@ -0,0 +1,281 @@
...
+    e_{fd}(k + \Delta t) = \frac{\Delta t}{T_{e}}(v_{e}(k) - e_{fd}(k)) + e_{fd}(k).
+$$
+
+To consider the saturation of $e_{fd}$ there are two different implementations, which is automatically selected depending of value of the parameter $K_{bc}$:
+
+**Standard ($K_{bc}=0$):**
+
+$$
+e^{*}_{fd} = e_{fd, max} \quad \quad if \quad \quad e^{*}_{fd} > e_{fd, max} \\
+e^{*}_{fd} = e_{fd, min} \quad \quad if \quad \quad e^{*}_{fd} > e_{fd, min},
```
georgii-tishenin: Possible typo: should the lower-limit condition be < efd,min instead of > efd,min?

docs/hugo/content/en/docs/Models/Synchronous Generator Regulators/images/DC1C_exciter.png
georgii-tishenin, line 1: Optional comment: if SVG versions of these figures are available, they might be preferable for documentation. But the current image files are quite small, so I do not see this as an issue.

georgii-tishenin (commented):
Thanks for the PR. I especially like that the new exciter models are added together with documentation and Python bindings.

I left a few small comments/questions. I do not see this as a strict requirement for this PR, but it would give a better feeling to have tests/examples covering all new exciter models.

dpsim-models/src/Base/Base_ReducedOrderSynchronGenerator.cpp
leonardocarreras: I am mostly sure we have a backlog on this aspect. I think one minor version after deprecating should be fine for most things, but we can definitely discuss.

dpsim-models/src/Base/Base_ReducedOrderSynchronGenerator.cpp
leonardocarreras: Will do

dpsim-models/src/Signal/ExciterDC1.cpp
leonardocarreras: Tried to implement, please see if we are thinking on the same direction

docs/hugo/content/en/docs/Models/Synchronous Generator Regulators/index.md
leonardocarreras: Fixed

docs/hugo/content/en/docs/Models/Synchronous Generator Regulators/images/DC1C_exciter.png
leonardocarreras: I currently do not have the SVGs, they belong to old commits

dpsim-models/src/Signal/ExciterStatic.cpp
leonardocarreras: Thanks for the catch. It was unintentional, changed to mVr.

dpsim-models/src/Signal/ExciterStatic.cpp
leonardocarreras: When Ta==Tb: Ca=1, Cb=0, the lead-lag becomes 1. Mathematically valid. Replaced the throw with an informational log message.

georgii-tishenin (approved):

Issue #391, opened 2025-05-09
Change installing for CIMpp


leonardocarreras:
Currently there is a deb and rpm packaged for libcimpp, we should test if it will be possible to use them instead of building from source. 

See [here](https://github.com/sogno-platform/libcimpp/releases/tag/release%2Fv2.2.0) the new version 2.2.0

PR #404, opened 2025-09-03
Changes to allow to use the rpm/deb installer for CIMpp

leonardocarreras:
This is to solve #391 when keeping backwards compatibility.


stv0g (dismissed):
LGTM

PR #408, opened 2025-10-15
fix std::max behaviour in simulation.cpp for windows build with #define NOMINMAX

### LINKED ISSUE #409: Bug in compilation for windows builds
See #408 reported initially by [cwirtz-fgh](https://github.com/cwirtz-fgh)
> * issue with std::max in simulation.cpp: windows.h seems to be included somewhere and standard C++ behaviour of std::max is broken
> * proposed fix:  define NOMINMAX for windows builds
>   #ifdef _WIN32
>   #define NOMINMAX
>   #endif

FGH-wirtz:
- issue with std::max in simulation.cpp: windows.h seems to be included somewhere and standard C++ behaviour of std::max is broken

- proposed fix:  define NOMINMAX for windows builds 
#ifdef _WIN32
#define NOMINMAX
#endif

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,11 @@ if(WIN32)
 	cmake_policy(SET CMP0074 NEW)
 endif()
 
+# Define NOMINMAX to prevent Windows macros from breaking std::max
+if(WIN32)
+  add_compile_definitions(NOMINMAX)
+endif()
+
 # Allow shadowing of CMake targets
 # As we include dependencies via add_subdirectory()
 # we import their targets into the DPsim CMake tree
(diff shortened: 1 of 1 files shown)


stv0g:
Thanks for the fix. I agree that we need to set this pre-processor variable to fix the build.
However, I believe that there is a better place. Setting it in `Simulation.cpp` only, will bite us again in the future. I propose to set it via CMake:

```cmake
if(WIN32)
  add_compile_definitions(NOMINMAX)
endif()
```

leonardocarreras:
Thanks! I agree that this goes better in CMake, we actually have done this way for other cases
https://github.com/sogno-platform/dpsim/blob/452323ca7ab3dc1e75644a56f682e6fe0c3e4ac4/CMakeLists.txt#L60-L62

leonardocarreras:
Hi @cwirtz-fgh, would you still be interested in applying the proposed fix?

leonardocarreras (dismissed):
Looks good, it only needs to pass the pre-commit workflow

leonardocarreras (approved):

Issue #326, opened 2024-11-12
Logger not used correctly


leonardocarreras:
This needs fixing, using `SPDLOG_LOGGER_INFO(`
https://github.com/sogno-platform/dpsim/blob/66fbda2f04458424513c811cfb77dd035bb02ceb/dpsim-models/src/EMT/EMT_Ph3_CurrentSource.cpp#L42

malikrohail:
can i fix this

leonardocarreras:
Hi! Thanks for your interest @malikrohail, would you mind then creating a PR in your own fork? Please remember to[ sign your commits ](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) to indicate that you adhere to the [contribution guidelines ](https://sogno.energy/docs/contribution-guidelines/).

You have examples of how to apply the required solution in the same file, for example https://github.com/sogno-platform/dpsim/blob/66fbda2f04458424513c811cfb77dd035bb02ceb/dpsim-models/src/EMT/EMT_Ph3_CurrentSource.cpp#L133-L134

If you have questions, feel welcome to ask!

shikhindahikar:
Hi @leonardocarreras may I try to fix this if no one else is working on it? I understand that we need to change the way of printing the logs to use spdlog library logger. I have couple of questions:

- I would like to know how to test these changes? Should I just make the changes and get them to compile successfully?
- I was able to successfully build the docker image and run the container but I used your fork with the branch [allow-use-cimpp-installer](https://github.com/leonardocarreras/dpsim/tree/allow-use-cimpp-installer) since I was getting errors in building the original repo. So how can I test the compilation for the original repo?

I am using the VS code attached to Docker container for development, as recommended in the docs.

leonardocarreras:
I think all should work as a start point now

leonardocarreras:
@shikhindahikar thanks for your PR! I think we can close #326 with #411

PR #411, opened 2025-10-18
Adding correct logger usage 

shikhindahikar:
This fixes the issue #326 

I searched for the definitions of the SPDLOG_LOGGER macros but could not find them in the repo or on internet, so I intuitively used the macros following how it was being used in code already.
It compiles correctly for me, but if anyone feels I have used incorrect function calling, please let me know, I will modify.


shikhindahikar:
sure, will rebase in a while

shikhindahikar:
Hi, is anyone reviewing this?

Issue #410, opened 2025-10-17
Compilation in workflow fails due to storage space


leonardocarreras:
Current compilation for all targets fails due to storage space problems.

Main culprits seems to be:
- Compiling with flags to enable code coverage
- Size of the container itself and its layers

Code coverage in C++ is targeted in #403, but maybe would be better idea if this changes to ubuntu directly (with the github actions runner natively?)

For the moment, sometimes fails and it is getting more frequent. Optimization was done with the container size, but seems not to be enough.

stv0g:
Are we using self-hosted runners yet for the GitHub actions worfklows?

leonardocarreras:
Hi! No, we are not... If  you would like/know how to do something alike, this could be a nice addition... This might also help to improve problems of #301 and make the whole pipeline faster

leonardocarreras:
Closing, we have self-hosted resources now, we can move workflows as needed.

PR #412, opened 2025-10-20
Disable code coverage in linux-fedora-examples on workflow

leonardocarreras:
Currently, errors like #410 are happening. 
To stop blocking the pipeline, we need to investigate a better way. I suggest we can turn off the code coverage compilation from cpp examples (is currently not yet used) meanwhile a solution is found.


georgii-tishenin (approved):
Hey Leo, yes, agree, we have to disable it, if it's a blocker for other PRs.

PR #424, opened 2025-11-21
Modelling and simulation of Kundur two area system

pipeacosta:
This PR adds an initial implementation of the Kundur two area system example in DPsim. The systems are implemented in c++ with corresponding Jupyter Notebooks for postprocessing.


leonardocarreras (dismissed):
Looks good to me

PR #465, opened 2026-04-15
Changed where the container workflow runs to a private runner

leonardocarreras:


Haifischbecken (approved):
Changing runner to local one as agreed

PR #470, opened 2026-04-23
Fix pybind on windows builds

leonardocarreras:
- [x] Fix filesystem cmake find
- [x] Fix setup execution for wheels in windows
- [x] Fixes on pybind-related 
- [x] Check/modify windows worflow and add a section for pybind support on compilation

Goes in the direction of #392, but will let the PyPA workflow to be solved separately

### DIFF
--- a/dpsim/src/pybind/main.cpp
+++ b/dpsim/src/pybind/main.cpp
@@ -16,19 +16,8 @@
 #include <pybind11/stl.h>
 
 #include <DPsim.h>
-#include <dpsim-models/IdentifiedObject.h>
-#include <dpsim/RealTimeSimulation.h>
-#include <dpsim/Simulation.h>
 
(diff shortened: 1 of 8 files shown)


dpsim/src/pybind/main.cpp
georgii-tishenin, line 1: so, now `dpsim/pybind/...` dependencies are all put into DPsimPy.h. But these are removed:

#include <dpsim-models/IdentifiedObject.h>
#include <dpsim/RealTimeSimulation.h>
#include <dpsim/Simulation.h>

#include <dpsim-models/CSVReader.h>

Are they not required or included through something else?

georgii-tishenin (approved):
Hey @leonardocarreras . Thanks for the PR. Looks good to me. I left one comment with question just for my understanding.

PR #472, opened 2026-05-02
Add generic EMT 3-phase two-terminal V-type SSN component

georgii-tishenin:
## Summary

This PR adds a generic EMT 3-phase two-terminal V-type State Space Nodal (SSN) component parameterized by the state-space matrices `A`, `B`, `C`, and `D`.

## Changes

- add generic EMT 3-phase two-terminal V-type SSN component
- introduce supporting SSN base class hierarchy
- add C++ example
- add pybind bindings for the new component
- add Python notebook validation example
- refactor existing EMT 3-phase SSN inductor and full serial RLC to reuse the new two-terminal V-type SSN base class

## Notes

- This PR continues #406 by adding the 3-phase implementation and supporting SSN base class hierarchy.
- The new component `GenericTwoTerminalVTypeSSN` is intended as a reusable generic SSN building block for EMT simulation. It can represent arbitrary two-terminal linear time-invariant subsystems with voltage input and current output, for example loads, lines, or grounding transformers.
- The base class `TwoTerminalVTypeSSNComp` is intended as a foundation for specialized SSN component implementations.
- Tentative next steps: refactor and remove redundant specialized SSN classes, add I-type SSN, add four-terminal SSN, implement `MNAVariableCompInterface` and extend the approach to DP and SP domains.

## Inheritance overview

```mermaid
classDiagram
    class MNASimPowerComp~Real~ {
        <<abstract>>
    }
    class SSNComp {
        <<abstract>>
    }
    class VTypeSSNComp {
        <<abstract>>
    }
    class TwoTerminalVTypeSSNComp {
        <<abstract>>
    }
    class GenericTwoTerminalVTypeSSN

    MNASimPowerComp~Real~ <|-- SSNComp
    SSNComp <|-- VTypeSSNComp
    VTypeSSNComp <|-- TwoTerminalVTypeSSNComp
    TwoTerminalVTypeSSNComp <|-- GenericTwoTerminalVTypeSSN

### CODE REVIEW

File: dpsim-models/include/dpsim-models/EMT/EMT_SSNComp.h [hunk 1]
```diff
@@ -0,0 +1,81 @@
...
+  void updateState(const Matrix &uOld, const Matrix &uNew);
+
+  virtual Attribute<Matrix>::Ptr inputAttribute() const = 0;
+  virtual Attribute<Matrix>::Ptr outputAttribute() const = 0;
+
+public:
+  void setParameters(const Matrix &A, const Matrix &B, const Matrix &C,
+                     const Matrix &D);
+
+  void mnaCompInitialize(Real omega, Real timeStep,
```
leonardocarreras, line 63: Consider if the intention is to avoid `overrides `of the parent class functions or not.

dpsim-models/include/dpsim-models/Components.h
```diff
@@ -84,6 +84,10 @@
 #ifdef WITH_SUNDIALS
 #include <dpsim-models/DP/DP_Ph3_SynchronGeneratorDQODE.h>
 #endif
+
+#include <dpsim-models/EMT/EMT_SSNComp.h>
```
leonardocarreras, line 88: Might be good to think if this will not belong to `Base` type of components, if fully abstract

dpsim-models/src/EMT/EMT_Ph3_GenericTwoTerminalVTypeSSN.cpp
```diff
@@ -0,0 +1,17 @@
...
+#include <dpsim-models/EMT/EMT_Ph3_GenericTwoTerminalVTypeSSN.h>
+
+using namespace CPS;
+
+EMT::Ph3::GenericTwoTerminalVTypeSSN::GenericTwoTerminalVTypeSSN(
+    String uid, String name, Logger::Level logLevel)
+    : TwoTerminalVTypeSSNComp(uid, name, logLevel) {}
+
+SimPowerComp<Real>::Ptr
+EMT::Ph3::GenericTwoTerminalVTypeSSN::clone(String name) {
```
leonardocarreras, line 13: Maybe consider `override` here (and also `final`)... Just a question, would not this be the same for all components of this kind? or do they need more than `mA`,`mB`,`mC`,`mD`... is okay the rest of the attributes are not initialized?

dpsim/examples/cxx/Components/EMT_Ph3_GenericTwoTerminalVTypeSSN.cpp
```diff
@@ -0,0 +1,271 @@
...
+
+using namespace DPsim;
+using namespace CPS::EMT;
+
+class Example_VS_RLBranch_RLCLoad {
+public:
+  Example_VS_RLBranch_RLCLoad()
+      : mTimeStep(0.0001), mFinalTime(0.2), mFrequency(50.0),
+        mSourceVoltage(CPS::Math::polar(1.0, 0.0)),
+        mIdentity3(Matrix::Identity(3, 3)), mBranchResistance(0.2 * mIdentity3),
```
leonardocarreras: `mBranchResistance(0.2 * mIdentity3)` could be `mBranchResistance(Math::singlePhaseParameterToThreePhase(0.2))` and this avoid to use this `mIdentity3` introduced here

dpsim-models/include/dpsim-models/EMT/EMT_Ph3_SSN_Inductor.h
```diff
@@ -8,29 +8,27 @@
 #pragma once
 
 #include <dpsim-models/Base/Base_Ph3_Inductor.h>
```
leonardocarreras, line 10: Is this still needed?

dpsim-models/include/dpsim-models/EMT/EMT_Ph3_TwoTerminalVTypeSSNComp.h
```diff
@@ -0,0 +1,34 @@
...
+/// - reconstruction of the initial input voltage from terminal phasors
+/// - conductance matrix stamp
+/// - right-side history current stamp
+/// - interface voltage update from the MNA solution
+class TwoTerminalVTypeSSNComp : public VTypeSSNComp {
+protected:
+  TwoTerminalVTypeSSNComp(String uid, String name,
+                          Logger::Level logLevel = Logger::Level::off);
+
+  MatrixComp buildInitialInputFromNodes(Real frequency) final;
```
leonardocarreras, line 24: Would it be possible to elaborate a bit more on how this works vs. the `initializeFromNodesAndTerminals` from other components?

dpsim-models/src/EMT/EMT_Ph3_SSN_Full_Serial_RLC.cpp
```diff
@@ -1,362 +1,59 @@
...
-    AttributeBase::List &attributeDependencies,
-    AttributeBase::List &modifiedAttributes,
-    Attribute<Matrix>::Ptr &leftVector) {
-  attributeDependencies.push_back(leftVector);
-  modifiedAttributes.push_back(mIntfVoltage);
-  modifiedAttributes.push_back(mIntfCurrent);
-}
+  Matrix inverseInductance = inductance.inverse();
+  Matrix inverseCapacitance = capacitance.inverse();
+  Matrix identity3 = Matrix::Identity(3, 3);
```
leonardocarreras, line 38: I set a suggestion for this type of cases, maybe it works

dpsim-models/src/EMT/EMT_VTypeSSNComp.cpp
```diff
@@ -0,0 +1,57 @@
...
+Attribute<Matrix>::Ptr EMT::VTypeSSNComp::outputAttribute() const {
+  return mIntfCurrent;
+}
+
+void EMT::VTypeSSNComp::initializeFromNodesAndTerminals(Real frequency) {
+  if (!mParametersSet)
+    throw std::logic_error("setParameters() must be called before "
+                           "initializeFromNodesAndTerminals().");
+
+  MatrixComp uInit = buildInitialInputFromNodes(frequency);
```
leonardocarreras, line 28: This function is used at the moment only 1 time, does it need to be a function?

dpsim-models/src/CMakeLists.txt
```diff
@@ -63,6 +63,9 @@ list(APPEND MODELS_SOURCES
 	# DP/DP_Ph3_SynchronGeneratorVBR.cpp
 	# DP/DP_Ph3_SynchronGeneratorVBRStandalone.cpp
 
+	EMT/EMT_SSNComp.cpp
+	EMT/EMT_VTypeSSNComp.cpp
+
```
leonardocarreras, line 68: Might be good to think if this will not belong to `Base` type of components, if fully abstract

dpsim-models/include/dpsim-models/EMT/EMT_Ph3_SSN_Full_Serial_RLC.h
```diff
@@ -1,96 +1,52 @@
...
- * This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at https://mozilla.org/MPL/2.0/.
- *********************************************************************************/
+// SPDX-FileCopyrightText: 2026 Institute for Automation of Complex Power Systems, EONERC, RWTH Aachen University
+// SPDX-License-Identifier: MPL-2.0
 
 #pragma once
 
 #include <dpsim-models/Base/Base_Ph3_Capacitor.h>
```
leonardocarreras, line 6: Are those still needed here?

dpsim-models/include/dpsim-models/Components.h
```diff
@@ -103,6 +107,7 @@
 #include <dpsim-models/EMT/EMT_Ph3_ControlledCurrentSource.h>
 #include <dpsim-models/EMT/EMT_Ph3_ControlledVoltageSource.h>
 #include <dpsim-models/EMT/EMT_Ph3_CurrentSource.h>
+#include <dpsim-models/EMT/EMT_Ph3_GenericTwoTerminalVTypeSSN.h>
```
leonardocarreras, line 110: Might be good to consider also `EMT_Ph3_VTypeSSNGenericTwoTerminal`, so they end up grouped together like the generators, for example

dpsim-models/src/EMT/EMT_Ph3_SSN_Full_Serial_RLC.cpp
```diff
@@ -1,362 +1,59 @@
...
       Base::Ph3::Resistor(mAttributes), Base::Ph3::Inductor(mAttributes),
-      Base::Ph3::Capacitor(mAttributes) {
-
-  mPhaseType = PhaseType::ABC;
-
-  **mIntfVoltage = Matrix::Zero(3, 1);
-  **mIntfCurrent = Matrix::Zero(3, 1);
-  setTerminalNumber(2);
-}
+      Base::Ph3::Capacitor(mAttributes) {}
```
leonardocarreras, line 12: Are the capacitor and inductor used?

leonardocarreras (changes requested):
Hi Georgii, thanks for the PR! I took the freedom to update a bit the diagram. I got some questions too...

```mermaid
classDiagram
    class MNASimPowerComp~Real~ {
        <<abstract>>
    }
    class SSNComp {
        <<abstract>>
    }
    class VTypeSSNComp {
        <<abstract>>
    }
    class TwoTerminalVTypeSSNComp {
        <<abstract>>
    }
    class GenericTwoTerminalVTypeSSN
    class EMT_Ph3_SSN_Full_Serial_RLC
    class EMT_Ph3_SSN_Inductor
    class EMT_Ph3_SSN_Capacitor

    MNASimPowerComp~Real~ <|-- SSNComp
    SSNComp <|-- VTypeSSNComp
    VTypeSSNComp <|-- TwoTerminalVTypeSSNComp
    TwoTerminalVTypeSSNComp <|-- GenericTwoTerminalVTypeSSN
    TwoTerminalVTypeSSNComp <|-- EMT_Ph3_SSN_Full_Serial_RLC
    TwoTerminalVTypeSSNComp <|-- EMT_Ph3_SSN_Inductor
    TwoTerminalVTypeSSNComp <|-- EMT_Ph3_SSN_Capacitor
```

- Is there any reason why SSN Capacitor was not included?

Optimization possible
- If Identity 3x3 is needed, maybe something like `Matrix::Identity` from eigen is better, maybe can be useful to do as with `Matrix::Zero`
- For the inverses, we have `Math::invertMatrix`, you may consider using it. It was discovered time ago with profiling that for 2x2 and 3x3, using the generic case takes a lot of time and kills the RT performance.

dpsim-models/include/dpsim-models/EMT/EMT_Ph3_SSN_Full_Serial_RLC.h
georgii-tishenin: In this PR I want to limit the scope of refactoring of specialized SSN components to what can be simplified **because of** inheriting from `EMT_Ph3_TwoTerminalVTypeSSNComp.h`. This is not the case.

dpsim-models/include/dpsim-models/EMT/EMT_Ph3_SSN_Inductor.h
georgii-tishenin: same here. In this PR I want to limit the scope of refactoring of specialized SSN components to what can be simplified **because of** inheriting from `EMT_Ph3_TwoTerminalVTypeSSNComp.h`. This is not the case.

dpsim-models/include/dpsim-models/EMT/EMT_Ph3_TwoTerminalVTypeSSNComp.h
georgii-tishenin: This does not replace `initializeFromNodesAndTerminals()`, but rather is called by it. 
The idea is following: 
- initializeFromNodesAndTerminals() is implemented in the `EMT_VTypeSSNComp`.
- It calls `buildInitialInputFromNodes()`, that is implemented in the derived class, because implementation depends on the number of terminals.

Both implementations are also `final`. In a sense that further derived specialized components will not have to re-implement anything.

dpsim-models/include/dpsim-models/EMT/EMT_SSNComp.h
georgii-tishenin: Yes, `final` here is intentional. The logic in `mnaCompInitialize` is common for all SSN components.

dpsim-models/include/dpsim-models/Components.h
georgii-tishenin: I agree, there are several places where this could fit. 

For now I put it into this folder because it is EMT-specific. Now I would keep it in the EMT, if you do not mind. In the future PRs I would like to extend the approach to other domains. Then, it would best belong simply one folder up: together with `MNASimPowerComp`.

dpsim-models/include/dpsim-models/Components.h
georgii-tishenin: Thanks for the proposal. The existing name sounds a bit more natural to pronounce. I would prefer to keep it if you do not mind.

dpsim-models/src/EMT/EMT_VTypeSSNComp.cpp
georgii-tishenin: Yes, because it's implemented in the derived class.

dpsim-models/src/EMT/EMT_Ph3_GenericTwoTerminalVTypeSSN.cpp
georgii-tishenin: Good point. Here the intention is that `clone() `copies the component definition, not the current runtime state. For `GenericTwoTerminalVTypeSSN`, the defining parameters are A, B, C, and D, so copying those is sufficient; the remaining quantities such as x, mdA, mdB, W, and yHist are runtime/discretization-dependent and are reinitialized later. For specialized SSN components, clone() may need to copy other physical parameters instead, so the implementation is not necessarily identical for all derived classes. I can also change the declaration to `override final` if that is clearer. Please, let me know.

dpsim-models/src/CMakeLists.txt
georgii-tishenin: I agree, there are several places where this could fit.

For now I put it into this folder because it is EMT-specific. Now I would keep it in the EMT, if you do not mind. In the future PRs I would like to extend the approach to other domains. Then, it would best belong simply one folder up: together with MNASimPowerComp.

dpsim/examples/cxx/Components/EMT_Ph3_GenericTwoTerminalVTypeSSN.cpp
georgii-tishenin: done

dpsim-models/src/EMT/EMT_Ph3_SSN_Full_Serial_RLC.cpp
georgii-tishenin: In this PR I want to limit the scope of refactoring of specialized SSN components to what can be simplified because of inheriting from EMT_Ph3_TwoTerminalVTypeSSNComp.h. This is not the case.

georgii-tishenin:
Hi @leonardocarreras. Thanks for your review and comments.

Here's reply to your general comments (not the ones in the code):
1. Capacitor is now implemented as I-Type SSN component. I did not refactor it, because I did not include base I-Type SSN component into this PR. I think this PR is already rather big, want to limit it's scope and do it by steps.
1. Local variable `Idenitity3` is not created/used in source code, but just in some examples. It is done as you say: `Matrix identity3 = Matrix::Identity(3, 3)`;
1. `inverse()` is now used only during initialization of components, this should not be critical for real-time simulation, as initialization should be done beforehand. Usage of `inverse()` during initialization of components is in line with DPsim's codebase.

dpsim-models/include/dpsim-models/EMT/EMT_Ph3_TwoTerminalVTypeSSNComp.h
leonardocarreras: Ok ok, I get your point, thanks

dpsim-models/include/dpsim-models/Components.h
leonardocarreras: No strong opinion at the moment, we can always refactor later

dpsim-models/src/EMT/EMT_Ph3_GenericTwoTerminalVTypeSSN.cpp
leonardocarreras: I think we can keep those things for a final refactor when your set of features is complete

leonardocarreras (approved):
Hi again! I agree we can improve gradually on the parts that have room for changes or refactoring. Thanks for the explanations, I think we can go for it if there are no other comments.
PS (for the future): Capacitor do not belong there in the class hierarchy

PR #473, opened 2026-05-07
Fix angle conversion in Math::polarDeg

georgii-tishenin:
Fix `Math::polarDeg` to convert degrees to radians before calling `std::polar`.


leonardocarreras (approved):
Thanks Georgii for spotting this. Looks good to me.

PR #474, opened 2026-05-10
Add variable SSN base and EMT 3-phase piecewise-linear inductor

georgii-tishenin:
## Summary

This PR adds a variable EMT V-type SSN inheritance branch and a three-phase, two-terminal piecewise-linear inductor with flux state.

## Changes

- add variable EMT V-type SSN base classes
- add EMT 3-phase two-terminal piecewise-linear inductor
- add C++ example
- add pybind bindings
- add Python notebook validation example

## Notes

- This PR builds on the linear EMT V-type SSN hierarchy introduced in #472 .
- The new variable SSN branch is intended as a basis for specialized non-linear (variable) EMT SSN components.
- The new `PiecewiseLinearInductor` models a three-phase, two-terminal inductor with
  - state: `phi_abc`
  - input: `v_abc`
  - output: `i_abc`
  - piecewise-linear relation `i_abc = f(phi_abc)`

## Inheritance overview
Classes added with this PR are blue.

```mermaid
classDiagram
    class MNASimPowerComp~Real~ {
        <<abstract>>
    }
    class SSNComp {
        <<abstract>>
    }
    class VTypeSSNComp {
        <<abstract>>
    }
    class TwoTerminalVTypeSSNComp {
        <<abstract>>
    }
    class GenericTwoTerminalVTypeSSN
    class Inductor
    class Full_Serial_RLC

    class VTypeVariableSSNComp {
        <<abstract>>
    }
    class TwoTerminalVTypeVariableSSNComp {
        <<abstract>>
    }
    class PiecewiseLinearInductor

    class MNAVariableCompInterface {
        <<interface>>
    }

    MNASimPowerComp~Real~ <|-- SSNComp
    SSNComp <|-- VTypeSSNComp
    VTypeSSNComp <|-- TwoTerminalVTypeSSNComp

    TwoTerminalVTypeSSNComp <|-- GenericTwoTerminalVTypeSSN
    TwoTerminalVTypeSSNComp <|-- Inductor
    TwoTerminalVTypeSSNComp <|-- Full_Serial_RLC

    VTypeSSNComp <|-- VTypeVariableSSNComp
    MNAVariableCompInterface <|.. VTypeVariableSSNComp
    VTypeVariableSSNComp <|-- TwoTerminalVTypeVariableSSNComp
    TwoTerminalVTypeVariableSSNComp <|-- PiecewiseLinearInductor

    style VTypeVariableSSNComp fill:#dbeafe,stroke:#1e40af,stroke-width:2px
    style TwoTerminalVTypeVariableSSNComp fill:#dbeafe,stroke:#1e40af,stroke-width:2px
    style PiecewiseLinearInductor fill:#dbeafe,stroke:#1e40af,stroke-width:2px

### CODE REVIEW

File: dpsim-models/include/dpsim-models/EMT/EMT_SSNComp.h [hunk 1]
```diff
@@ -28,6 +24,8 @@ class SSNComp : public MNASimPowerComp<Real> {
   const Int mOutputSize;
 
 protected:
+  Real mTimeStep;
```
leonardocarreras, line 27: What does this timestep represent?
Is it diffferent than 
https://github.com/sogno-platform/dpsim/blob/361e4d7c0d039e4e98ce1855352f12876699a5f6/dpsim/include/dpsim/Simulation.h#L42

dpsim-models/src/EMT/EMT_Ph3_PiecewiseLinearInductor.cpp
```diff
@@ -0,0 +1,106 @@
...
+    const std::vector<Real> &fluxBreakpoints,
+    const std::vector<Real> &currentBreakpoints) {
+  if (fluxBreakpoints.size() < 2)
+    throw std::invalid_argument("At least two flux breakpoints are required.");
+
+  if (fluxBreakpoints.size() != currentBreakpoints.size())
+    throw std::invalid_argument(
+        "Flux and current breakpoint vectors must have the same size.");
+
+  constexpr Real originTolerance = 1e-12;
```
leonardocarreras, line 31: can we put this somewhere more prominent? or is just some special case?
see 
https://github.com/sogno-platform/dpsim/blob/361e4d7c0d039e4e98ce1855352f12876699a5f6/dpsim/include/dpsim/PFSolver.h#L81

dpsim-models/src/EMT/EMT_VTypeVariableSSNComp.cpp
```diff
@@ -0,0 +1,99 @@
...
+  **mIntfVoltage = uInit.real();
+  **mX = Matrix::Zero((**mX).rows(), 1);
+
+  SPDLOG_LOGGER_INFO(mSLog,
+                     "\n--- Initialization from nodes and terminals ---");
+
+  // Fixed-point initialization for operating-point-dependent SSN models.
+  // This keeps the implementation generic while still allowing the component
+  // parameters to depend on the current state.
+  constexpr Int maxIterations = 10;
```
leonardocarreras, line 55: Same as the others, is this fixed or you want it more prominent

leonardocarreras (commented):
Hi Georgii! Just some small comments, depends on what is going to be achieved. 
Ah, this is unrelated but it seems to me that the models only works for symmetric cases (because phase A gets copied to B and C, rotated). I do not know if we should make that more clear for users or there is something I maybe missed.

dpsim-models/include/dpsim-models/EMT/EMT_SSNComp.h
georgii-tishenin: `mTimeStep` in `EMT_SSNComp` is the same global simulation time step, copied into the SSN component from `Simulation` during `mnaCompInitialize(...)`.

We store it locally because variable SSN components need to recompute their discrete-time state-space model later during the simulation and that recomputation requires to know the simulation time step. The component does not otherwise have direct access to the `Simulation` object, so the value is cached in the component.

dpsim-models/src/EMT/EMT_Ph3_PiecewiseLinearInductor.cpp
georgii-tishenin: This tolerance is used locally in function `setParameters()` to validate that the first entered point of current-flux characteristic is at (0,0). I would say for now it is a special case, as normally in DPsim there are not a lot of validations of component parameters, but also the piecewise linear inductor parameters have a bit more constraints than usual. I did not want to make it a member variable because right now it's used just in one place.

dpsim-models/src/EMT/EMT_VTypeVariableSSNComp.cpp
georgii-tishenin: Good point. 
For now, I kept `maxIterations` local because it is only used inside `initializeFromNodesAndTerminals()` for the fixed-point initialization of variable SSN components. At the moment, the only concrete component using this logic is the piecewise-linear inductor, so the current value is tied to that specific initialization procedure. Since this part of the implementation may still need adjustment once further variable SSN components are added, I preferred to keep the iteration limit local for now rather than move it to a more prominent place too early.

leonardocarreras (approved):

georgii-tishenin:
> Hi Georgii! Just some small comments, depends on what is going to be achieved. Ah, this is unrelated but it seems to me that the models only works for symmetric cases (because phase A gets copied to B and C, rotated). I do not know if we should make that more clear for users or there is something I maybe missed.

Hi @leonardocarreras! Thanks for your quick initial review and comments. I replied to your code comments directly. Here's a reply to your comment about symmetry:

- There is an assumption about symmetry during initialization: buildInitialInputFromNodes() reconstructs phase A and generates phases B and C by rotation, so initializeFromNodesAndTerminals() currently assumes a balanced positive-sequence initial condition. But I think this is an assumption in all of the DPsim models.

- The model itself is still phase-domain: voltage, flux, current, and parameter updates are handled per phase, so the component can evolve unbalanced once initialized.

- What is additionally shared across phases is the current-flux characteristic, i.e. all three phases currently use the same piecewise-linear flux-current curve, but I think this is not confusing because characteristic is set only once.

PR #478, opened 2026-05-27
Add MNA state-space extraction and modal analysis

georgii-tishenin:
## Summary

This PR adds state-space extraction for real-valued EMT simulations using the MNA solver. The implementation extracts the global discrete-time state matrix from the EMT MNA simulation model and provides a standalone state-space modal-analysis class for computing discrete-time and continuous-time eigenvalues.

## Changes

- add `MNAStateSpaceContributor` infrastructure for component-level state-space contributions
- add `MNAStateSpaceExtractor` for extracting the discrete-time state matrix from real-valued EMT MNA models
- add `StateSpaceModalAnalysis` for computing discrete-time and continuous-time eigenvalues from the extracted state matrix
- integrate state-space extraction into the MNA direct solver task flow and expose it through `Simulation`
- expose state-space extraction and modal analysis in pybind
- add C++ example for EMT 3-phase RLC state-space extraction
- add Python notebook validation example with assertions

## Notes

- The extraction is available for real-valued EMT domain simulations using the direct MNA solver.
- Supported state-contributing components are EMT 3-phase inductors, capacitors, and two-terminal V-type SSN components.
- Supported algebraic components without extraction states are EMT 3-phase resistors, voltage sources and switches. Testing of switches is not covered by examples in this PR.
- `StateSpaceModalAnalysis` is intentionally kept outside `Simulation` and `MnaSolver`; it operates on the extracted state-space model exposed by `MNAStateSpaceExtractor`.

Potential further steps:

- extend modal-analysis functionality, for example with participation factors, damping ratios, and additional modal metrics
- extend state-space extraction to other modeling domains, such as DP and SP
- support further component types, including composite network components
- support more variable components as the SSN component library is extended

## Class interaction overview

```mermaid
classDiagram
    class Simulation
    class MnaSolver
    class MnaSolverDirect
    class StateSpaceExtractionTask

    class MNAStateSpaceExtractor
    class MNAStateSpaceContributor
    class MNAStateSpaceContributorFactory

    class InductorStateSpaceContributor
    class CapacitorStateSpaceContributor
    class SSNStateSpaceContributor

    class StateSpaceModalAnalysis

    Simulation --> MnaSolver : owns solver
    MnaSolver <|-- MnaSolverDirect
    MnaSolver --> MNAStateSpaceExtractor : owns
    MnaSolverDirect --> StateSpaceExtractionTask : creates
    StateSpaceExtractionTask --> MnaSolverDirect : calls

    MNAStateSpaceExtractor --> MNAStateSpaceContributorFactory : uses
    MNAStateSpaceContributorFactory ..> MNAStateSpaceContributor : creates
    MNAStateSpaceExtractor --> MNAStateSpaceContributor : owns contributors

    MNAStateSpaceContributor <|-- InductorStateSpaceContributor
    MNAStateSpaceContributor <|-- CapacitorStateSpaceContributor
    MNAStateSpaceContributor <|-- SSNStateSpaceContributor

    StateSpaceModalAnalysis ..> MNAStateSpaceExtractor : uses extracted state matrix

### CODE REVIEW

File: dpsim/include/dpsim/MNASolverDirect.h [hunk 1]
```diff
@@ -79,6 +79,11 @@ template <typename VarType> class MnaSolverDirect : public MnaSolver<VarType> {
   /// LU factorization configuration
   DirectLinearSolverConfiguration mConfigurationInUse;
 
+  /// Tracks variable-component changes from the last solve step.
+  Bool mVariableComponentChanged = false;
```
leonardocarreras, line 83: Hi Georgii! What are we thinking that changed? In relation to this, is anything different? Not to do something specific, but maybe we want to change how we work with the _VariableComponentInterface_
https://github.com/sogno-platform/dpsim/blob/490b91ea7721d796fd11178e29e9e96933975e36/dpsim-models/src/DP/DP_Ph1_Switch.cpp#L111-L118

dpsim/include/DPsim.h
```diff
@@ -8,6 +8,7 @@
 
 #include <dpsim/Config.h>
 #include <dpsim/Simulation.h>
+#include <dpsim/StateSpaceModalAnalysis.h>
```
leonardocarreras, line 11: Do we need this to be always present?

dpsim/src/MNASolverDirect.cpp
```diff
@@ -111,7 +111,8 @@ void MnaSolverDirect<VarType>::solveWithSystemMatrixRecomputation(
     mRightSideVector += *stamp;
 
   // Get switch and variable comp status and update system matrix and lu factorization accordingly
-  if (hasVariableComponentChanged())
+  mVariableComponentChanged = hasVariableComponentChanged();
```
leonardocarreras, line 114: Any reason that now we have a new variable?

dpsim/src/MNASolverDirect.cpp
```diff
@@ -301,6 +325,8 @@ void MnaSolverDirect<VarType>::solve(Real time, Int timeStepCount) {
     } while (numCompsRequireIter > 0);
   }
 
+  mSystemMatrixChanged = (mCurrentSwitchStatus != previousSwitchStatus);
```
leonardocarreras, line 328: Is this the only change on system matrix that needs to be considered as of now?

dpsim/src/Simulation.cpp
```diff
@@ -3,10 +3,12 @@
 #include <algorithm>
 #include <chrono>
 #include <iomanip>
+#include <stdexcept>
```
leonardocarreras, line 6: Maybe we want to think about adding what may be needed here, or are they too different?
https://github.com/sogno-platform/dpsim/blob/490b91ea7721d796fd11178e29e9e96933975e36/dpsim-models/include/dpsim-models/Definitions.h#L135-L139

dpsim/src/Simulation.cpp
```diff
@@ -3,10 +3,12 @@
...
 #include <typeindex>
 
 #include <dpsim-models/Utils.h>
 #include <dpsim/DiakopticsSolver.h>
+#include <dpsim/MNASolver.h>
```
leonardocarreras, line 11: Why is this now needed and before was not?

dpsim/src/Simulation.cpp
```diff
@@ -144,14 +146,22 @@ template <typename VarType> void Simulation::createMNASolver() {
...
-      solver = MnaSolverFactory::factory<VarType>(**mName + copySuffix, mDomain,
-                                                  mLogLevel, mDirectImpl,
-                                                  mSolverPluginName);
+      auto mnaSolver = MnaSolverFactory::factory<VarType>(
+          **mName + copySuffix, mDomain, mLogLevel, mDirectImpl,
+          mSolverPluginName);
+
+      mnaSolver->doStateSpaceExtraction(mStateSpaceExtraction);
+
+      solver = mnaSolver;
```
leonardocarreras, line 164: Why do we need to do it in several steps? I do not see the change compared to the previous implementation, maybe I am missing something...

leonardocarreras (commented):
Hi Georgii! This is just an initial look, I have more things like doubts/questions/comments... In general, does this affect the real-time performance? Or can we say the performance if not using the method is comparable?

dpsim/include/dpsim/MNASolverDirect.h
georgii-tishenin: Good point. What changed here is not the detection logic itself, but the lifetime of the result.

The information was already evaluated in `solveWithSystemMatrixRecomputation()` via `hasVariableComponentChanged()`, which in turn calls `hasParameterChanged()` on the variable components. Previously, this was only needed locally to decide whether the system matrix has to be recomputed.

With state-space extraction, I also need this information later in `extractStateSpace()` to avoid unnecessary work in `MNAStateSpaceExtractor`. In particular, the extractor only restamps/rebuilds the variable contribution and recomputes the state matrix when the variable model or the active system matrix changed.

For that reason I added `mVariableComponentChanged` as a cached member in `MnaSolverDirect`. It is set in `solveWithSystemMatrixRecomputation()` and then reused in `extractStateSpace()`.

dpsim/include/DPsim.h
georgii-tishenin: I think not. Generally, we need it whenever we want to use modal analysis. So far, there is one C++ example and one Jupyter Notebook (included into this PR). But I would also prefer to have it compiled for Binder. 
Do you have recommendations on how to do it? Or what other optional feature is a good example?

If it's a bigger work, I'd also suggest to do it as a follow-up PR.

dpsim/src/MNASolverDirect.cpp
georgii-tishenin: This is related to the other comment you've made.

The evaluation of `hasVariableComponentChanged()` was already done here anyway. With state-space extraction, I also need this information later in `extractStateSpace()` to avoid unnecessary work in `MNAStateSpaceExtractor`.

For that reason I added `mVariableComponentChanged` as a cached member in `MnaSolverDirect`. It is set in `solveWithSystemMatrixRecomputation()` and then reused in `extractStateSpace()`.

dpsim/src/MNASolverDirect.cpp
georgii-tishenin: Yes, in the current implementation I think this is the only additional system-matrix change that needs to be considered here.

There are two relevant solver paths: one with system matrix recomputation and one without recomputation. In the recomputation path, changes of variable components are handled by recomputing/restamping the system matrix. In the path without recomputation, the system matrix itself is not rebuilt; instead, the active/precomputed matrix can change due to switching. This is the case this variable is intended to capture.

The variable is only used to inform `MNAStateSpaceExtractor` that the active system matrix has changed and that the extracted state matrix may need to be recomputed. Similar to `mVariableComponentChanged`, the information is detected in one place in the solver path but used later in `extractStateSpace()`, so I store it as a member variable instead of re-evaluating it there.

If we later introduce other mechanisms that can change the active system matrix without recomputation, they should probably be included

dpsim/src/Simulation.cpp
georgii-tishenin: This is now used here:
`    if (mTearComponents.size() > 0) {
      if (mStateSpaceExtraction) {
        throw std::logic_error("MNA state-space extraction does not support "
                               "Diakoptics/tearing.");
      }`

Currently, the exceptions from `Definitions.h` do not carry message. For that reason I used `std::logic_error.`
I agree that we can add some more exceptions in `Definitions.h`. If this is still okay, I'd do it as a follow up.

dpsim/src/Simulation.cpp
georgii-tishenin: This is needed now because I added `doStateSpaceExtraction()` to `MnaSolver`, not directly to the `Solver`.

In my understanding, state-space extraction belongs to `MnaSolver`, because I currently do not plan to extract state-space matrices from other solver types. Previously this include was not needed because the functions called from `Simulation` were available through the parent `Solver` class.

In principle, it would also be possible to move `doStateSpaceExtraction()` to `Solver`, but then all solvers would expose an option that is only relevant for MNA. Let me know what you think.

dpsim/src/Simulation.cpp
georgii-tishenin: This is related to your previous comment about the need to` #include <dpsim/MNASolver.h>`.

The difference is that I now need access to one MNA-specific function before storing the solver as the generic `Solver::Ptr`.

Previously, the solver returned by `MnaSolverFactory` could be assigned directly to `solver`, because all functions called afterwards were available through the base `Solver` class.

With this PR, I added `doStateSpaceExtraction()` to `MnaSolver`, not to `Solver`, because state-space extraction is currently specific to the MNA solver. Therefore I keep the concrete `MnaSolver` pointer temporarily, call `doStateSpaceExtraction()`, and then assign it to the generic `solver` pointer so the remaining initialization can stay unchanged.

So the multiple steps are not needed because of a change in the normal solver initialization itself, but only because one new configuration option is MNA-specific.

georgii-tishenin:
> Hi Georgii! This is just an initial look, I have more things like doubts/questions/comments... In general, does this affect the real-time performance? Or can we say the performance if not using the method is comparable?

Hi @leonardocarreras. Thanks for the initial look and your comments!

The intention is that this should not affect real-time performance when state-space extraction is not used. The feature is disabled by default, and in that case no state-space extraction task is added to the scheduler. The normal MNA solve path should stay comparable to the previous implementation. The only additional overhead in the normal solve path is the evaluation/storage of the boolean that indicates whether the active/precomputed system matrix changed due to switching. This is needed so the extractor can decide later whether the extracted state matrix has to be updated.

When state-space extraction is enabled, then yes, it adds extra work after the MNA solve. However, MNAStateSpaceExtractor is already partly optimized: it reuses the LU factorization from the MNA solver instead of factorizing the system matrix again, avoids explicit matrix inversion, and recomputes the extracted state matrix only when the system matrix or variable component contribution has changed. So I would say: without state-space extraction enabled, the performance should remain comparable; with extraction enabled, this was not tested as a real-time feature.

dpsim/include/dpsim/MNASolverDirect.h
leonardocarreras: I understand. I think this in the future should be done by only one call, otherwise I have the feeling that we can have an update that does not affect all the logic. Basically if someone touches code, maybe thinking all is safe and is not... I will resolve for now, but probably we need to adjust stuff later when this becomes more stable.

dpsim/include/DPsim.h
leonardocarreras: Just was thinking if this should be added to another upper level .h file. Lets then keep it this way.

dpsim/src/Simulation.cpp
leonardocarreras: We do it in follow ups, I think it becomes a side quest otherwise.

dpsim/src/Simulation.cpp
leonardocarreras: Keep it here, and then the next comment is not relevant.

leonardocarreras:
@georgii-tishenin lets then keep an eye, we can later see what we get with the RT tests, if needed we need to have some compile-time option to deactivate this logic.

leonardocarreras (approved):
Looks good, some things need to be separate issues later.

PR #480, opened 2026-05-28
Fix Windows build with std::filesystem

### LINKED ISSUE #479: Windows build fails on <experimental/filesystem>
**Describe the bug**
The Windows CI build fails because `dpsim-models/include/dpsim-models/Filesystem.h` includes `<experimental/filesystem>`. On the current GitHub Actions Windows runner with MSVC, this header is no longer available, causing the build to fail with:

```text
error C1083: Cannot open include file: 'experimental/filesystem': No such file or directory
```

The CMake configuration already checks for native C++17 filesystem support using `<filesystem>` and `std::filesystem`, so the filesystem wrapper appears to be inconsistent with the CMake detection logic.

**To Reproduce**

1. Open a pull request that triggers the Windows CI workflow.
2. Let the `Build on Windows` job run.
3. The build reaches compilation of targets such as `dpsim-models`.
4. See the MSVC error from `dpsim-models/include/dpsim-models/Filesystem.h`.

**Expected behavior**
The Windows build should compile successfully using the standard C++17 filesystem implementation, i.e. `<filesystem>` and `std::filesystem`, when the native filesystem backend is selected.

**Screenshots**
Not applicable. The relevant CI error is:

```text
dpsim-models/include/dpsim-models/Filesystem.h(12,10): error C1083:
Cannot open include file: 'experimental/filesystem': No such file or directory
```
**Additional context**
A possible fix is to update `dpsim-models/include/dpsim-models/Filesystem.h` from:

```cpp
#include <experimental/filesystem>
namespace fs = std::experimental::filesystem;
```

to:

```cpp
#include <filesystem>
namespace fs = std::filesystem;
```

As an optional cleanup, the MSVC definition for silencing experimental filesystem deprecation warnings could later be removed from the root `CMakeLists.txt`.

georgii-tishenin:
This PR fixes the Windows CI build failure caused by the use of <experimental/filesystem> with newer MSVC versions.

Changes:

- Replaces std::experimental::filesystem with C++17 std::filesystem
- Removes the obsolete MSVC warning suppression for experimental filesystem

Closes #479.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,7 +95,6 @@ include(CheckCXXCompilerFlag)
 if(MSVC)
 	# Silence Visual Studio deprecation warnings
 	add_definitions(-D_SILENCE_ALL_CXX17_DEPRECATION_WARNINGS)
-	add_definitions(-D_SILENCE_EXPERIMENTAL_FILESYSTEM_DEPRECATION_WARNING)
 
 	# Silence Visual Studio error C2220
 	add_definitions(-D_SILENCE_STDEXT_ARR_ITERS_DEPRECATION_WARNING)
--- a/dpsim-models/include/dpsim-models/Filesystem.h
+++ b/dpsim-models/include/dpsim-models/Filesystem.h
@@ -9,8 +9,8 @@
 #pragma once
 
 #ifndef USE_GHC_FS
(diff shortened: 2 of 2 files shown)


leonardocarreras (approved):
Looks good to me

PR #482, opened 2026-06-04
fix: missing links related to suitesparse

leonardocarreras:
WITH_SPARSE has been a CMake option since before the multi-backend solver was introduced, but was never wired up, setting it OFF had no effect. SparseLUAdapter was always compiled in, and WITH_KLU could turn on even with WITH_SPARSE=OFF.

This couples the flags correctly: WITH_KLU now depends on WITH_SPARSE, SparseLUAdapter is conditionally compiled, all #ifdef guards are in place, and the MnaSolverDirect constructor default no longer hardcodes KLU unconditionally.

Default behaviour (WITH_SPARSE=ON) is unchanged.


georgii-tishenin (approved):
Thanks Leo! Looks good to me.

PR #483, opened 2026-06-04
update: improve .gitignore coverage

leonardocarreras:
# .gitignore

New goal: one root `.gitignore`. A few subdirectory files were consolidated into it, and unused options discarded.

## What is ignored now

- **Build output** — `build/`, `dist/`, `outputs/`, `wheelhouse/`, `*.whl`
- **Generated files** — `*.csv`, `*.log`, `*.pdf`, `*.so`, `*.o`, `*.svg`, `*.png`
- **IDE and OS noise** — `.idea/`, `.vscode/`, `.DS_Store`, `Thumbs.db`, etc.
- **CMake artifacts** — `CMakeFiles/`, `CMakeCache.txt`, etc.
- **Docs build output** — Sphinx, Doxygen, and Hugo generated directories
- **Coverage reports** — `*.gcda`, `*.gcno`, `htmlcov/`, etc.
- **Python artifacts** — `__pycache__/`, `*.egg-info/`, `.ipynb_checkpoints/`, virtual envs

## Image exception (docs and Notebooks)

`*.svg` and `*.png` are ignored globally, but not under `docs/` or
`examples/Notebooks/` — those directories contain committed diagrams and
figures that should stay tracked.


.gitignore
```diff
@@ -42,10 +43,35 @@ install_manifest.txt
...
 
+# Windows
+Thumbs.db
+
+# Docs
+docs/sphinx/_build/
+docs/doxygen/html/
+docs/doxygen/xml/
+docs/hugo/public/
+docs/hugo/node_modules/
```
georgii-tishenin, line 71: I've noticed docs/ and docs/hugo folders have their own .gitignore files. Is the intention to now have all things here in this file. Should we remove those .gitignore files then?

georgii-tishenin (commented):
Thanks for the PR. I added one comment.

Using the opportunity one additional question related to this topic (but not specifically to this PR): .svg is now in .gitignore. Is this intended? What would be then a good way/format to add new figures to the documentation?
The documentation already has some .svg figures from before.

.gitignore
leonardocarreras: Yes, the intention is to move all to one place. I forgot to delete and commit the special ones in subfolders.

leonardocarreras:
> Thanks for the PR. I added one comment.
> 
> Using the opportunity one additional question related to this topic (but not specifically to this PR): .svg is now in .gitignore. Is this intended? What would be then a good way/format to add new figures to the documentation? The documentation already has some .svg figures from before.

I will try to find a better way. I did forgot that svg and png files can be fine in some places. I was thinking of the outputs of notebooks and examples in general.

.gitignore
georgii-tishenin: would it be okay for you to additionally do it in this PR?

leonardocarreras: yes sure, will check again and add those changes

georgii-tishenin (approved):

PR #484, opened 2026-06-04
Fix: Bumped CI to node 24 actions

### LINKED ISSUE #468: Upgrade github action versions due to deprecation of node 20 in github runners.
[Github is deprecating node 20 in github actions](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/).

The custom image of the self hosted runner currently installs the 2023 runner, to keep the actions ( such as actions/checkout@v4 used in multiple workflows ) in line with their node version, namely node 20. 

The deprecation is set to happen on september of 2026, until then, all actions should be upgraded ( e.g actions/checkout@v6 instead of v4 ), which will correspond to node 24, and we can hopefully create a runner image that always uses the latest github runner version.

### LINKED ISSUE #475: Update sonar
Sonar analysis is now not running correctly. It looks mainly caused by node 20, but we do not see the last results since december apparently

leonardocarreras:
## Summary

GitHub is removing the Node 20 runner runtime in fall 2026 (Node 24 becomes the forced default in June 2026). Every action pinned to a node20 or older runtime will break at that point. This PR proactively upgrades all 14 workflows to Node 24-compatible versions and fixes the SonarCloud workflow, which was additionally broken by a deprecated action API.

The changes here are expected to close #475 and #468.

## Changes by category

**SonarCloud (functional fix):**
- `sonarqube-scan-action` v6 → **v8.1.0** — first release shipping node24; earlier v7.x still emitted node20 warnings
- `install-build-wrapper` bumped to the same version
- Added `SONAR_HOST_URL: https://sonarcloud.io` — required by v7+; the old action defaulted it silently

**GitHub first-party actions (node24 runtime):**
- `actions/checkout` v3/v4 → **v6**
- `actions/setup-node` v4 → **v6** · explicit `node-version` bumped 20/22 → 24
- `actions/setup-python` v3 → **v6** · added explicit `python-version: '3.x'` (default changed in v6)
- `actions/cache` v4 → **v5**
- `actions/upload-artifact` v4 → **v7**
- `actions/download-artifact` v4 → **v8**

**Third-party actions (node24 runtime):**
- `peaceiris/actions-hugo` @v3 → **@v3.2.1** (v3.1.0 introduced node24)
- `peaceiris/actions-gh-pages` @v4 → **@v4.1.0**
- `docker/login-action` v3 → **v4**
- `docker/build-push-action` v5 → **v7**
- `cachix/cachix-action` v14 → **v17**
- `codecov/codecov-action` v5 → **v6**
- `pypa/gh-action-pypi-publish` v1.13.0 → **v1.14.0**

## Not changed

`pypa/cibuildwheel` stays at **v2.23.2** as v3 is a major-version bump with breaking changes to the wheel-build configuration (default manylinux image, dropped Python 3.6–3.8, changed build frontend). It uses a composite/Python runtime so it emits no Node warnings. A separate PR should change it in case this is needed, but is not critical at the moment.

### DIFF
--- a/.github/workflows/build_nix.yaml
+++ b/.github/workflows/build_nix.yaml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
         with:
           submodules: recursive
 
@@ -24,7 +24,7 @@ jobs:
           nix_path: nixpkgs=channel:nixos-unstable
 
       - name: Setup Nix cache
-        uses: cachix/cachix-action@v14
+        uses: cachix/cachix-action@v17
         with:
           name: dpsim
           authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
--- a/.github/workflows/build_test_linux_fedora.yaml
+++ b/.github/workflows/build_test_linux_fedora.yaml
@@ -15,7 +15,7 @@ jobs:
     container: sogno/dpsim:dev
     steps:
     - name: Checkout
-      uses: actions/checkout@v4
+      uses: actions/checkout@v6
       with:
         submodules: recursive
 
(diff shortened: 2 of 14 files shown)


leonardocarreras:
One extra comment: Sonar will be able to be tested only after integrating it to the repository. Now it fails because it cannot see the token from the context of the fork.

georgii-tishenin (approved):
Thanks Leo! Let's try this out.

PR #276, opened 2024-02-26
Use pyproject.toml instead of setup.cfg, pytest.ini & requirements.txt and other Python related config files

stv0g:
Modern Python is slowly adopting `pyproject.toml` as a single place to define a Python package and related tool settings like black for formatting, flake8 for linting and pytest for testing.

Lets try to move as much of the Python-related project configuration to this file and unclutter the root of the repo.

This PR does:
- Merge `requirements*.txt`, `pytest.ini` and `setup.cfg` files into `pyproject.toml`
- Update shell scripts and Dockerfiles to use `pyproject.toml`
- Introduce a pre-commit configuration file
- Add a GitHub actions workflow for running pre-commit also in the CI
- Apply/fix pre-commit checks to all files in the repo:
  - Harmonize file endings (single empty line)
  - Remove trailing whitespaces
  - Apply formatting of all Python code (also in Jupyter notebooks)
- Remove obsolete CI jobs from `.gitlab-ci.yml`

PR #485, opened 2026-06-04
Migrate Python project config to pyproject.toml

leonardocarreras:
## Summary

`pyproject.toml` is now the single source of truth for all Python project configuration. Continues work started in #276.

### What is moved

- `setup.cfg` -> `[project]` in `pyproject.toml` (name, version, authors, classifiers, URLs)
- `setup.py` metadata (`install_requires`, `python_requires`, `setup_requires`, `tests_require`, `package_dir`) -> `pyproject.toml`. `setup.py` stays but simpler: the CMake extension build can't be expressed in pure TOML.
- `requirements.txt` -> `[project.optional-dependencies].dev`
- `requirements-manylinux.txt` -> `[project.optional-dependencies].manylinux`
- `tests_require` from `setup.py` -> `.test` extra

### How Dockerfiles and scripts install dependencies

Each Dockerfile and shell script installs dependencies via an ephemeral `pip-compile` step: `pip-tools` is installed, used to resolve the declared extras into a temporary requirements file, deps are installed from it, and both `pip-tools` and the file are discarded. The dpsim package is never built or installed at this stage.

`build_wheels.sh` has been removed, as cp37 is EOL, superseded by cibuildwheel.

### Also fixed

- `Dockerfile.rpm` and `install-ubuntu-deps.sh` referenced `requirements-jupyter.txt`, which never existed. Jupyter packages are now installed explicitly.
- `pytest-runner` removed from `[build-system].requires`.

### Pending (future PR)
- `setuptools` floor bump to `>=61` (PEP 621 support) because it needs a newer manylinux container version


packaging/Python/build_wheels.sh
```diff
@@ -40,7 +40,10 @@ for PYDIR in ${PYDIRS}; do
 					   -DPYTHON_INCLUDE_DIRS=${INC} \
 					   -DPYTHON_LIBRARY=${LIB}"
 
-   "${PYDIR}/bin/pip" install --requirement requirements.txt
+   tmpdir=$(mktemp -d)
+   cp pyproject.toml "$tmpdir/"
+   "${PYDIR}/bin/pip" install "$tmpdir[manylinux]"
```
georgii-tishenin: Here `tmpdir` contains only `pyproject.toml`, but `pip install "$tmpdir[manylinux]"` installs the project from `tmpdir` with the `manylinux` extra. This may affect wheel/container dependency setup, where we want to install dependencies before building DPsim.

The current CI checks dependency resolution with pip-compile, but it does not seem to run this pip install `"$tmpdir[...]"` path directly.

pyproject.toml
```diff
@@ -1,7 +1,68 @@
 [build-system]
-requires = ["setuptools>=42"]
+requires = ["setuptools>=42", "pytest-runner", "wheel"]
```
georgii-tishenin: Is `setuptools>=42` still sufficient after moving the project metadata into `pyproject.toml`?

The new configuration now relies more directly on `pyproject.toml `project metadata. It may therefore be safer to require a newer setuptools version, for example `setuptools>=61`.

Also, is `pytest-runner` still required as a build-system dependency, or would it be enough to keep test-related dependencies in the optional extras?

georgii-tishenin (commented):
Hi @leonardocarreras, thanks for the PR!
I am not an expert in Python packaging, so please treat these as questions to double-check. I used ChatGPT-assisted review, and the main concern is whether the changed install paths are covered by CI.

pyproject.toml
leonardocarreras: Both fixed, nice catch :)

packaging/Python/build_wheels.sh
leonardocarreras: You were right. The tmpdir hack was not actually achieving "deps only, no package", so I am now generating the files from the pip-compile and with that the requirements file to install is run the same way as before. We probably want to modernize at some point with uv, but with what we have is not possible to have one clean command.
I removed the script for build_wheels, is EOL and not used anymore thanks to the CI.

georgii-tishenin (dismissed):

pyproject.toml
leonardocarreras: @georgii-tishenin I will need to roll back the version bump to 61, the manylinux container that we use has an ancient version of python, and does not support this. we can see in a next PR how to transition, if that is okay for you. I am now compiling the containers, soon will also update this PR.

georgii-tishenin: yes, totally fine👍

georgii-tishenin (approved):
Thanks for the additional updates.

PR #486, opened 2026-06-05
Feature: New PSS

leonardocarreras:
## Power System Stabilizer - PSS1A

Based on work by @martinmoraga (ACS). Adaptations and integration by @leonardocarreras.
Ongoing work.

New models:
- [x] Add Model: PSS1A (washout + two lead-lag blocks; Milano 2010, p. 371)

Updates:
- [x] Add Base PSS model (`Base_PSS.h`)
- [x] Add model to Factory
- [x] Add model to Component list
- [x] Extend existing `Exciter::step()` with optional `Vpss` input (backward-compatible)
- [x] Update generator models to compute and forward PSS output to exciter
- [x] Update existing C++ examples (`SP_ReducedOrderSG_SMIB_Fault` — `WithPSS` flag)
- [x] Update Python bindings
- [x] Add documentation + block diagram image
- [x] Notebook example


dpsim-models/src/Signal/PSS1A.cpp
```diff
@@ -0,0 +1,136 @@
...
+                     "\nmV3(t=0): {:e}"
+                     "\nmVs(t=0): {:e}",
+                     mV1, mV2, mV3, mVs);
+  mSLog->flush();
+}
+
+Real Signal::PSS1A::step(Real omega, Real activePower, Real Vd, Real Vq,
+                         Real dt) {
+  Real Vh = sqrt(pow(Vd, 2.) + pow(Vq, 2.));
+  mVh_prev = Vh;
```
georgii-tishenin: Could we double-check the handling of the previous input values here?

As far as I understand, the Euler update below uses previous values for `omega` and `activePower,` but `mVh_prev` is overwritten with the current voltage magnitude before the update. This seems to mix previous `omega`/`activePower `with current Vh.

Would it be more consistent to initialize `mVh_prev `in initialize(...), use it during the step together with the other previous inputs, and only assign `mVh_prev = Vh` at the end of `step(...)`?

Or as an alternative: leave a short comment about this decision in the code and remove the second assignment of `mVh_prev = Vh `at the end of the function

dpsim-models/src/Base/Base_ReducedOrderSynchronGenerator.cpp
```diff
@@ -672,6 +686,32 @@ void Base::ReducedOrderSynchronGenerator<VarType>::addGovernor(
...
+    std::shared_ptr<Base::PSS> pss,
+    std::shared_ptr<Base::PSSParameters> parameters) {
+  if (!mHasExciter) {
+    SPDLOG_LOGGER_ERROR(this->mSLog, "Cannot attach PSS: no exciter present. "
+                                     "Attach an exciter first.");
+    throw CPS::InvalidArgumentException();
+  }
+  mPSS = pss;
+  mPSS->setParameters(parameters);
+  mHasPSS = true;
```
georgii-tishenin, line 705: Would it make sense to check that `pss` is not null before setting `mHasPSS = true`? 
The same may also apply to the `addPSS(...)` in `Base_SynchronGenerator.cpp`.

docs/hugo/content/en/docs/Models/Synchronous Generator Regulators/index.md
```diff
@@ -271,6 +271,69 @@ $$
...
+<center>
+<figure margin=30%>
+    <img src="./images/PSS_Type1.png" width=65% alt="DC1_exciter">
+    <figcaption>Fig. 6: Control diagram of the PSS Type 1 </br>
+                Adapted from: Milano, Power System Modelling and Scripting
+    </figcaption>
+</figure>
+</center>
+
+The control diagram of this PSS is depicted in Fig. 6. It includes a washout filter and two lead-lag blocks and is described by the following set of differential equations:
```
georgii-tishenin, line 294: Does this part fully align with the implementation?

As far as I understand, the equations here describe the speed-input special case. In the code, `PSS1A` also includes `Kp` and `Kv`, so the input seems to be based on `Kw * omega + Kp * activePower + Kv * Vh`, with `Vh` computed from `Vd` and `Vq`.

dpsim-models/include/dpsim-models/Signal/PSS1A.h
```diff
@@ -0,0 +1,78 @@
...
+  /// Third stabilizer time constant (s)
+  Real T3;
+  /// Fourth stabilizer time constant (s)
+  Real T4;
+  /// Max stabilizer output signal (pu)
+  Real Vs_max;
+  /// Min stabilizer output signal (pu)
+  Real Vs_min;
+  /// Wash-out time constant (s)
+  Real Tw;
```
georgii-tishenin: For consistency would it make sense to initialize these fields explicitly, e.g. with zero values? Same like in the exciters (https://github.com/sogno-platform/dpsim/blob/master/dpsim-models/include/dpsim-models/Signal/ExciterDC1.h)

dpsim-models/include/dpsim-models/Signal/PSS1A.h
```diff
@@ -0,0 +1,78 @@
...
+
+protected:
+  /// Wash-out output
+  Real mV1;
+  /// Output of the first phase compensation block
+  Real mV2;
+  /// Output of the second phase compensation block
+  Real mV3;
+  /// PSS output at t=k
+  Real mVs;
```
georgii-tishenin, line 66: Small encapsulation question. Do we expect some class to derive from `PSS1A`? Otherwise, these could be `private`.

georgii-tishenin (commented):
Hi @leonardocarreras. Thanks for your PR! I reviewed and added a couple of comments.

dpsim-models/src/Signal/PSS1A.cpp
leonardocarreras: Thanks, this is a good catch and was a bug. Now is handled as expected, in the direction you said.

dpsim-models/src/Base/Base_ReducedOrderSynchronGenerator.cpp
leonardocarreras: Done

dpsim-models/include/dpsim-models/Signal/PSS1A.h
leonardocarreras: Yes, fixed, now there is initialization

dpsim-models/include/dpsim-models/Signal/PSS1A.h
leonardocarreras: Changed to private

docs/hugo/content/en/docs/Models/Synchronous Generator Regulators/index.md
leonardocarreras: Aligned docu to implementation. Only problem is that the image for now will depict the previous situation. There is a caption to mention this, a new task should replace the image.

georgii-tishenin: thanks. The solution is okay for me.

georgii-tishenin (approved):
Thank you for addressing comments!
In the future would be good to increase the code coverage. But let's do it as future work.

leonardocarreras:
> Thank you for addressing comments! In the future would be good to increase the code coverage. But let's do it as future work.

Sure, the plan is to integrate the governors and then have some more playground with all three (exciter+pss+governor).

PR #491, opened 2026-06-11
Feature: steam turbine governor + base governor and turbine

leonardocarreras:
## Summary

Adds a steam turbine governor and a three-stage steam turbine (HP/IP/LP) as separate objects, plus the Base::Governor and Base::Turbine interfaces.

- New models: SteamTurbineGovernor (lead-lag + integrator with rate limiter) and SteamTurbine (HP/IP/LP stages)
- addGovernorAndTurbine() on both SG base classes, same pattern as addExciter() (#402) and addPSS() (#486)
- DQTrapez and VBR generators step the new chain when attached, fall back to TurbineGovernorType1 otherwise
- Python bindings for set_parameters(), initialize_states(), and add_governor_and_turbine()
- Docs with block diagrams and equations for both models

## What comes next

Next PR adds the hydro variants using the same base classes.

## Other comments

Continuation of the controllers work from @martinmoraga at ACS.

### DIFF
--- a/dpsim-models/include/dpsim-models/Signal/SteamTurbine.h
+++ b/dpsim-models/include/dpsim-models/Signal/SteamTurbine.h
@@ -1,0 +1,70 @@
+// SPDX-FileCopyrightText: 2026 Institute for Automation of Complex Power Systems, EONERC, RWTH Aachen University
+// SPDX-License-Identifier: MPL-2.0
+
+#pragma once
+
+#include <dpsim-models/Base/Base_Turbine.h>
+#include <dpsim-models/Logger.h>
+#include <dpsim-models/SimSignalComp.h>
+
+namespace CPS {
+namespace Signal {
+
+class SteamTurbineParameters : public Base::TurbineParameters,
+                               public SharedFactory<SteamTurbineParameters> {
+public:
+  /// Power fraction of the high-pressure stage
+  Real Fhp = 0;
+  /// Power fraction of the intermediate-pressure stage
+  Real Fip = 0;
+  /// Power fraction of the low-pressure stage
+  Real Flp = 0;
+  /// Time constant of main inlet volume and steam chest (s)
+  Real Tch = 0;
+  /// Time constant of reheater (s)
+  Real Trh = 0;
+  /// Time constant of crossover piping and LP inlet volumes (s)
+  Real Tco = 0;
+};
+
+/// Three-stage steam turbine (tandem compound: HP/IP/LP)
+/// Ref.: A. Roehder, B. Fuchs, J. Massman, M. Quester, A. Schnettler,
+///       "Transmission system stability assessment within an integrated grid
+///       development process"
+class SteamTurbine : public SimSignalComp,
+                     public Base::Turbine,
+                     public SharedFactory<SteamTurbine> {
+private:
+  std::shared_ptr<SteamTurbineParameters> mParameters;
+
+  // ### State variables at time k — loggable via .attr() ###
+  /// High-pressure stage power
+  const Attribute<Real>::Ptr mPhp;
+  /// Intermediate-pressure stage power
+  const Attribute<Real>::Ptr mPip;
+  /// Low-pressure stage power
+  const Attribute<Real>::Ptr mPlp;
+  /// Mechanical output power (pu)
+  const Attribute<Real>::Ptr mPm;
+
+  // ### Staging variables for next step ###
+  Real mPhp_next = 0;
+  Real mPip_next = 0;
+  Real mPlp_next = 0;
+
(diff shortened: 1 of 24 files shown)


dpsim-models/include/dpsim-models/Signal/SteamTurbineGovernor.h
```diff
@@ -0,0 +1,96 @@
...
+#pragma once
+
+#include <dpsim-models/Base/Base_Governor.h>
+#include <dpsim-models/Logger.h>
+#include <dpsim-models/SimSignalComp.h>
+
+namespace CPS {
+namespace Signal {
+
+class SteamGorvernorParameters
```
georgii-tishenin: Probably, a typo here: should this be `SteamGovernorParameters` instead of `SteamGorvernorParameters`?

dpsim-models/src/Signal/SteamTurbineGovernor.cpp
```diff
@@ -0,0 +1,130 @@
...
+    throw CPS::TypeException();
+  }
+  if (params->T3 == 0) {
+    SPDLOG_LOGGER_ERROR(mSLog, "T3 must not be zero for {}", this->name());
+    throw CPS::InvalidArgumentException();
+  }
+  if (params->R == 0) {
+    SPDLOG_LOGGER_ERROR(mSLog, "R must not be zero for {}", this->name());
+    throw CPS::InvalidArgumentException();
+  }
```
georgii-tishenin, line 35: would it make sense to allow only positive values, i.e. to check `T3 > 0`, ` R > 0`? Or too restrictive

dpsim-models/include/dpsim-models/Signal/SteamTurbine.h
```diff
@@ -0,0 +1,70 @@
...
+  explicit SteamTurbine(const String &name)
+      : SimSignalComp(name, name), mPhp(mAttributes->create<Real>("Php")),
+        mPip(mAttributes->create<Real>("Pip")),
+        mPlp(mAttributes->create<Real>("Plp")),
+        mPm(mAttributes->create<Real>("Pm")) {}
+
+  SteamTurbine(const String &name, CPS::Logger::Level logLevel);
+
+  void setParameters(std::shared_ptr<Base::TurbineParameters> parameters) final;
+  void initializeStates(Real Pminit) final;
```
georgii-tishenin, line 65: just for my understanding: who is responsible for setting initial states? the generator or the user? If generator: will it be implemented in a future PR? (I think it's not a part of this PR yet)

georgii-tishenin (commented):
Hey @leonardocarreras . Thanks for your PR. I added a couple of comments.

dpsim-models/include/dpsim-models/Signal/SteamTurbineGovernor.h
leonardocarreras: Yes, this is a typo and it propagated everywhere. Fixed.

dpsim-models/src/Signal/SteamTurbineGovernor.cpp
leonardocarreras: Fixed. However, we may want to see globally at some point how do we deal with this consistently. Some other parameters for example should be at least non-negative (physical world quantities), but the user might want to still do it for convenience/modeling reasons. Not the case here with the controllers, all times Tx for example should be positive (or zero in some cases is tolerable).

dpsim-models/include/dpsim-models/Signal/SteamTurbine.h
leonardocarreras: We do not want the user to have to do them explicitly.

In `Base_ReducedOrderSynchronGenerator.cpp`, this happens already when the powerflow initialization occurs

```
if (mHasGovernorAndTurbine) {
    mGovernor->initializeStates(**mMechTorque);
    mTurbine->initializeStates(**mMechTorque);
}
```

For the `Base_SynchronGenerator.cpp`, I am fixing now.

leonardocarreras:
@georgii-tishenin thanks for the review. I fixed as suggested, some things like the checks maybe need a different global approach in my opinion, but refactor is always possible.

dpsim-models/src/Signal/SteamTurbineGovernor.cpp
georgii-tishenin: This comment was more to get your opinion than a call to action. But thanks for addressing already.
Yes, I agree, better to deal consistently and globally. For now from my side I am just trying to include minimal validations for the newly added code.

georgii-tishenin (approved):

PR #492, opened 2026-06-11
Feature: hydro turbine governor

leonardocarreras:
## Summary
- New `HydroTurbineGovernor`
- New `HydroTurbine`
- Python bindings
- Docs with block diagrams and equations for both models

## Notes
- Needs merging #491 first

Continuation of the controllers work from @martinmoraga at ACS.

### DIFF
--- a/dpsim-models/include/dpsim-models/Signal/HydroTurbine.h
+++ b/dpsim-models/include/dpsim-models/Signal/HydroTurbine.h
@@ -38,11 +38,8 @@ private:
   Real mX1_next = 0;
 
 public:
-  explicit HydroTurbine(const String &name)
-      : SimSignalComp(name, name), mX1(mAttributes->create<Real>("X1")),
-        mPm(mAttributes->create<Real>("Pm")) {}
-
-  HydroTurbine(const String &name, CPS::Logger::Level logLevel);
+  HydroTurbine(const String &name,
+               CPS::Logger::Level logLevel = CPS::Logger::Level::off);
 
   void setParameters(std::shared_ptr<Base::TurbineParameters> parameters) final;
   void initializeStates(Real Pminit) final;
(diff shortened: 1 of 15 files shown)


dpsim-models/include/dpsim-models/Signal/HydroTurbine.h
```diff
@@ -0,0 +1,50 @@
...
+public:
+  /// Water starting time (s)
+  Real Tw = 0;
+};
+
+/// Hydro turbine — water starting time model.
+/// Transfer function: (1 - 0.5*s*Tw) / (1 + 0.5*s*Tw)
+/// Ref.: A. Roehder, B. Fuchs, J. Massman, M. Quester, A. Schnettler,
+///       "Transmission system stability assessment within an integrated grid
+///       development process"
```
georgii-tishenin, line 24: Could we double-check the transfer function stated in the header?

The implementation uses a PT1 block with time constant `Tw/2` and then computes `Pm = 3*X1 - 2*Pgv`. This decomposition seems to correspond to `(1 - s*Tw) / (1 + 0.5*s*Tw)`, while the header comment says `(1 - 0.5*s*Tw) / (1 + 0.5*s*Tw)`.

The implementation and the documentation equations seem consistent with each other, so maybe only the transfer-function comment is off.

georgii-tishenin (commented):
Hey @leonardocarreras thanks for the PR! I added a small comment.

dpsim-models/include/dpsim-models/Signal/HydroTurbine.h
leonardocarreras: Thank you, yes the comment is incorrect. Fixed.

georgii-tishenin (approved):
Thanks for addressing the comment!

PR #497, opened 2026-06-15
Add GlobalDQ0 frame for state-space modal analysis

georgii-tishenin:
This PR extends state-space modal analysis with a `GlobalDQ0` analysis frame for extracted EMT state-space model.

Main changes:

* add metadata for abc-frame state triples in MNA state-space extraction,
* track extraction time for frame-dependent modal analysis,
* add `StateSpaceAnalysisFrame::GlobalDQ0` to `StateSpaceModalAnalysis`,
* expose the new modal-analysis frame in Python,
* update the inverter state-space extraction notebook to compare `GlobalDQ0` eigenvalues with native-frame Floquet multipliers, analytical dq0 eigenvalues, and EMT simulation results.

The notebook demonstrates that the `GlobalDQ0` modal results give the same stability conclusion as native-frame Floquet analysis for the tested stable and unstable cases.


dpsim-models/src/EMT/EMT_SSNComp.cpp
```diff
@@ -18,6 +18,11 @@ UInt EMT::SSNComp::getStateCount() const {
   return static_cast<UInt>(mA.rows());
 }
 
+std::vector<std::array<UInt, 3>>
+EMT::SSNComp::getLocalAbcStateIndexTriples() const {
+  return {};
```
leonardocarreras, line 25: Is this intentional? If yes, maybe some comment helps. Components that do not override will get a default value (that might not be what you want)

dpsim/src/MNAStateSpaceContributor.cpp
```diff
@@ -112,6 +116,12 @@ class EMTPh3CapacitorStateSpaceContributor final
                                             Matrix::Identity(3, 3));
   }
 
+  void contributeMetadata(StateSpaceMetadata &metadata,
```
leonardocarreras, line 124: This is identical to lines 81-86, maybe in a follow up, cleaning should be possible. No immediate action needed.

leonardocarreras (dismissed):
Hi Georgii! Thanks for the PR. I have small comments, they are non-blocking. Feel free to decide how you want to proceed.

dpsim-models/src/EMT/EMT_SSNComp.cpp
georgii-tishenin: added a comment.

dpsim/src/MNAStateSpaceContributor.cpp
georgii-tishenin: added a helper function to avoid duplication.

leonardocarreras (approved):
Looks good to me, thanks for having a look into the suggestions!

PR #503, opened 2026-06-17
Add attribute usage guidelines

### LINKED ISSUE #330: add guidelines for attributes

georgii-tishenin:
This PR documents guidelines for when DPsim component variables should be attributes instead of normal C++ variables and when not. 

Closes #330.

### DIFF
--- a/docs/hugo/content/en/docs/Development/attribute-usage.md
+++ b/docs/hugo/content/en/docs/Development/attribute-usage.md
@@ -1,0 +1,154 @@
+---
+title: "Attribute Usage Guidelines"
+linkTitle: "Attribute Usage"
+---
+
+# Attribute Usage Guidelines
+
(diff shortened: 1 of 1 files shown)


leonardocarreras (approved):
+1

pipeacosta (commented):
Dear Georgii,

Many thanks for the useful documentation. I was thinking if there is a good idea to include examples for static and dynamic casting of attributes.

georgii-tishenin:
> Dear Georgii,
> 
> Many thanks for the useful documentation. I was thinking if there is a good idea to include examples for static and dynamic casting of attributes.

Hi @pipeacosta,
Thanks, good point. As we agreed offline: full explanation of casting belongs more naturally to the general Attributes page. This page focuses on decision rule for when to create attributes. If this is okay for you, I would leave it out of this PR. Afterwards, we can look together where exactly and what to document.

PR #505, opened 2026-06-18
docs: document MNA solver and component initialization

leonardocarreras:
Adds `Overview/Initialization` page covering the full `MnaSolver::initialize()` sequence, per-method contracts, class hierarchy and sequence diagrams. Includes a custom `{{< mermaid >}}` shortcode (Docsy 0.6 ships without one).

Depends on `feature/virtual-node-ordering` #496 for the composite lifecycle changes the page describes.

Closes #59

PR #506, opened 2026-06-18
docs: expand task scheduling page (#188)

### LINKED ISSUE #188: Extend docu on task system

leonardocarreras:
Replaces the current short text with a full concept overview (task base class, attribute-only constraint, dependency resolution, level scheduling, scheduler variants) and a developer guide for adding tasks to signal and MNA components.

Depends on #503 for the attribute-usage cross-reference. Branch is on top of `feature/initialization-docs`; rebase onto main after #505 merges.

Closes #188

### DIFF
--- a/docs/hugo/content/en/docs/Overview/Scheduling/index.md
+++ b/docs/hugo/content/en/docs/Overview/Scheduling/index.md
@@ -19,7 +19,7 @@ That schedule is then replayed on every timestep with no further graph analysis.
 
 Every task is an instance of a class that inherits from `CPS::Task`
 (`dpsim-models/include/dpsim-models/Task.h`).
-Each subclass implements one method:
+Each subclass implements one member function:
 
 ```cpp
 virtual void execute(Real time, Int timeStepCount) = 0;
@@ -38,7 +38,7 @@ See [Attributes]({{< ref "../Attributes/index.md" >}}) for details on the attrib
 
 Only attributes can participate in scheduling.
 Plain C++ member variables (a `Real`, a `Matrix`, an internal state struct) are invisible to the scheduler, so no dependency edge can be formed around them.
-The same constraint applies to simulation data recording: both the CSV logger (`DataLogger`) and the real-time data logger (`RealTimeDataLogger`) implement `DataLoggerInterface`, whose `logAttribute()` method only accepts `AttributeBase::Ptr`.
+The same constraint applies to simulation data recording: both the CSV logger (`DataLogger`) and the real-time data logger (`RealTimeDataLogger`) implement `DataLoggerInterface`, whose `logAttribute()` member function only accepts `AttributeBase::Ptr`.
 The VILLASnode interface works the same way.
 Any value that needs to cross a task boundary, be recorded to a data file, or be exchanged with an external tool must be stored in an `Attribute<T>`.
 
@@ -47,16 +47,20 @@ It is not part of the scheduling system and can print any value regardless of wh
 
 For practical rules on when a variable should be an attribute versus a plain member variable, see [Attribute Usage Guidelines]({{< ref "../../Development/attribute-usage.md" >}}).
 
-## PreStep and PostStep conventions
+## Common component task conventions
+
+The names below are component and solver conventions, not scheduler-level concepts.
+The scheduler only sees the attribute dependencies a task declares; it has no notion of a "PreStep" or "PostStep" and never orders tasks by these names.
 
 MNA components typically define two task classes per component:
 
 | Task | Typical responsibility |
 |------|----------------------|
-| `MnaPreStep` | Copy previous-timestep state; stamp the right-hand side before the matrix solve |
-| `MnaPostStep` | Extract interface voltage and current from the solution vector |
+| `MnaPreStep` | Component-specific preparation before the matrix solve, often updating internal state and stamping the right-hand-side contribution |
+| `MnaPostStep` | Component-specific update after the matrix solve, often reading the solution vector to update interface voltages and currents |
 
-Signal-domain components (regulators, governors, control blocks) usually define a `PreStep` that copies state from the previous step and a `Step` that updates the block outputs.
+This is a common pattern rather than a fixed rule; the exact work each task does is component-specific.
+Signal-domain components (regulators, governors, control blocks) define their own task list via `getTasks()`; many separate previous-step state handling from output updates, for example a `PreStep` that copies state from the previous step and a `Step` that updates the block outputs.
 
 The solver itself contributes a task that solves the MNA system; individual components do not depend on it by name, they depend on `leftVector` instead (see below).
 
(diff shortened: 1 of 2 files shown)


docs/hugo/content/en/docs/Overview/Scheduling/index.md
```diff
@@ -1,15 +1,234 @@
...
+## PreStep and PostStep conventions
+
+MNA components typically define two task classes per component:
+
+| Task | Typical responsibility |
+|------|----------------------|
+| `MnaPreStep` | Copy previous-timestep state; stamp the right-hand side before the matrix solve |
+| `MnaPostStep` | Extract interface voltage and current from the solution vector |
+
+Signal-domain components (regulators, governors, control blocks) usually define a `PreStep` that copies state from the previous step and a `Step` that updates the block outputs.
```
georgii-tishenin: Small wording suggestion: I think this section should be a bit less absolute, because `MnaPreStep` / `MnaPostStep` are task wrappers and the exact behavior is component-specific.

Maybe something like:

| Task | Typical responsibility |
|------|------------------------|
| `MnaPreStep` | Component-specific preparation before the matrix solve, often updating internal state and stamping the right-hand-side contribution |
| `MnaPostStep` | Component-specific update after the matrix solve, often reading the solution vector to update interface voltages/currents |

Similarly, for signal components, I would phrase it as a common pattern rather than a fixed convention: they define their own task list via `getTasks()`, and many control components separate previous-step state handling from output updates.

docs/hugo/content/en/docs/Overview/Scheduling/index.md
```diff
@@ -1,15 +1,234 @@
...
+────────────────────────
+ edge:  A → B   (A must run before B)
+```
+
+A special `Root` sentinel task is inserted as a sink for all `mPrevStepDependencies` entries.
+Its role is explained in the pruning step below.
+
+## Topological sort and pruning
+
+`Scheduler::topologicalSort()` first runs a backward BFS from `Root`, marking every task that transitively contributes to a simulation output.
```
georgii-tishenin: BFS = Breadth First Search? Maybe add the full form first time it's mentioned

docs/hugo/content/en/docs/Overview/Scheduling/index.md
```diff
@@ -1,15 +1,234 @@
...
+
+Signal-domain components (regulators, governors, control blocks) usually define a `PreStep` that copies state from the previous step and a `Step` that updates the block outputs.
+
+The solver itself contributes a task that solves the MNA system; individual components do not depend on it by name, they depend on `leftVector` instead (see below).
+
+---
+
+# Building the schedule
+
+## Task collection
```
georgii-tishenin, line 71: As I understand MNA and signal component tasks are collected within the solver. Here is my understanding:

## Task collection

`Simulation::prepSchedule()` collects all tasks before the first timestep from three top-level sources:

- **Solvers**: each solver contributes its task list via `Solver::getTasks()`. For MNA solvers, this includes:
  - the matrix-solve task,
  - MNA component pre-/post-step tasks from `MNASimPowerComp::mnaTasks()`,
  - signal-domain component tasks returned by `SimSignalComp::getTasks()`,
  - optional solver-side tasks, such as state-space extraction, when enabled.
- **Interfaces**: each interface contributes its own tasks via `Interface::getTasks()`. These tasks typically depend on the attributes exchanged with external systems.
- **Loggers**: each logger contributes a logging task via `Logger::getTask()`. Logger tasks depend on the logged attributes so that values are written after the producing tasks have run.

All tasks are placed in a flat `Task::List` and handed to the scheduler, which resolves the declared attribute dependencies into an execution order.

Does this make sense?

docs/hugo/content/en/docs/Overview/Scheduling/index.md
```diff
@@ -1,15 +1,234 @@
...
+The same constraint applies to simulation data recording: both the CSV logger (`DataLogger`) and the real-time data logger (`RealTimeDataLogger`) implement `DataLoggerInterface`, whose `logAttribute()` method only accepts `AttributeBase::Ptr`.
+The VILLASnode interface works the same way.
+Any value that needs to cross a task boundary, be recorded to a data file, or be exchanged with an external tool must be stored in an `Attribute<T>`.
+
+The component text logger (`CPS::Logger`, backed by spdlog) is a separate mechanism used for human-readable debug and diagnostic output.
+It is not part of the scheduling system and can print any value regardless of whether it is an attribute.
+
+For practical rules on when a variable should be an attribute versus a plain member variable, see [Attribute Usage Guidelines]({{< ref "../../Development/attribute-usage.md" >}}).
+
+## PreStep and PostStep conventions
```
georgii-tishenin: Small suggestion: this section is useful, but maybe we should make explicit that `PreStep`, `Step`, and `PostStep` are component/solver conventions rather than scheduler-level concepts. Renaming the section to something like “Common component task conventions” and adding one sentence to that effect could make this more clear.

docs/hugo/content/en/docs/Overview/Scheduling/index.md
```diff
@@ -1,15 +1,234 @@
...
+    attributeDependencies.push_back(leftVector);  // wait for matrix solve
+    modifiedAttributes.push_back(mIntfVoltage);
+    modifiedAttributes.push_back(mIntfCurrent);
+}
+```
+
+`PostStep` must always list `leftVector` in `attributeDependencies`.
+This creates the edge from the solver's matrix-solve task to every component's `PostStep`, ensuring the solution vector is available before voltages and currents are extracted.
+
+## Dependency declaration checklist
```
georgii-tishenin, line 233: I think this part is very useful. I suggest adding some concrete examples of common problems caused by incorrect configuration of pre/post -step dependencies:

- A `PreStep` or `PostStep` task is dropped completely because none of its
  declared modified attributes is needed by another task, logger, interface, or
  previous-step dependency.
- The same task appears to work only when a certain variable is logged or
  exchanged by an interface, because that logger/interface adds a dependency on
  the attribute and keeps the producing task reachable.

In the first case the simulation results are always wrong. In the second case the results depend on logger or interface configuration, even though the physical model did not change.

docs/hugo/content/en/docs/Overview/Scheduling/index.md
```diff
@@ -1,15 +1,234 @@
...
+
+---
+
+# Tasks
+
+## The Task base class
+
+Every task is an instance of a class that inherits from `CPS::Task`
+(`dpsim-models/include/dpsim-models/Task.h`).
+Each subclass implements one method:
```
georgii-tishenin: could it be better to use the word "function" in the context of C++?

docs/hugo/content/en/docs/Overview/Scheduling/index.md
```diff
@@ -1,15 +1,234 @@
...
+| `mAttributeDependencies` | Attributes this task *reads* in `execute()` |
+| `mModifiedAttributes` | Attributes this task *writes* in `execute()` |
+| `mPrevStepDependencies` | Attributes whose value from the *previous* timestep this task needs |
+
+All three lists hold `AttributeBase::Ptr` objects, the same pointers used throughout the component model.
+See [Attributes]({{< ref "../Attributes/index.md" >}}) for details on the attribute system.
+
+Only attributes can participate in scheduling.
+Plain C++ member variables (a `Real`, a `Matrix`, an internal state struct) are invisible to the scheduler, so no dependency edge can be formed around them.
+The same constraint applies to simulation data recording: both the CSV logger (`DataLogger`) and the real-time data logger (`RealTimeDataLogger`) implement `DataLoggerInterface`, whose `logAttribute()` method only accepts `AttributeBase::Ptr`.
```
georgii-tishenin: could it be better to use the word "function" in the context of C++?

georgii-tishenin (commented):
Hi @leonardocarreras. Thanks for your PR! I added some comments with optional suggestions for your consideration.

georgii-tishenin (approved):
Thanks for including the suggestions from comments!

PR #507, opened 2026-06-18
Bugfix for 2 minor issues in DiakopticsSolver:

### LINKED ISSUE #504: Diakoptics Solver: logged node voltages lag by one timestep / node voltages not logged at t=0
Found two small issues working on Diakoptics Solver

Bug 1 — DiakopticsSolver: node voltages not logged at t=0

dpsim/src/DiakopticsSolver.cpp initComponents

Bug Description: The MNA solver calls node->initialize() for all nodes during setup (MNASolver.cpp:160), copying mInitialVoltage into the v attribute so the t=0 sample is correct. DiakopticsSolver::initComponents never does this, leaving the v attributes at zero for the first logged sample.

Result: The t=0 logged sample reads 0 V; all samples from t0+1 onward are correct. Affects only the initial sample.
Proposed Fix: Call node->initialize() for all subnet nodes at the end of initComponents.

Bug 2 — DiakopticsSolver: logged node voltages lag by one timestep

dpsim/include/dpsim/DiakopticsSolver.h PostSolveTask

Bug Description: PostSolveTask updates node v attributes via mnaUpdateVoltage but declares only Scheduler::external as a modified attribute, not the node v attributes it actually writes. The scheduler runs the DataLogger before PostSolveTask, so each logged sample lags by one timestep.

Result: Every logged node voltage is shifted by exactly one timestep relative to the MNA solution.

Proposed fix: Declare each node's v attribute as a modified attribute of PostSolveTask.

FGH-wirtz:
- fix t=0 logging and one-timestep logger lag
- Call node->initialize() for all subnet nodes at the end of initComponents
- Declare each node's v attribute as a modified attribute of PostSolveTask

Closes #504

### DIFF
--- a/dpsim/include/dpsim/DiakopticsSolver.h
+++ b/dpsim/include/dpsim/DiakopticsSolver.h
@@ -168,6 +168,9 @@ public:
         : Task(solver.mName + ".PostSolve"), mSolver(solver) {
       for (auto &net : solver.mSubnets) {
         mAttributeDependencies.push_back(net.leftVector);
+        for (UInt node = 0; node < net.mRealNetNodeNum; ++node) {
+          mModifiedAttributes.push_back(net.nodes[node]->attribute("v"));
+        }
       }
       mModifiedAttributes.push_back(Scheduler::external);
     }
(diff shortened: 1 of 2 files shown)


leonardocarreras (approved):
Thanks for the PR @FGH-wirtz, looks good to me

leonardocarreras:
When you rebase, I will integrate it into the master branch

leonardocarreras:
@FGH-wirtz please use `git rebase master` and `git push --force-with-lease`, as we use linear history (do not merge master into your own branch)

PR #509, opened 2026-06-19
docs: add CONTRIBUTING.md and a notebook output compliance check

leonardocarreras:
Adds CONTRIBUTING.md at the repository root, where GitHub surfaces it when opening a pull request, and brings the Contribution and Development Guidelines pages in line with it.

A compliance workflow fails a pull request that changes a notebook still carrying saved outputs, scoped to the range the pull request touches and also available as a pre-commit hook. Linear history stays documented as a policy, since merge commits are a matter for the repository settings.

This is the base of a stack: #607 adds conventional commits on top, #608 adds release-please.

### DIFF
--- a/.github/workflows/compliance.yaml
+++ b/.github/workflows/compliance.yaml
@@ -1,0 +1,31 @@
+name: Compliance checks
+
+on:
+  pull_request:
+    branches: [master, main]
+
+jobs:
+  no-merge-commits:
+    name: No merge commits
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Fail if branch contains merge commits
+        run: |
+          BASE=${{ github.event.pull_request.base.sha }}
+          HEAD=${{ github.event.pull_request.head.sha }}
+          MERGES=$(git log --merges "$BASE".."$HEAD" --oneline)
+          if [ -n "$MERGES" ]; then
+            echo "Error: merge commits found in this branch:"
+            echo "$MERGES"
+            echo ""
+            echo "Please rebase your branch instead of merging the target branch:"
+            echo ""
+            echo "  git fetch upstream"
+            echo "  git rebase upstream/master"
+            echo "  git push --force-with-lease"
+            exit 1
+          fi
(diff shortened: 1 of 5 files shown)


docs/hugo/content/en/docs/Development/guidelines.md
```diff
@@ -51,6 +51,30 @@ There are no strict formal requirements besides the following:
...
+
+    ```bash
+    jupyter nbconvert --ClearOutputPreprocessor.enabled=True --inplace <notebook.ipynb>
+    ```
+
+    A `pre-commit` hook does this automatically once you have run `pre-commit install`; CI also rejects any notebook that still has outputs saved.
+
 # Creating New Releases (info for maintainers)
 
 DPsim currently uses to [Semantic Versioning](https://semver.org/). The periodic creation of
```
pipeacosta: Please correct the grammar of the statement: `DPsim currently uses to  [Semantic Versioning](https://semver.org/)`

pipeacosta (commented):
Thanks for your PR @leonardocarreras. The guidelines are useful and coherent. I just added a grammar check comment. In addition, a couple of suggestions from the LLM might make sense, e.g. the one that has to do with the nbconvert version and the merge commit detection.

.github/workflows/compliance.yaml
```diff
@@ -0,0 +1,68 @@
...
+# SPDX-License-Identifier: MPL-2.0
+
+name: Compliance checks
+
+on:
+  pull_request:
+    branches: [master, main]
+
+jobs:
+  no-merge-commits:
```
stv0g: I am curious: Doesnt GitHub already provide a feature to enforce a linear history via their Branch / PR rules?

leonardocarreras: I think it does, but we are still struggling to get more rights in the repo (so this is a workaround). So I cannot change most of the options on the Settings

leonardocarreras: Dropped

georgii-tishenin (dismissed):
Thanks @leonardocarreras. Good stuff! I think this should eventually make contributing and reviewing easier. From my point of view looks fine.
But we can see if @pipeacosta and @stv0g have further comments.

pipeacosta (approved):

leonardocarreras:
About the reviews/automated... We can have pinned commits for tools as a repo-wide change, but changing only for this one is wrong...

PR #500, opened 2026-06-17
Add FourTerminal_SSN classes / Fixed matpower.py import / Fixed PF Solver

matthiasmees:
## Added

- `FourTerminalVTypeSSN` class
- `GenericFourTerminalVTypeSSN` component
- Test notebook for SSN with transformer example
- Test notebook for two decoupled RL elements

## Fixed

- `matpower.py` import for case14 comparison
- Updated `PFSolver` classes; added warm-start feature for PF
- Missing `override` specifiers in `PFSolverPowerPolar`

PR #512, opened 2026-06-19
fix: PF Newton-Raphson globalization + matpower reader/notebook CI fixes

leonardocarreras:
Follow-up suggestion to #500. 

- **PF globalization:** replace #500's `alpha=0.6` damping + per-component clamps (linear convergence, forced `1e-6`/80 iters) with whole-step "max voltage change" scaling. Preserves Newton direction, restores `1e-8`/20-iter cap, drops the silent `V in [0.7,1.3]` clamp. Converges 4–5 iters on case14 + WSCC9.
- **matpower CI fixes (from #500):** `set_parameters` got a pandas Series → use `.values[0]`; case9 notebook `int("N1")` → strip the `N` prefix. Both notebooks now run end-to-end.


matthiasmees (dismissed):
All good

matthiasmees (approved):
All good

PR #513, opened 2026-06-19
Diakoptics development: KLU-based subnet solving with localized syste…

### LINKED ISSUE #431: Further development of the diakoptics approach
The diakoptics approach is interesting for avoiding excessive system matrices when simulating large grids. 

In some test investigations, this was particularly interesting for the refactorization step. For system matrix changes, local refactoring of the submatrix is sufficient in contrast to refactorization of the entire matrix. This has the potential to speed up the step enormously and is beneficial in addition to parallelization. This would therefore mainly benefit simulations of e.g. cascading fault events, where multiple recalculations are necessary and system matrixes can not be precalculated for every case.
The tests were implemented on an older dpsim version and do not take into account implementations of, for example, PARTIAL_REFACTORIZATION_METHOD in the KLU solver as a comparative value, which could already provide an alternative solution.

Proposed procedure:
- Test and check whether the same improvement in performance is evident and the approach remains interesting.
- Implement the diakoptics functionality on other components (e.g. 3ph).
- Add an example case with a high number of nodes to show benefits.

FGH-wirtz:
Diakoptics development: KLU-based subnet solving with localized system-matrix recomputation

Changes (DiakopticsSolver.cpp / h):
- Replaces dense per-subnet factorization (Eigen::PartialPivLU) with the sparse DirectLinearSolver (KLUAdapter) per subnet (Like for the monolithic MNA solver)
- Build the tear coupling from block-diagonal solves (Y_block^-1 * C) - dropped full dense mSystemInverse

Added:
- System-matrix recomputation for switching events: detect variable-component changes (hasVariableComponentChanged) and refactor only the subnetwork (recomputeSubnetMatrix) instead of the whole system
- Partial refactorization (KLU factorization path) of the changed subnet instead full refactorization
- Restrict the Schur-complement update to the tear columns coupled to the changed subnet
- EMT three-phase CIM diakoptics examples: EMT_IEEE_39bus_CIM.cpp, EMT_IEEE_39bus_Multi_CIM.cpp, EMT_Diakoptics.cpp
- Performance notebook EMT_IEEE_39bus_Multi_Diakoptics.ipynb: monolithic vs diakoptics (serial and OpenMP), node-voltage validation, average-step and switch-step timing analysis

Features:
- Topology changes (switch/fault) refactorize only the affected subnetwork block, not the full system matrix
- Uses Sparse KLU now: reuses the same DirectLinearSolver interface as MnaSolverDirect

Closes #431

### DIFF
--- a/dpsim/examples/cxx/CMakeLists.txt
+++ b/dpsim/examples/cxx/CMakeLists.txt
@@ -34,6 +34,7 @@ set(CIRCUIT_SOURCES
 	Circuits/EMT_VSI.cpp
 	Circuits/EMT_PiLine.cpp
 	Circuits/EMT_DecouplingLine_Ph3.cpp
+	Circuits/EMT_Diakoptics.cpp
 	Circuits/EMT_Ph3_R3C1L1CS1_RC_vs_SSN.cpp
 	Circuits/EMT_Ph3_RLC1VS1_RC_vs_SSN.cpp
 	Circuits/EMT_Ph1_General2TerminalSSN.cpp
@@ -220,6 +221,10 @@ if(WITH_CIM)
 		CIM/EMT_CIGRE_MV_withoutDG.cpp
 		CIM/EMT_CIGRE_MV_withDG.cpp
 		CIM/EMT_CIGRE_MV_withDG_withLoadStep.cpp
+
+		# IEEE 39 examples
+		CIM/EMT_IEEE_39bus_CIM.cpp
+		CIM/EMT_IEEE_39bus_Multi_CIM.cpp
 	)
 
 	list(APPEND TEST_SOURCES
@@ -228,6 +233,7 @@ if(WITH_CIM)
 		CIM/EMT_WSCC-9bus_IdealCS.cpp
 		CIM/EMT_WSCC-9bus_IdealVS.cpp
 		CIM/SP_WSCC9bus_SGReducedOrderVBR.cpp
+		CIM/EMT_IEEE_39bus_Multi_CIM.cpp
 	)
 
 	if(WITH_RT)
--- a/examples/Notebooks/Performance/EMT_IEEE_39bus_Multi_Diakoptics.ipynb
+++ b/examples/Notebooks/Performance/EMT_IEEE_39bus_Multi_Diakoptics.ipynb
@@ -1,0 +1,284 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "cell-0",
+   "metadata": {},
+   "source": [
+    "# EMT IEEE-39 Bus Multi-Area Diakoptics Runtime Measurement\n",
+    "\n",
+    "Loads four copies of the IEEE-39 bus system, merges them into one system and couples them in a ring. The `EMT_IEEE_39bus_Multi_CIM` example runs four variants and reports total time and the per-step solve times:\n",
+    "\n",
+    "- monolithic (single MNA system matrix)\n",
+    "- monolithic + OpenMP level scheduler\n",
+    "- diakoptics (ring couplers as tear components -> four subsystems)\n",
+    "- diakoptics + OpenMP level scheduler\n",
+    "\n",
+    "The coupling node voltages of the monolithic and diakoptics runs are compared to confirm both solvers agree and per-step solve times are analysed to compare the system-matrix recomputation triggered by the fault switch at t = 0.005 s."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "650eb188",
+   "metadata": {},
+   "source": [
+    "## Background: Why diakoptics, and what changed with KLU\n",
+    "\n",
+    "**Original Idea:** A large network is torn into subnetworks; the coupling is a small dense Schur complement. On a topology change (switch/fault) only the affected subnetwork block must be refactorized instead of the whole system matrix — a benefit even without parallelization (https://doi.org/10.1049/icp.2025.4299).\n",
+    "\n",
+    "**DPsim:** DPsim's KLU adapter now already supports partial refactorization, which the monolithic KLU solver uses. This also lets the monolithic solver refactorize cheaply at a switch.\n",
+    "\n",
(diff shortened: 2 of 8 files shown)


dpsim/examples/cxx/CMakeLists.txt
leonardocarreras, line 235: You probably also need to add them here
```cpp
# IEEE 39 examples
		CIM/EMT_IEEE_39bus_CIM.cpp
		CIM/EMT_IEEE_39bus_Multi_CIM.cpp
```

leonardocarreras (commented):
Thanks @FGH-wirtz for the PR. I just want to understand this a bit better, so we know how to continue. Additionally, you may want to add the test files to the list that can be executed from the notebooks (I think that is the cause of the CI error), and use pre-commit to format the files.

With diakoptics the grid is split into separate areas that solve their matrix on their own. I understand that in the parallel case, for example the OpenMP variant, these per-area solves run at the same time on different threads. The scheduler is okay with this because each one only writes its own separate part of the result.

The new recomputeSubnetMatrix seems to run inside those parallel solves (one per area), but it looks like it writes to shared data that all areas use. It seems to update mTearSchur and then rebuild the full mTotalTearImpedance. Because the areas are coupled and share columns, two threads could eventually write the same spot at once.

This likely only happens when a fault changes more than one area in the same step. The notebook faults just one area (I mean the IEEE39 case, area0), so only one area recomputes and the threads never collide, which is why CI will pass. Most of our other tests do not consider such cases at the moment.

I have some questions, none of them really blocking:
- Is this meant to work when a fault hits several areas in the same step, or is one area at a time the intended use? If it's one at a time, it would be good to note that somewhere.
- If several areas should work together, would it help to move the shared mTearSchur and mTotalTearImpedance update into one step that runs after the area solves, so only one thread touches it?

I'm a bit more inclined toward the all-areas solution, but would be good to know what you think (or correct me if I didn't get it right).

Just to be clear, I don't think the potential issue is a regression. The old dense code didn't recompute the matrix, so it also didn't support variable components. If you think the multi-area fix is better as a separate PR (or keeping it one area at a time for now), that's fine with me too.

FGH-wirtz:
Ah thanks for noticing. In my old version i actually had a mutex guard to prevent that race condition, which i skipped here.
I looked into it and think about moving Schur+LU to PreSolveTask, which would also solve this. Im implementing the test later.

examples/Notebooks/Performance/EMT_IEEE_39bus_Multi_Diakoptics.ipynb
```diff
@@ -0,0 +1,484 @@
...
...ZqkeCmZcuWqvZM+Pv7q0BNlq+pXr26bvmUZXZGjBhh1bGvvvqqqlGT90JEzomBGhE5BFmLz8wc3Lz44ouq9iyj1q1bW4I3sWTJEpQrV06tRbhr1y61JqAEeO7u7qqGTmq0ihUrhiZNmtxSA5aeno6tW7eqtQFlLcSqVavmKu/79+/HyZMnUaZMGdSqVStXz0FEjomBGhE5laxNn88//7yqbTt//rwKkiQwk+bJkSNH4vPPP1cLKMvCyu3bt1cLSptJYNWzZ0+VLl++PLZv364Wf54xYwbc3Kz712o0GlVftbVr16pAMCIiAkFBQSrI9PHxyacSICJ7wkCNiJzelStXVB8wCdjOnDmjAq+JEyeqfX5+fjh8+LCqLRs/frylxkv6mPXq1Qsffvih2o6Li0PDhg3x7bff4plnnrGqTKUG7++//1Y1d1JrZx5AkJyczECNiBROz0FETm/w4MEqSBOlS5dGSEgIhgwZooI0UaVKFRQuXBhHjx5V23v37sWOHTvUsbNnz8asWbOwaNEiVKhQAatWrbK6PL29vVWtmjR9mrVp0waBgYFOf06IyIQ1akTk9LIGRjJCNLt9MkhBnD59Wv2VoCxjvzVfX19UrFjR6vKU5tePP/5Y1c5JUNiuXTs89NBDqgmViIiBGhFRLphr36TZUwYi3IsXXngB48aNw549ezBnzhw12GHFihXqLxERmz6JiHKoUaNGqsbthx9+yLRf07Q7zuWWVVRUFFJSUtRI03r16uH9999H5cqVsWXLFp4TIlLY9ElElEMyInPSpEmqH5uMFm3bti0iIyPVaE2ZJ01GjFrj0KFDePTRRzFgwABVMyf93mQwQ7du3XhOiEhhjRoRORzppC/TXsg0G1llnfBWlmaSWqyMZPoOGRiQUZ8+fdQ8Z2b9+/dXgwokwFq3bp3qv/bjjz/eNUiT2jKZ5kNq31q1aqUGIcicbbKygtTSyUhQmRLEPAKUk90SOTeDJv8tiIgo340dOxZXr15V6d9//x0uLnf+rSxNoTJFiPjss88QHh7Os0TkZBioEREREdkoNn0SERER2SgGakREREQ2ioEaERERkY1ioEZERERkoxioEREREdkoBmpERERENoqBGhEREZGNYqBGREREZKMYqBERERHZKAZqRERERDaKgRoRERERbNP/AXZtO+ql8p06AAAAAElFTkSuQmCC",
```
leonardocarreras: Please clean the outputs before committing (please rewrite the commit)

FGH-wirtz:
ok there seems to be an issue using KLU and openMP. i can change the compile options: 
add_compile_options(--coverage -O0 -g -fprofile-update=atomic) 
to the cmake  like the Icov report proposed, but i would like your feedback before changing this.

leonardocarreras:
> ok there seems to be an issue using KLU and openMP. i can change the compile options: add_compile_options(--coverage -O0 -g -fprofile-update=atomic) to the cmake like the Icov report proposed, but i would like your feedback before changing this.

I think it should be okay, but we will not know until we run the CI again. we can evaluate after you change (if it slows down or brings a different error)

leonardocarreras (approved):
Looks good, thanks @FGH-wirtz  for the PR and for solving the comments!

PR #514, opened 2026-06-19
feat: sparse-Jacobian powerflow solver with KLU refactorization reuse for large grids

leonardocarreras:
## What

Adds `PFSolverPowerPolarSparse`: a Newton-Raphson powerflow solver that assembles the Jacobian sparse (fixed pattern, values updated in place) and routes the linear solve through DPsim's `DirectLinearSolver` (KLU). The symbolic factorization is analyzed **once**; each iteration refactorizes (reusing ordering + symbolic), replacing the dense `mJ` + a fresh `Eigen::SparseLU`-per-iteration that does not scale past ~1000 buses.

The dense `PFSolverPowerPolar` and the #512 convergence logic are untouched. The two solvers are selectable on `Simulation` via `set_pf_solver_use_sparse(bool)` (default **false**: the dense solver remains the default and the sparse solver is opt-in). When no sparse linear solver is compiled in, the flag is ignored and the dense solver is used.

## How

- `PFSolver` (base): two virtual hooks, `setUpJacobianStorage()` and `solveJacobianSystem()`; the defaults reproduce the old dense path byte-for-byte.
- `PFSolverPowerPolarSparse`: builds the sparsity pattern from the bus-admittance connectivity, fills J1..J4 in place each iteration, `factorize()` on the first iteration of each run (full pivoting) then `refactorize()` afterwards. Uses full refactorization, not `partialRefactorize` (correct for NR, where most Jacobian values change each iteration).
- `Simulation`: `mPFSolverUseSparse` attribute (default false) + `set/get_pf_solver_use_sparse` (C++ and pybind).

## Docs

- `docs/.../Concepts/powerflow.md` now documents the two solver implementations and how to select them, notes that DPsim implements only the power-mismatch + polar-coordinate formulation (of the four listed), reconciles the textbook Jacobian with the implementation's relative voltage-increment scaling, and adds a "Convergence and Step Control" section (tolerance, max iterations, and the whole-step globalization bounds).
- The benchmark notebook's intro reflects dense-as-default / sparse opt-in.

## Verification

Converged voltages match the dense solver to round-off, with a growing speedup vs network size, but already noticeable for small networks:

| case | buses | dense [s] | sparse [s] | speedup | rel. diff |
|---|---|---|---|---|---|
| case9 | 9 | 0.0036 | 0.0026 | 1.38x | 2.4e-16 |
| case14 | 14 | 0.0072 | 0.0054 | 1.35x | 2.0e-16 |
| case145 | 145 | 0.871 | 0.613 | 1.42x | 3.8e-16 |
| case300 | 300 | 2.141 | 1.484 | 1.44x | 1.6e-16 |

New notebook `examples/Notebooks/Grids/PF_Sparse_vs_Dense.ipynb` runs this sweep both ways, asserts the voltages agree (relative `< 1e-8`), and plots run time vs network size.

### DIFF
--- a/dpsim/include/dpsim/PFSolverPowerPolarSparse.h
+++ b/dpsim/include/dpsim/PFSolverPowerPolarSparse.h
@@ -1,0 +1,40 @@
+// SPDX-FileCopyrightText: 2026 Institute for Automation of Complex Power Systems, EONERC, RWTH Aachen University
+// SPDX-License-Identifier: MPL-2.0
+
+#pragma once
+
+#include <dpsim/DirectLinearSolver.h>
+#include <dpsim/PFSolverPowerPolar.h>
+
+namespace DPsim {
+/// Polar-coordinate powerflow solver that assembles a sparse Jacobian and reuses
+/// the symbolic factorization across Newton-Raphson iterations (scales to large grids).
+class PFSolverPowerPolarSparse : public PFSolverPowerPolar {
+protected:
+  /// Sparse Jacobian (row-major, fixed pattern, values updated in place each iteration)
+  CPS::SparseMatrixRow mJsparse;
+  /// Direct linear solver reused across iterations (symbolic factorization computed once)
+  std::shared_ptr<DirectLinearSolver> mLinearSolver;
+  /// Empty list: PF uses full refactorization, not partial refactorization
+  std::vector<std::pair<CPS::UInt, CPS::UInt>> mVariableSystemMatrixEntries;
+
+  /// Build the fixed sparsity pattern, create the linear solver and analyze the pattern once
+  void setUpJacobianStorage() override;
+  /// Fill the sparse Jacobian values in place (pattern unchanged)
+  void calculateJacobian() override;
+  /// Factorize (first iteration of a run) or refactorize, then solve
+  void solveJacobianSystem() override;
+
+  /// Construct the structural sparsity pattern of the Jacobian from the bus admittance matrix
+  void buildJacobianPattern();
+  /// Whether buses k and j are connected (off-diagonal Jacobian entry exists)
+  bool isConnected(CPS::UInt k, CPS::UInt j);
+
+public:
+  /// Constructor to be used in simulation examples.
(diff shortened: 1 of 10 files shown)


dpsim/include/dpsim/PFSolverPowerPolarSparse.h
```diff
@@ -0,0 +1,45 @@
+/* Copyright 2017-2021 Institute for Automation of Complex Power Systems,
```
georgii-tishenin: We now have a new format for those copyright messages right?

georgii-tishenin (dismissed):
Hi @leonardocarreras. Thanks for the PR! I left a small comment.

This looks good to me. One question on future plans: sparse solver seems to perform well even for small networks,what is the intended long-term role of the dense PFSolverPowerPolar? Should the dense implementation remain as the default/reference solver for benchmarking and fallback purposes, or is the plan to eventually make the sparse solver the default once it has more coverage?

dpsim/include/dpsim/PFSolverPowerPolarSparse.h
leonardocarreras: Yes, thanks for that, I just copied the other existing, started to make changes and forgot

leonardocarreras:
> Hi @leonardocarreras. Thanks for the PR! I left a small comment.
> 
> This looks good to me. One question on future plans: sparse solver seems to perform well even for small networks,what is the intended long-term role of the dense PFSolverPowerPolar? Should the dense implementation remain as the default/reference solver for benchmarking and fallback purposes, or is the plan to eventually make the sparse solver the default once it has more coverage?

I want to make this the default, but do not want to introduce a breaking change without testing. I would recommend to have some time for testing, and because the initialization depends on this it may make some simulations fail silently for people. In my initial tests, there are no meaningful differences in the notebooks (all are very very small), except speed up.

dpsim/include/dpsim/PFSolverPowerPolarSparse.h
leonardocarreras: Done :)

georgii-tishenin (approved):

PR #515, opened 2026-06-22
DP/SFA SSN: base classes and single-phase serial RLC component

leonardocarreras:
Adds the State-Space Nodal (SSN) method to the DP (dynamic phasor / shifted-frequency) domain, mirroring the landed EMT-SSN hierarchy.

## New additions
- `CPS::DP::SSNComp` base + `VTypeSSNComp` / `ITypeSSNComp`: real-augmented carrier shift `A_aug=[[A,ωI],[-ωI,A]]`, discretized with the existing trapezoidal helper, folded to a complex envelope Norton model.
- `DP::Ph1::TwoTerminalVTypeSSNComp` single-phase two-terminal stamping layer.
- `DP::Ph1::SSN::Full_Serial_RLC` reference component with Python binding and a C++ example.

## Validation
Cross-domain (SP / DP / EMT / DP-SSN / EMT-SSN) on a series RLC: DP-SSN reproduces classical DP and matches the EMT waveform to ~1e-6 (steady and off-nominal excitation).

### Obs
Base-class design needs to be coordinated with @georgii-tishenin and @matthiasmees  to keep the EMT and DP SSN approaches aligned.

PR #516, opened 2026-06-22
DP/SFA SSN: generic two-terminal V/I-type components

leonardocarreras:
Adds user-parametrizable single-phase SSN one-ports to the DP domain: declare any linear two-terminal component by its state-space `(A, B, C, D)` and simulate it. Builds on the DP SSN base hierarchy in #515.

## New additions
- `DP::Ph1::TwoTerminalITypeSSNComp`: single-phase I-type stamping layer (Norton-equivalent admittance `G = W^-1`, history current, complex node read). The V-type stamping layer already existed; this adds its I-type counterpart.
- `DP::Ph1::GenericTwoTerminalVTypeSSN` / `GenericTwoTerminalITypeSSN`: concrete components parametrized directly by `(A, B, C, D)`, with Python bindings (`dp.ph1.GenericTwoTerminalVTypeSSN` / `...ITypeSSN`, `x` exposed read-only).
- C++ example `DP_Ph1_General2TerminalSSN.cpp`.

## Validation
On the example circuits, the generic SSN one-ports reproduce equivalent discrete-component references.

The I-type case exercises the `W^-1` inversion path for the first time in the DP domain.

## Obs
- Should be merged after #515.
- Cross domain notebook validation as follow-up.

### DIFF
--- a/dpsim-models/include/dpsim-models/Components.h
+++ b/dpsim-models/include/dpsim-models/Components.h
@@ -59,6 +59,8 @@
 #include <dpsim-models/DP/DP_ITypeSSNComp.h>
 #include <dpsim-models/DP/DP_Ph1_ControlledCurrentSource.h>
 #include <dpsim-models/DP/DP_Ph1_ControlledVoltageSource.h>
+#include <dpsim-models/DP/DP_Ph1_GenericTwoTerminalITypeSSN.h>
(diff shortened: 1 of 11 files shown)


georgii-tishenin (approved):
Hi @leonardocarreras. Thanks for the PR! 
Again, this looks consistent for me with EMT SSN and DP conventions. Let's continue!

PR #517, opened 2026-06-22
DP/SFA SSN: single-phase validation and accuracy notebooks

leonardocarreras:
Adds two Python notebooks and a documentation page for the single-phase DP SSN components, on top of the generic V/I-type work in #516 (now merged).

## Notebooks

`DP_generalizedSSN_RLC.ipynb` (validation). A series RLC across DP and EMT, classical vs SSN: the dedicated `Full_Serial_RLC` and the generic `GenericTwoTerminalVTypeSSN` / `GenericTwoTerminalITypeSSN` reproduce the classical DP envelopes, and the DP-SSN envelope reconstructed to the time domain matches the EMT and EMT-SSN waveforms.

`DP_SSN_RLC_accuracy.ipynb` (accuracy study). Steady-state phasor error against a small-step EMT reference, swept over time step and signal frequency, characterising the shifted-frequency accuracy of DP-SSN relative to EMT-SSN.

## Documentation

Adds `docs/hugo/content/en/docs/Concepts/state-space-nodal.md`, a concept page for State-Space Nodal in the dynamic phasor domain: the SSN method, the real-augmented SFA frequency shift, the available components (`Full_Serial_RLC`, generic V/I-type), and a validation-and-examples section pointing to the two notebooks. Includes references to the foundational SSN and frequency-warping literature.

## Notes

Builds on #516 (generic two-terminal V/I-type components)

### DIFF
--- a/docs/hugo/content/en/docs/Concepts/state-space-nodal.md
+++ b/docs/hugo/content/en/docs/Concepts/state-space-nodal.md
@@ -1,0 +1,101 @@
+---
+title: "State-Space Nodal"
+linkTitle: "State-Space Nodal"
+date: 2026-06-23
+---
+
+The state-space nodal (SSN) method represents a component by its own continuous
+state-space model and couples it to the network through the nodal admittance
+matrix. A component is described by
+
+```math
+\frac{d\mathbf{x}}{dt} = \mathbf{A}\mathbf{x} + \mathbf{B}\mathbf{u},
+\qquad
+\mathbf{y} = \mathbf{C}\mathbf{x} + \mathbf{D}\mathbf{u},
+```
+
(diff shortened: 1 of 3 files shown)


georgii-tishenin (approved):
Thanks for the PR! Looks good to me: the notebooks include plausible assertions and results interpretations.

Also thanks for adding the documentation on SSN. I think it's already good. Here are some discussion points for the future:
1. My understanding is that section "Overview" is DPsim-implementation-specific, while section "Concepts" is purely about theoretical concepts. I'd suggest moving DPsim-specific details like using real augmented matrix to "Overview" or just removing. But this is maybe just my understanding.
2. As a follow-up I would also add the list of existing EMT SSN components.

PR #519, opened 2026-06-26
fix: SP::Ph1::Capacitor produces NaN admittance for zero capacitance

leonardocarreras:
## Problem
A capacitor's admittance is jwC, but the code computed 1/(1/jwC), which is NaN at C = 0. One zero-C capacitor then poisons the Y diagonal and breaks the power flow (hit in practice via a transformer snubber with no rated power).

## Fix 
Take admittance straight from the susceptance. Identical for C != 0; C = 0 now gives 0, so existing results are unchanged (WSCC-9 unaffected). Also drops the dead isinf/cout guard (isinf is folded to false under -Ofast), replacing it with a logger check on a new shared Math::isFinite helper (bit-pattern test, so it stays correct under -ffast-math) that other PF stamps can reuse.

### DIFF
--- a/dpsim-models/include/dpsim-models/MathUtils.h
+++ b/dpsim-models/include/dpsim-models/MathUtils.h
@@ -34,6 +34,11 @@ public:
   static Complex polar(Real abs, Real phase);
   static Complex polarDeg(Real abs, Real phase);
 
+  // -Ofast/-ffast-math folds std::isnan/isfinite to false; test the IEEE-754
+  // exponent bits directly instead.
+  static bool isFinite(Real value);
+  static bool isFinite(Complex value);
+
   // #### Vector Operations ####
   //
   // | Re(row,0)_harm1 | Re(row,colOffset)_harm1 |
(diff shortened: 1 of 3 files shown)


georgii-tishenin (approved):
Looks good to me.

Small cleanup question: after this change, do we still need `mImpedance` and `mImpedancePerUnit` in `SP::Ph1::Capacitor` at all?

For stamping, the component only needs `mSusceptance` / `mAdmittancePerUnit`, and the impedance values seem to be used only for the old admittance calculation and logging. If they are not part of any intended API, maybe they could be removed in a follow-up cleanup?

leonardocarreras:
> Looks good to me.
> 
> Small cleanup question: after this change, do we still need `mImpedance` and `mImpedancePerUnit` in `SP::Ph1::Capacitor` at all?
> 
> For stamping, the component only needs `mSusceptance` / `mAdmittancePerUnit`, and the impedance values seem to be used only for the old admittance calculation and logging. If they are not part of any intended API, maybe they could be removed in a follow-up cleanup?

I agree, and it likely is the case for other components.

PR #520, opened 2026-06-26
fix: guard PF convergence check against non-finite mismatch

leonardocarreras:
`checkConvergence()` tested `abs(mF(i)) > tol`, but `abs(NaN) > tol` is `false` in IEEE-754, so a NaN mismatch made the solver report convergence after one iteration and hide a broken solve.

Fix: treat any non-finite `mF(i)` as not converged (return `false`, log a warning). The finiteness test is a bit-pattern check, since the project's `-Ofast` build folds `std::isnan` to false.

Needs #519 merged first to use the helper function.

### DIFF
--- a/dpsim/src/PFSolver.cpp
+++ b/dpsim/src/PFSolver.cpp
@@ -6,6 +6,7 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  *********************************************************************************/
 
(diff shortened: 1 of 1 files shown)


georgii-tishenin (approved):
Thanks @leonardocarreras. The fix looks good to me.

PR #521, opened 2026-06-26
fix: refresh matrix-node-index cache after terminal swap in SP Transformer PF

leonardocarreras:
`SP::Ph1::Transformer` swaps its terminals when `|ratio| < 1` to keep the HV side at position 0, but the `mMatrixNodeIndices` cache was populated before the swap. As a result `pfApplyAdmittanceMatrixStamp` stamped admittances at the wrong node indices.

Fix: call `updateMatrixNodeIndices()` at the end of the terminal-swap block so the power-flow stamp uses the corrected ordering.


georgii-tishenin (approved):
Hi, the fix looks reasonable.

PR #524, opened 2026-06-29
ci: run all Linux jobs on self-hosted runner

leonardocarreras:
Move all Linux GitHub Actions jobs to a self-hosted runner. 
Windows jobs stay on GitHub runner.


leonardocarreras:
@SystemsPurge seems to work, seems to be slow :) (tests 49m vs approx 30m, 50% slower)

SystemsPurge:
@leonardocarreras Is this the notebook tests?
If so, could that maybe be because of the added timeout in spots where requests are likely to fail?
Otherwise we could compare single test durations , as they seem to be in the logs

leonardocarreras:
@SystemsPurge yes, maybe is that... We can test a bit more, it can also be that too many process are being allowed in the one running the tests

leonardocarreras:
closing for the moment

PR #525, opened 2026-06-29
SP power-flow: remove dead isinf guards and unused capacitor fields

leonardocarreras:
## What

Cleanup of dead NaN/inf-guard code in the SP power-flow admittance stamps.

Under `-Ofast`/`-ffast-math` the compiler folds `std::isnan`/`std::isinf`/`std::isfinite` to a compile-time `false`, so every `std::isinf` finite-guard in the SP tree was silently dead. This replaces them with the bit-pattern `CPS::Math::isFinite`, which actually works in this compilation conditions.

### Capacitor (follow-up to Georgii's review on #519)
- Remove dead `mImpedance` / `mImpedancePerUnit` fields and their computations/log lines. The admittance path uses `mSusceptance` directly; the impedance round-trip was unused.

### Dead-guard sweep (implement pattern)
Same dead `std::isinf` + paired `std::cout` pattern across 6 more SP components, each rerouted through `SPDLOG_LOGGER_ERROR` + `throw InvalidArgumentException()` and the `std::cout` dropped:
- `SP_Ph1_Resistor`, `SP_Ph1_Shunt`, `SP_Ph1_PiLine`, `SP_Ph1_RXLine`, `SP_Ph1_Transformer`, `SP_Ph1_SolidStateTransformer`

### DIFF
--- a/dpsim-models/include/dpsim-models/SP/SP_Ph1_Capacitor.h
+++ b/dpsim-models/include/dpsim-models/SP/SP_Ph1_Capacitor.h
@@ -33,13 +33,9 @@ protected:
   /// base current [A]
   Real mBaseCurrent;
 
-  /// Impedance [pu]
-  Complex mImpedancePerUnit;
   /// Admittance [pu]
   Complex mAdmittancePerUnit;
 
-  /// Impedance [Ohm]
-  Complex mImpedance;
   /// Admittance [S]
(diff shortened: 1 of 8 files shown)


georgii-tishenin (approved):
Thanks @leonardocarreras. Looks good to me.

PR #532, opened 2026-06-29
SP transformer: fix snubber NaN from missing or invalid rated power

leonardocarreras:
Transformers built without a rated power keep `mRatedPower = 0`, so snubbers size to infinite R and zero C, stamping a non-finite admittance into the PF matrix (NaN on matpower-sourced grids).

- `SP_Ph1_Transformer`: skip snubbers when `mRatedPower <= 0` (WARN); reject negative rated power (`|S|` is a magnitude); fix log unit `[W]` -> `[VA]`.
- `matpower.py`: size rated power from branch `rateA`, fall back to system base only when `rateA == 0`.


georgii-tishenin (approved):
Looks good to me. 

The fix addresses the realistic failure mode: missing/unset transformer rated power leads to invalid snubber sizing, so disabling snubbers for non-positive ratings is a reasonable fallback. Maybe one question for the future is: do we really need snubbers at all? I think typical low frequency transformer models do not model capacitances.


The MATPOWER rateA handling also makes sense, with system base power only used for unrated branches. One possible suggestion for the future is to log when the system base power is taken (I think now there is no logging done in matpower.py at all)

Issue #301, opened 2024-06-07
CI Flow does not work for Dockerfile changes


n-eiling:
Because the CI does not build the dev container but instead uses one that has to be build manually, we cannot properly check changes to the Dockerfile.

I feel like the switch to the github actions really was a regression in many ways. With the gitlab runner we used to build the containers for each PR separately and were still faster than github.

Has anyone somehow tested changes to the dockerfiles without just pushing a new image to the registry and hoping for the best?

georgii-tishenin:
Hi, @n-eiling, I am pretty new to this topic and have not yet made any changes to the dockerfiles. Do you already have some ideas to improve on testing the changes of the Dockerfile? E.g., can the dev container build be automated?

m-mirz:
As @n-eiling mentioned, we already had this automated but building the container images in the ci takes forever since we only use free resources in GitHub.

leonardocarreras:
@stv0g apparently this is something that we will need to think now [due to new pricing scheme](https://docs.github.com/en/billing/reference/actions-runner-pricing) -> Update: a colleague says that is **still going to be free for public repos**

stv0g:
Haha, I was about to commment on exactly that :)

stlankes:
Yes, we can do it. I will talk directly to @leonardocarreras

Ace2932:
Whichever way the self-hosted decision lands, here's a free win independent of it: none of the
build workflows (Fedora, Windows, RockyLinux, Nix, SonarCloud, PyPI) set a concurrency group,
so a re-push keeps the previous run building the whole matrix to completion. That's ~950
runner-minutes a month of pure superseded-run waste across those workflows. One block per
workflow cancels it:

  concurrency:
    group: ${{ github.workflow }}-${{ github.ref }}
    cancel-in-progress: true

(The PyPI publish workflow also runs on both push and pull_request, so it double-fires on PR
branches, ~another 100 min/mo.)

Numbers came from a scanner I built, fwiw: https://gitspider.com/scan/sogno-platform/dpsim?ref=ghissue

leonardocarreras:
Hi @Ace2932, thanks for the suggestion! By the way, cool toolset!

Ace2932:
Thanks, that means a lot, glad the concurrency tip helps! If you ever want it watching dpsim's
CI continuously there's a free GitHub App: the dashboard trends new
regressions, flaky tests, and waste over time, vs the one-off snapshot you saw. The concurrency
[...]

PR #534, opened 2026-06-30
ci: cancel superseded workflow runs via concurrency groups

leonardocarreras:
Adds a `concurrency` group to the PR-triggered build workflows (Fedora, Fedora Minimal, RockyLinux, Windows, Nix, SonarCloud, PyPI) so a re-push cancels the in-progress run instead of letting the whole matrix finish.

Group is keyed on `${{ github.workflow }}-${{ github.head_ref || github.ref }}`, so each PR (and each branch) only cancels its own superseded runs.

Picks up the suggestion from #301, thanks @Ace2932.

### DIFF
--- a/.github/workflows/build_nix.yaml
+++ b/.github/workflows/build_nix.yaml
@@ -7,7 +7,7 @@ on:
   pull_request:
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 ## Build ##
--- a/.github/workflows/build_nix.yaml
+++ b/.github/workflows/build_nix.yaml
@@ -6,6 +6,10 @@ on:
       - master
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
 ## Build ##
 
 jobs:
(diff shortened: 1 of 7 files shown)


georgii-tishenin (dismissed):
Hi @leonardocarreras. Thanks for the PR. We can try this out.
One small question: is there any relevant risk that two different PRs could share the same concurrency group if they come from different forks but use the same source branch name?

As far as I understand, `github.head_ref` is the PR source branch name. This should normally be fine, but in the edge case of two PRs from different forks both using a common branch name like `fix-ci` or `main`, they might end up with the same group for the same workflow and cancel each other’s runs.

Not a blocker from my side, since the practical risk seems low, but I wanted to ask whether this edge case is worth considering.

leonardocarreras:
> Hi @leonardocarreras. Thanks for the PR. We can try this out. One small question: is there any relevant risk that two different PRs could share the same concurrency group if they come from different forks but use the same source branch name?
> 
> As far as I understand, `github.head_ref` is the PR source branch name. This should normally be fine, but in the edge case of two PRs from different forks both using a common branch name like `fix-ci` or `main`, they might end up with the same group for the same workflow and cancel each other’s runs.
> 
> Not a blocker from my side, since the practical risk seems low, but I wanted to ask whether this edge case is worth considering.

I made changes to mitigate this. Now switched the group key to `github.event.pull_request.number`

georgii-tishenin (approved):
Thanks for addressing the edge case. Let's try that out!

PR #535, opened 2026-07-01
SP SynchronGenerator: add reactive-power limits (Qmin/Qmax)

leonardocarreras:
## Summary
- Add `mQLimMax`/`mQLimMin` (default ±infinity, i.e. unlimited) to `SP::Ph1::SynchronGenerator`
- New `setParameters` overload with optional `qLimMax`/`qLimMin`, exposed to Python as keyword args `q_limit_max`/`q_limit_min`
- No behavioral change as generators with no limits set (the default) are unaffected

This is a foundation commit for a follow-up PR that adds PV<->PQ outer-loop enforcement of these limits in `PFSolver`.

### DIFF
--- a/dpsim-models/include/dpsim-models/SP/SP_Ph1_SynchronGenerator.h
+++ b/dpsim-models/include/dpsim-models/SP/SP_Ph1_SynchronGenerator.h
@@ -8,6 +8,8 @@
 
 #pragma once
 
+#include <limits>
+
 #include <dpsim-models/MNASimPowerComp.h>
 #include <dpsim-models/Solver/PFSolverInterfaceBus.h>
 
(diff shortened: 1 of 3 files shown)


georgii-tishenin (approved):
Thanks @leonardocarreras. This looks good to me.

Maybe some questins (I do not have an immediate answer myself):

1.  is this justified to have these new variables as attributes? This definitely follows the existing conventions in SP_Ph1_SynchronGenerator, but could it be not necessary?
2. could any of these new variables (e.g. limits) be private?

leonardocarreras:
> Thanks @leonardocarreras. This looks good to me.
> 
> Maybe some questins (I do not have an immediate answer myself):
> 
> 1. is this justified to have these new variables as attributes? This definitely follows the existing conventions in SP_Ph1_SynchronGenerator, but could it be not necessary?
> 2. could any of these new variables (e.g. limits) be private?

Hi, I agree with the questions.
1. I just decided to follow existing ways for the moment, but yes is not optimal.
2. Yes, and also there are variables that depend from each other (quantity vs per-unit quantity). I think this is a bigger refactor. I will make an issue for usage of attributes, in general.

PR #536, opened 2026-07-01
PFSolver: enforce generator Q-limits via PV<->PQ outer loop

leonardocarreras:
## Summary
- Bidirectional PV<->PQ outer loop in `PFSolver`: a PV bus whose generator Q exceeds Qmin/Qmax is pinned at the limit and converted to PQ, reverting to PV once the constraint relaxes. Bounded to prevent oscillation.
- Opt-in via `set_pf_solver_enforce_q_limits(bool)`, default off, then no change expected to existing cases.
- Adds `PF_Generator_Qlimits.ipynb`: verifies pinning, PV-hold, and dense == sparse agreement.
- Adds a short documentation.

Depends on #535.


dpsim/src/PFSolver.cpp
```diff
@@ -515,6 +551,24 @@ Bool PFSolver::solvePowerflow() {
...
+  // Q-limit outer loop: after each converged inner solve, switch PV buses that
+  // violated their reactive limits to PQ (pinned at the limit) and pinned buses
+  // whose constraint relaxed back to PV; re-solve until no bus switches.
+  for (CPS::UInt outer = 0; converged && outer < mMaxOuterIterations; ++outer) {
+    if (!enforceReactiveLimits())
+      break; // all generators within their reactive limits
+    reclassifyBuses();
+    converged = runNewtonRaphson();
+  }
+  return converged;
```
georgii-tishenin, line 583: Small logging question: should we log a warning if the Q-limit outer loop exits because `mMaxOuterIterations` is reached?

In that case the last inner Newton solve may have converged, so `solvePowerflow()` returns true, but the PV/PQ switching loop may not have finished with a final "no bus switched" pass. I am not suggesting changing the algorithm, but a warning could make this edge case easier to diagnose.

dpsim/src/PFSolverPowerPolar.cpp
```diff
@@ -513,6 +521,135 @@ void PFSolverPowerPolar::calculateQAtPVBuses() {
...
+  for (auto &c : toPQ) {
+    auto node = std::get<0>(c);
+    bool atMax = std::get<1>(c);
+    CPS::Real qLimPU = std::get<2>(c);
+    mPVBuses.erase(std::remove(mPVBuses.begin(), mPVBuses.end(), node),
+                   mPVBuses.end());
+    mPQBuses.push_back(node);
+    UInt idx = node->matrixNodeIndex();
+    Qesp(idx) = qLimPU - loadReactivePowerPerUnit(node);
+    sol_Q(idx) = Qesp(idx);
```
georgii-tishenin, line 616: Small consistency question: when a PV generator is converted to PQ and pinned at Qmax/Qmin here, should the generator's own `Q_set` / `Q_set_pu` also be updated to the active limit?

As I understand it, the bus/network injection is pinned through `Qesp` / `sol_Q`, but this generator will no longer be handled by `calculateQAtPVBuses()` because it has been moved out of `mPVBuses`.

georgii-tishenin (commented):
Hi @leonardocarreras. Thanks for the PR. I left two small questions in the comments. Please, have a look if they're relevant before we merge.

dpsim/src/PFSolver.cpp
leonardocarreras: Fixing the causes, logging was not enough. Addressing partially here, partially in #543

dpsim/src/PFSolverPowerPolar.cpp
leonardocarreras: Fixing. Addressing partially here, partially in #543

georgii-tishenin (dismissed):
Thanks for the updates. From my side, this PR is a useful improvement over the current behavior.

I have one remaining non-blocking usability concern. Looking again at my previous comment, I think I may have suggested a questionable solution: updating the generator Q_set / Q_set_pu when a PV bus is converted to PQ. This on the one hand makes the converted generator consistent with the pinned Q-limit result, but Q_set sounds like a user-defined input/setpoint. I am now questioning: is it fine to rewrite the value defined by the user?

leonardocarreras:
> Thanks for the updates. From my side, this PR is a useful improvement over the current behavior.
> 
> I have one remaining non-blocking usability concern. Looking again at my previous comment, I think I may have suggested a questionable solution: updating the generator Q_set / Q_set_pu when a PV bus is converted to PQ. This on the one hand makes the converted generator consistent with the pinned Q-limit result, but Q_set sounds like a user-defined input/setpoint. I am now questioning: is it fine to rewrite the value defined by the user?

I did avoid the rewriting. I am planning to support multiple generators per bus (and split the power among them), so this can be handled better that way at that point.

Will require your review again :)

georgii-tishenin (approved):
Thanks, I checked the latest change. Removing the updates to `Q_set` / `Q_set_pu` addresses my concern. Approved.

leonardocarreras:
PR #562 is solving the test failures, not related to this one.

PR #537, opened 2026-07-01
Add participation factors to state-space modal analysis

georgii-tishenin:
This PR extends state-space modal analysis with right/left eigenvectors, participation factors, and state names in the selected analysis frame.

Main changes:

* add named state metadata to MNA state-space extraction,
* compute right eigenvectors, left eigenvectors, and participation factors in `StateSpaceModalAnalysis`,
* expose modal participation results to Python,
* update the inverter state-space extraction notebook with a `GlobalDQ0` participation-factor check against the analytical discrete dq0 model.


leonardocarreras (approved):
Looks good to me in general.

This is more like a general question... Should we use CPS Exceptions? In this particular case, the standard one seems to provide more information. The only thing is that this is not written in a logger.

georgii-tishenin:
> Looks good to me in general.
> 
> This is more like a general question... Should we use CPS Exceptions? In this particular case, the standard one seems to provide more information. The only thing is that this is not written in a logger.

Thanks, @leonardocarreras. I agree with your comment, I opened a discussion #539 to decide on the harmonized approach for exceptions.

PR #538, opened 2026-07-02
DP_Ph3: add CurrentSource, fix per-phase source signals

leonardocarreras:
## Summary
- Add `DP::Ph3::CurrentSource`, the last missing basic DP_Ph3 component needed to port the EMT_Ph3 SSN validation reference circuits to the DP domain.
- Fix `DP::Ph3::VoltageSource` (and the new `CurrentSource`) to take an independent per-phase `MatrixComp` reference instead of a single `Complex` value rotated by a hardcoded +-120 degrees, which could only represent a balanced/symmetric source.
- Add cxx example `DP_Ph3_R3C1L1CS1.cpp` exercising the new component.

### DIFF
--- a/dpsim-models/include/dpsim-models/Components.h
+++ b/dpsim-models/include/dpsim-models/Components.h
@@ -86,6 +86,7 @@
 #include <dpsim-models/DP/DP_VTypeSSNComp.h>
 
 #include <dpsim-models/DP/DP_Ph3_Capacitor.h>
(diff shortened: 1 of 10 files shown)


georgii-tishenin (approved):
Hi @leonardocarreras, thanks for the PR! Looks good to me.

The new `MatrixComp` parameter is definitely more general and needed for independent per-phase sources. In some other case it could be considered a breaking change of public API, but I think for DP_Ph3 components this is fine as they were not really used/tested and even not exposed in Python.

PR #543, opened 2026-07-06
PFSolver: zone-based base voltage verification, configurable tolerances, and matpower Q-limit fixes

leonardocarreras:
## Summary
- Base voltage is verified per electrical zone: nodes joined by a line share a voltage level, transformers are boundaries. Mismatches throw.
- Generator/Transformer/NetworkInjection/VSI ratings are authoritative, checked with a strict tolerance. Load and Shunt are checked against the zone's rating with a looser tolerance.
- VSI's base voltage comes from the connection transformer's grid-side rating.
- Both tolerances are configurable, via PFSolver/Simulation setters and Python bindings. Updated the three CIGRE MV notebooks that need a wider loose tolerance, and fixed a missing import in one of them.
- Found bad data for transformers in cim-grid-data's case300/case145, confirmed against the real MATPOWER source; both notebooks skipped with a FIXME.
- matpower.Reader now passes Q-limits to SynchronGenerator.
- Added case145_case300_PF_DPsim_vs_Matpower.ipynb, validating both cases via the native reader against MATPOWER's own solution.
- PF_Sparse_vs_Dense.ipynb now sweeps all four cases (case9/14/145/300) via the native matpower reader, restoring case145/case300 to the sweep instead of dropping them for the CIM data bug.

Better merged after #536.

### DIFF
--- a/dpsim-models/include/dpsim-models/SP/SP_Ph1_AvVoltageSourceInverterDQ.h
+++ b/dpsim-models/include/dpsim-models/SP/SP_Ph1_AvVoltageSourceInverterDQ.h
@@ -132,6 +132,9 @@ public:
   // #### Powerflow section ####
   /// Get nominal voltage
   Real getNomVoltage() const;
+  /// Get the nominal voltage at the network-connected terminal (accounts for
+  /// the connection transformer's ratio, if present)
+  Real getBaseVoltage() const;
 
   // #### MNA section ####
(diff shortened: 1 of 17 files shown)


georgii-tishenin (approved):
Hi @leonardocarreras. Thanks for the PR! Looks reasonable to me.

I think there still may be some edge cases, but as long as we keep working further on that, we'll figure out.

PR #544, opened 2026-07-06
DP: add Ph1 variable-SSN base for mixed real+complex state components

leonardocarreras:
## Summary
- Adds `DP::Ph1::MixedVTypeVariableSSNComp`, a per-step relinearized SSN base for components whose state mixes baseband real states with carrier-band complex states, packed into one real vector and folded to/from a complex Norton stamp at the network boundary.
- Adds a plumbing-only sanity component/example (`Variable_Serial_RLC`) verifying the new base against the existing fixed-parameter SSN RLC one-port.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -30,6 +30,8 @@
 /dpsim-models/include/dpsim-models/DP/DP_Ph1_varResSwitch.h @gnakti
 /dpsim-models/include/dpsim-models/DP/DP_Ph1_TwoTerminalVTypeSSNComp.h @leonardocarreras
 /dpsim-models/include/dpsim-models/DP/DP_Ph1_SSN_Full_Serial_RLC.h @leonardocarreras
+/dpsim-models/include/dpsim-models/DP/DP_Ph1_MixedVTypeVariableSSNComp.h @leonardocarreras
+/dpsim-models/include/dpsim-models/DP/DP_Ph1_SSN_Variable_Serial_RLC.h @leonardocarreras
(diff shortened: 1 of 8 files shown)


georgii-tishenin (approved):
Thanks, @leonardocarreras. This looks good overall and I am happy to approve it. I noticed one possible issue in the default initialization of `MixedVTypeVariableSSNComp::initializeFromNodesAndTerminals()`.

To reproduce it, set the initial node voltage in the comparison example `DP_Ph1_MixedVariableSSN_vs_FixedSSN`:

```cpp
const Complex sourceVoltage =
    CPS::Math::polar(1.0, CPS::Math::degToRad(-90.0));

auto n1 = SimNode::make("n1");
n1->setInitialVoltage(sourceVoltage);

auto vs = Ph1::VoltageSource::make("vs_" + simName);
vs->setParameters(sourceVoltage);
```

With this change, `Full_Serial_RLC` starts directly in steady state, while `Variable_Serial_RLC` shows an initial transient in current.

My guess is that the base initializer in `MixedVTypeVariableSSNComp` assumes an unshifted physical-state matrix and solves using `j*omega*I - A`, while the variable RLC already includes the `+/- omega` terms in its packed DP-envelope matrix. In that case, the carrier shift may be applied twice during initialization.

This does not affect the inverter in the following PR, since it overrides `initializeFromNodesAndTerminals()` with its own steady-state initialization. I also assume it will not affect further variable components that we will introduce, as they likely will also override the initialization. From my point of view is okay to merge the PR and deal with it somehow in the follow-up.

PR #545, opened 2026-07-06
DP: add averaged-VSI state-space inverter component

leonardocarreras:
## Summary
- Adds `DP::Ph1::AvVoltSourceInverterStateSpace`, a state-space/SSN model of an  averaged grid-following voltage-source inverter (PLL, power filter, power/current control, LCL-ish output filter) for the DP domain, built on the variable-SSN base from #544.
- Adds a per-step `updateLogAttributes` hook to `MixedVTypeVariableSSNComp` (mirrors the existing EMT SSN hierarchy's hook), needed so this component's logged attributes (vc_d/q, irc_d/q, p_inst, q_inst, omega_pll) update every step instead of only at initialization.
- Added into CMake, `Components.h`, and pybind (`dpsimpy.dp.ph1.AvVoltSourceInverterStateSpace`).

Needs #544.

### DIFF
--- a/dpsim-models/include/dpsim-models/Components.h
+++ b/dpsim-models/include/dpsim-models/Components.h
@@ -57,10 +57,12 @@
 #include <dpsim-models/DP/DP_Ph1_ProfileVoltageSource.h>
 #endif
 #include <dpsim-models/DP/DP_ITypeSSNComp.h>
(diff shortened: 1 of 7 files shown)


georgii-tishenin (dismissed):
Thanks @leonardocarreras. This looks good to me!

georgii-tishenin (approved):

PR #546, opened 2026-07-06
DP: add cxx example for averaged-VSI state-space inverter

leonardocarreras:
## Summary
Ports the EMT cxx example for the averaged-VSI state-space inverter to DP::Ph1, registered in CMakeLists.txt.

Needs #545

PR #547, opened 2026-07-06
DP: add cross-domain validation notebook for averaged-VSI state-space inverter

leonardocarreras:
## Summary
Adds a DP_Ph1 vs EMT_Ph3 cross-domain validation notebook for the averaged-VSI state-space inverter: both simulations run in the same notebook (shared power-flow init), then checked against the EMT notebook's own tolerances (p_inst/q_inst/omega_pll/vc_q, PLL-lock and post-disturbance windows), plus a direct DP-vs-EMT sample-by-sample tracking check in both windows.

Needs #546


georgii-tishenin (dismissed):
The notebook looks good!

georgii-tishenin (approved):

PR #549, opened 2026-07-06
docs: document DP::Ph1 averaged-VSI state-space inverter

leonardocarreras:
Adds a `DP::Ph1::AvVoltSourceInverterStateSpace` section to the Models/power-electronics doc, mirroring the existing EMT section's structure (state vector, model equations in DP's own native scale, terminal I/O, source/example/notebook links), plus a brief pointer from the Concepts/state-space-nodal page to it.

Needs #547


georgii-tishenin (dismissed):
Thanks for adding the documentation. Looks good.
Control part of the equations is the same for EMT and DP inverters, electrical circuit (but not the equations domain) is also same. In the future we may want to extract common part and describe specializations of EMT and DP models.

georgii-tishenin (approved):

PR #552, opened 2026-07-08
feat: configurable PFSolver base-power fallback and working max-iterations setter

leonardocarreras:
## Summary
- Adds `setBaseApparentPowerFallback()`/`getBaseApparentPowerFallback()` to `PFSolver`, wired through `Simulation` and pybind the same way as `setKeepLastSolution()`. Defaults to the existing hardcoded 100e6 VA, so behavior is unchanged unless a caller opts in.
- Adds `setMaxIterations()`/`getMaxIterations()` to `PFSolver`, operating on its own `mMaxIterations` field rather than the base `Solver::mMaxIterations` that `Simulation::setMaxNumberOfIterations()` writes to (a same-named field it shadows, which meant that call had zero effect on the PF solver's actual iteration cap, silently stuck at its hardcoded default of 20).

Idea prompted by a branch from Matthias, which only hardcoded a different fallback constant; no code from those commits is reused here.

### DIFF
--- a/dpsim/include/dpsim/PFSolver.h
+++ b/dpsim/include/dpsim/PFSolver.h
@@ -85,6 +85,8 @@ protected:
   CPS::UInt mIterations;
   /// Base power of per-unit system
   CPS::Real mBaseApparentPower;
+  /// Fallback base apparent power if no generator or transformer rating is found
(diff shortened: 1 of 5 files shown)


matthiasmees (approved):
all good

PR #556, opened 2026-07-09
Implement functions for mapDisconnector and mapBreaker

pipeacosta:
This PR implements two new mapping functions in the CIM importer to map a disconnector and a breaker.

### DIFF
--- a/dpsim-models/include/dpsim-models/CIM/Reader.h
+++ b/dpsim-models/include/dpsim-models/CIM/Reader.h
@@ -49,6 +49,8 @@ class PowerTransformer;
 class EquivalentShunt;
 class TopologicalNode;
 class ConductingEquipment;
+class Disconnector;
+class Breaker;
 }; // namespace CIMPP
 #else
 #include <CIMNamespaces.hpp>
@@ -164,6 +166,10 @@ private:
   mapExternalNetworkInjection(CIMPP::ExternalNetworkInjection *extnet);
   /// Returns a shunt
   TopologicalPowerComp::Ptr mapEquivalentShunt(CIMPP::EquivalentShunt *shunt);
+  /// Returns a switch
+  TopologicalPowerComp::Ptr mapDisconnector(CIMPP::Disconnector *disc);
+  /// Returns a switch
+  TopologicalPowerComp::Ptr mapBreaker(CIMPP::Breaker *cb);
 
   // #### Helper Functions ####
   /// Determine base voltage associated with object
--- a/dpsim-models/src/CIM/Reader.cpp
+++ b/dpsim-models/src/CIM/Reader.cpp
@@ -115,7 +115,10 @@ TopologicalPowerComp::Ptr Reader::mapComponent(BaseClass *obj) {
   if (CIMPP::EquivalentShunt *shunt =
           dynamic_cast<CIMPP::EquivalentShunt *>(obj))
     return mapEquivalentShunt(shunt);
-
+  if (CIMPP::Disconnector *disc = dynamic_cast<CIMPP::Disconnector *>(obj))
+    return mapDisconnector(disc);
+  if (CIMPP::Breaker *cb = dynamic_cast<CIMPP::Breaker *>(obj))
+    return mapBreaker(cb);
   return nullptr;
 }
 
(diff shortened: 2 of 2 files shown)


dpsim-models/src/CIM/Reader.cpp
```diff
@@ -6,10 +6,13 @@
  * file, You can obtain one at https://mozilla.org/MPL/2.0/.
  *********************************************************************************/
 
+#include "dpsim-models/Definitions.h"
+#include "dpsim-models/EMT/EMT_Ph1_PiLine.h"
```
leonardocarreras: not needed EMT_Ph1_PiLine

dpsim-models/src/CIM/Reader.cpp
```diff
@@ -1101,6 +1107,144 @@ Reader::mapEquivalentShunt(CIMPP::EquivalentShunt *shunt) {
   return cpsShunt;
 }
 
+TopologicalPowerComp::Ptr Reader::mapDisconnector(CIMPP::Disconnector *disc) {
+
+  SPDLOG_LOGGER_INFO(mSLog, "Found Disconnector {} with status {}",
+                     cimString(disc->name), (bool)disc->open.value);
+
+  Real baseVoltage = determineBaseVoltageAssociatedWithEquipment(disc);
```
leonardocarreras: Suggestion: use the Switch, otherwise we cannot change this in simulation time.
Obs: see if the values make sense for open and close. Do not use ^ in cpp (this is XOR, not exponent).

```cpp
TopologicalPowerComp::Ptr Reader::mapSwitch(const String &mRID,
                                            const String &name, Bool open,
                                            Real openResistance,
                                            Real closedResistance) {
  if (mPhase == PhaseType::ABC) {
    if (mDomain == Domain::EMT) {
      auto sw = std::make_shared<EMT::Ph3::SeriesSwitch>(mRID, name,
                                                         mComponentLogLevel);
      sw->setParameters(openResistance, closedResistance, !open);
      return sw;
    } else if (mDomain == Domain::DP) {
      auto sw = std::make_shared<DP::Ph3::SeriesSwitch>(mRID, name,
                                                         mComponentLogLevel);
      sw->setParameters(openResistance, closedResistance, !open);
      return sw;
    } else {
      SPDLOG_LOGGER_INFO(mSLog, "Mapping of switches for SP::Ph3 not existent!");
      return nullptr;
    }
  } else {
    if (mDomain == Domain::EMT) {
      auto sw = std::make_shared<EMT::Ph1::Switch>(mRID, name,
                                                    mComponentLogLevel);
      sw->setParameters(openResistance, closedResistance, !open);
      return sw;
    } else if (mDomain == Domain::DP) {
      auto sw = std::make_shared<DP::Ph1::Switch>(mRID, name,
                                                   mComponentLogLevel);
      sw->setParameters(openResistance, closedResistance, !open);
      return sw;
    } else {
      auto sw = std::make_shared<SP::Ph1::Switch>(mRID, name,
                                                   mComponentLogLevel);
      sw->setParameters(openResistance, closedResistance, !open);
      return sw;
    }
  }
}

TopologicalPowerComp::Ptr Reader::mapDisconnector(CIMPP::Disconnector *disc) {
  Bool open = disc->open.value;
  SPDLOG_LOGGER_INFO(mSLog, "Found Disconnector {} with status {}",
                     cimString(disc->name), (bool)open);
  return mapSwitch(disc->mRID, disc->name, open, /*openResistance*/ 1e9,
                   /*closedResistance*/ 1e-3);
}

TopologicalPowerComp::Ptr Reader::mapBreaker(CIMPP::Breaker *cb) {
  Bool open = cb->open.value;
  SPDLOG_LOGGER_INFO(mSLog, "Found Breaker {} with status {}",
                     cimString(cb->name), (bool)open);
  return mapSwitch(cb->mRID, cb->name, open, /*openResistance*/ 1e9,
                   /*closedResistance*/ 1e-3);
}
```

dpsim-models/include/dpsim-models/CIM/Reader.h
```diff
@@ -164,6 +166,8 @@ class Reader {
   mapExternalNetworkInjection(CIMPP::ExternalNetworkInjection *extnet);
   /// Returns a shunt
   TopologicalPowerComp::Ptr mapEquivalentShunt(CIMPP::EquivalentShunt *shunt);
+  TopologicalPowerComp::Ptr mapDisconnector(CIMPP::Disconnector *disc);
```
leonardocarreras, line 170: Add a comment on top of this ones, they return a... (they are not a shunt)

leonardocarreras (changes requested):

leonardocarreras:
Hi Andres! Thanks for the PR! I find that the use of MNA Switch component would be beneficial to have the chance to create events for the dynamic simulation cases, would it be possible to think about that? I left a suggestion, needs checking

leonardocarreras (approved):
I think this is generally okay. 

However, I would still suggest as a follow-up some cleanup on the duplications (both functions are almost identical), is non-blocking, but an issue about that would be nice (or if you have time to fix now).

PR #558, opened 2026-07-09
fix: CIM reader PQ-generator mapping and slack voltage unit bug

leonardocarreras:
## Summary
- SynchronousMachine without a RegulatingControl association now maps to PQ (fixed P, Q=0) instead of forcing PV with a 0V setpoint, which singularized the PF Jacobian and crashed the solver.
- PFSolver::determinePFBusType() now recognizes a SynchronGenerator with PowerflowBusType::PQ.
- ExternalNetworkInjection's voltage set-point now correctly applies UnitMultiplier via unitValue() instead of treating targetValue as already per-unit and multiplying by baseVoltage again, causing non-convergence.


matthiasmees (approved):
well done

PR #567, opened 2026-07-12
DP: fix double carrier-shift in MixedVTypeVariableSSNComp init

### LINKED ISSUE #566: Investigate mismatch for DP SSN components
Thanks, @leonardocarreras. This looks good overall and I am happy to approve it. I noticed one possible issue in the default initialization of `MixedVTypeVariableSSNComp::initializeFromNodesAndTerminals()`.

To reproduce it, set the initial node voltage in the comparison example `DP_Ph1_MixedVariableSSN_vs_FixedSSN`:

```cpp
const Complex sourceVoltage =
    CPS::Math::polar(1.0, CPS::Math::degToRad(-90.0));

auto n1 = SimNode::make("n1");
n1->setInitialVoltage(sourceVoltage);

auto vs = Ph1::VoltageSource::make("vs_" + simName);
vs->setParameters(sourceVoltage);
```

With this change, `Full_Serial_RLC` starts directly in steady state, while `Variable_Serial_RLC` shows an initial transient in current.

My guess is that the base initializer in `MixedVTypeVariableSSNComp` assumes an unshifted physical-state matrix and solves using `j*omega*I - A`, while the variable RLC already includes the `+/- omega` terms in its packed DP-envelope matrix. In that case, the carrier shift may be applied twice during initialization.

This does not affect the inverter in the following PR, since it overrides `initializeFromNodesAndTerminals()` with its own steady-state initialization. I also assume it will not affect further variable components that we will introduce, as they likely will also override the initialization. From my point of view is okay to merge the PR and deal with it somehow in the follow-up.

_Originally posted by @georgii-tishenin in https://github.com/sogno-platform/dpsim/pull/544#pullrequestreview-4680246534_

leonardocarreras:
## Summary
`MixedVTypeVariableSSNComp::initializeFromNodesAndTerminals()` re-applied the carrier shift (`jω·I − A`) on top of an `mA` that variable components already hand it pre-shifted, doubling it. Fixed by solving the steady state directly against `mA`. Updated the comparison example to set an explicit initial voltage so it actually catches this.

Discovered by @georgii-tishenin while reviewing #544. Fixed here.

Closes #566

### DIFF
--- a/dpsim-models/src/DP/DP_Ph1_MixedVTypeVariableSSNComp.cpp
+++ b/dpsim-models/src/DP/DP_Ph1_MixedVTypeVariableSSNComp.cpp
@@ -186,23 +186,18 @@ void DP::Ph1::MixedVTypeVariableSSNComp::initializeFromNodesAndTerminals(
                      "\n--- Initialization from nodes and terminals ---");
 
   Bool converged = false;
-  const Real omega = 2.0 * PI * frequency;
+
+  Matrix uColumn = Matrix::Zero(2, 1);
+  uColumn(0, 0) = u.real();
+  uColumn(1, 0) = u.imag();
 
   for (Int iter = 0; iter < mInitializationMaxIterations; ++iter) {
     const Matrix xPrev = **mX;
 
     updateComponentParameters();
 
-    MatrixComp h =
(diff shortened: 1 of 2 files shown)


georgii-tishenin (approved):
Thanks for addressing this so quickly

PR #570, opened 2026-07-14
Feat: EMT Ph3 SSN grid-forming inverter + automatic GFM controller derivation + CIM reader robustness

matthiasmees:
### EMT Ph3 SSN grid-forming inverter (`EMT_Ph3_SSN_GFM`)
- New three-phase grid-forming inverter as a locally-linearized affine state-space (SSN) component: VSG swing dynamics, reactive/voltage droop, cascaded voltage + current control, LC filter in abc, numerical Jacobian linearization each step
- pybind bindings, cxx example (`EMT_SSN_GFM_Test`) and equivalent notebook (`EMT_SSN_GFM_LoadStep`)
- Tested with a load-step case: self-consistent initialization, voltage forming, frequency droop and recovery all working

### Automatic GFM controller derivation
- Added derived equations for automatic controller parameter setting of the GFM
- Minor fixes to the cxx example
- Tested at 20 kV: working

### CIM reader robustness
- Null-guards in `processSvPowerFlow` and `processTopologicalNode`
- Skip topological nodes / terminals with no equipment the reader can map (gate list mirrors `mapComponent`)


dpsim-models/src/CIM/Reader.cpp
```diff
@@ -1339,9 +1417,16 @@ void Reader::processTopologicalNode(CIMPP::TopologicalNode *topNode) {
                        nodeRid, cimString(topNode->name),
                        mPowerflowNodes[nodeRid]->matrixNodeIndex());
 
+  // Safety Log
```
leonardocarreras: Safety in which sense would be good to elaborate

dpsim-models/src/CIM/Reader.cpp
```diff
@@ -1339,9 +1417,16 @@ void Reader::processTopologicalNode(CIMPP::TopologicalNode *topNode) {
...
-    // Insert Terminal if it does not exist in the map and add reference to node.
-    // This could be optimized because the Terminal is searched twice.
+    if (!term) {
+      SPDLOG_LOGGER_WARN(
+          mSLog, "TopologicalNode {} contains a null Terminal, ignoring",
+          nodeRid);
+      continue;
+    }
+
+    const auto &termRid = cimString(term->mRID);
```
leonardocarreras: declared twice, likely the reason why CI did not run correctly
```suggestion
```

dpsim-models/src/CIM/Reader.cpp
```diff
@@ -1359,34 +1444,44 @@ void Reader::processTopologicalNode(CIMPP::TopologicalNode *topNode) {
...
-          SPDLOG_LOGGER_WARN(mSLog, "Could not map equipment {}", equipmentRid);
-          continue;
-        }
-      }
+      continue;
+    }
+
+    // Unsupported Equipment check
+    if (!isSupportedConductingEquipment(equipment)) {
+      SPDLOG_LOGGER_DEBUG(
```
leonardocarreras: I would prefer this to be WARN, so the user knows we skipped

dpsim-models/src/CIM/Reader.cpp
```diff
@@ -101,6 +101,37 @@ Real Reader::unitValue(Real value, CIMPP::UnitMultiplier mult) {
...
+
+  if (dynamic_cast<CIMPP::EquivalentShunt *>(obj))
+    return true;
+
+  if (dynamic_cast<CIMPP::Disconnector *>(obj))
+    return true;
+
+  if (dynamic_cast<CIMPP::Breaker *>(obj))
+    return true;
+
```
leonardocarreras, line 131: I think we might be missing some components that are only a node, like  BusbarSection, Junction, GroundDisconnector and so... Not sure how important is, but would be deleting early...

leonardocarreras (changes requested):
Hi Matthias! Thanks for the PR and the idea.

I think the procedure will be interesting to have for all kind of circuits (and selectable, so you can also opt-out). For this to be possible it needs to be done later after the topology is complete, from the SystemTopology as a helper function would be an option. If we do it only here, then circuits imported with other methods or created manually will not benefit. What do you think? If you want to split this PR and have only the GFM fixes considered first, feel free to adapt the PR and create a new one for the topology checking.

dpsim-models/src/CIM/Reader.cpp
matthiasmees: normally you only have nodes connected to objects, i think we can keep, if there are some CIMs with this kind of stuff, we can fix

leonardocarreras (approved):
Hi @matthiasmees! Thanks for addressing the comments, we can merge this.

I agree that doing it here does not produce any extra effects. We need to keep and eye to have the lists in sync.

I renamed and edited the PR texts.

A kind request for next times, please split the changes that do not belong to the same things to make the review easier.

PR #571, opened 2026-07-14
Add state-space extraction for DP mixed variable SSN components

georgii-tishenin:
This PR adds state-space extraction support for DP Ph1 mixed variable SSN components, such as the averaged voltage-source inverter.

It includes C++ and notebook examples that validate native DP extracted eigenvalues against an analytical dq-frame reference and compare DP, EMT `GlobalDQ0`, and analytical results for stable and unstable cases.

PR #573, opened 2026-07-14
Support EMT Ph3 and DP Ph1 composite components in state-space extraction

georgii-tishenin:
Adds state-space extraction support for the following EMT Ph3 and DP Ph1 composite components by processing their immediate MNA subcomponents:

- `NetworkInjection`
- `PiLine`
- `RXLoad`
- `RxLine`
- `Shunt`
- `Transformer`

Includes C++ and Jupyter validation for EMT Ph3 and DP Ph1, comparing composite `PiLine` and `RXLoad` models against equivalent explicit RLC models.

Updates documentation of supported scope of state-space extraction, including newly supported components.

Depends on and should be merged after #571.


dpsim/examples/cxx/StateSpace/EMT_Ph3_Composite_StateSpaceExtraction.cpp
```diff
@@ -0,0 +1,315 @@
+#include <DPsim.h>
```
leonardocarreras, line 3: SPDX header is missing

docs/hugo/content/en/docs/Overview/state-space-extraction-dpsim.md
```diff
@@ -12,30 +12,74 @@ The implementation is organized around three main parts:
...
+contributions, while the composite parent remains part of the simulation and
+retains its normal MNA stamping.
+
+Other component types are rejected explicitly when state-space extraction is
+enabled.
 
 # Usage
 
-In C++, state-space extraction can be enabled on a simulation as follows:
+In C++, state-space extraction can be enabled for an EMT simulation as follows:
```
leonardocarreras: only EMT? maybe we can reword, the DP comment is a bit far so in a first glance it looks limited

dpsim/examples/cxx/StateSpace/DP_Ph1_Composite_StateSpaceExtraction.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+    // RXLoad branches are connected in parallel between ground and the load
+    // node.
+    loadResistance->connect({SimNode::GND, loadNode});
+    loadInductance->connect({SimNode::GND, loadNode});
+
+    // Keep the component order consistent with the internal subcomponent
+    // order of PiLine and RXLoad. This gives the same extraction-state order
+    // in both models.
+    auto system = SystemTopology(
```
leonardocarreras, line 179: could we eventually test at some point the other components too, besides RXLoad and PiLine?

dpsim/examples/cxx/StateSpace/DP_Ph1_Composite_StateSpaceExtraction.cpp
```diff
@@ -0,0 +1,309 @@
...
+    simulation.doStateSpaceExtraction(true);
+    simulation.run();
+
+    return collectExtractionResult(simulation);
+  }
+
+  void validateResults(const ExtractionResult &compositeResult,
+                       const ExtractionResult &explicitResult) const {
+    constexpr UInt expectedStateCount = 8;
+    constexpr Real tolerance = 1e-10;
```
leonardocarreras: I think we will need to pick a value, sometimes is 1e-9, others is 1e-10... An option would be `DOUBLE_EPSILON` that is in Definitions.h or a new one for the notebooks of this SSN kind

leonardocarreras (changes requested):
Hi @georgii-tishenin, thanks for the PR. There are some small changes needed in SPDX. The rest can be solved also in a future work.

Maybe would be important to notice that currently, composites can recursively have infinite sub-composites. This could be a comment if we do not want to support that in the extractor.

Another point is that create, getSupportedRealComposite and getSupportedComposite are very similar lists but in different parts. We might want to document that (either in docu or comments), because in this case adding a new component needs changes in more than one place.

dpsim/examples/cxx/StateSpace/EMT_Ph3_Composite_StateSpaceExtraction.cpp
georgii-tishenin: thanks, done

docs/hugo/content/en/docs/Overview/state-space-extraction-dpsim.md
georgii-tishenin: reworded

dpsim/examples/cxx/StateSpace/DP_Ph1_Composite_StateSpaceExtraction.cpp
georgii-tishenin: done

dpsim/examples/cxx/StateSpace/DP_Ph1_Composite_StateSpaceExtraction.cpp
georgii-tishenin: yes. Now is just a bit tricky to use the same test approach e.g. for transformer, because we do not have a standalone ideal transformer component. But I agree, tests should cover all supported components.

leonardocarreras (approved):
Thank you for addressing the comments

georgii-tishenin:
> Hi @georgii-tishenin, thanks for the PR. There are some small changes needed in SPDX. The rest can be solved also in a future work.
> 
> Maybe would be important to notice that currently, composites can recursively have infinite sub-composites. This could be a comment if we do not want to support that in the extractor.
> 
> Another point is that create, getSupportedRealComposite and getSupportedComposite are very similar lists but in different parts. We might want to document that (either in docu or comments), because in this case adding a new component needs changes in more than one place.

thanks for the review, @leonardocarreras. I addressed the in-code comments (except for new tests). I also agree that if you create new composite component classes for two domains it's a double job here to add them to the lists. But currently in DPsim it's a double-job everywhere😅 (classes, cmake, examples, etc.) because composite components are not domain neutral (which they maybe could be?)

PR #574, opened 2026-07-15
Support EMT Ph3 and DP Ph1 switches in state-space extraction

georgii-tishenin:
Adds state-space extraction support for `EMT::Ph3::Switch` and `DP::Ph1::Switch` as state-less components whose active resistance is represented by the MNA system matrix.

Includes C++ and Jupyter validation against equivalent resistor models for open, closed, and event-driven switching, covering both precomputed switch matrices and system-matrix recomputation.

Updates documentation.


leonardocarreras (approved):
It looks good to me.

A chance to improve in a follow-up is to verify both the open and closed state in the same simulation, in both directions close->open and open->close.

PR #575, opened 2026-07-15
Add IEEE 9-bus 4th-order SG examples (SP, DP, EMT)

leonardocarreras:
Power-flow-initialized IEEE 9-bus dynamic simulations in the SP, DP and EMT domains, with the same generator models (4th-order `SynchronGenerator4OrderVBR`, `ExciterDC1Simp` AVR, `TurbineGovernorType1`) so the domains are comparable.

- Three examples under `dpsim/examples/cxx/Circuits/`, shared parameters in `Examples.h` (`Grids::IEEE9`)
- Notebook `examples/Notebooks/Circuits/IEEE9_4Order_SP_DP_EMT.ipynb` comparing bus voltages and generator behaviour across the domains
- System frequency selectable via `-f` (default 60 Hz); per-unit data reused for a 50 Hz variant

_Note:_ work for dynamic simulations supported for EMT 3ph and DP 1ph by @denibeleri1 during his time at ACS. Adaptations by me, including addition of SP domain.

### DIFF
--- a/dpsim/examples/cxx/CMakeLists.txt
+++ b/dpsim/examples/cxx/CMakeLists.txt
@@ -71,6 +71,11 @@ set(CIRCUIT_SOURCES
 	Circuits/EMT_DP_SP_Trafo.cpp
 	Circuits/EMT_DP_SP_Slack_PiLine_PQLoad_FrequencyRamp_CosineFM.cpp
 
+	# IEEE 9-bus, 4th-order synchronous generators (PF-initialized)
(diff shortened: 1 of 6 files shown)


georgii-tishenin (dismissed):
Cool beans!

Two thoughts (not directly related to this PR):

1. Maybe we want to highlight somewhere in documentation that we have these kind of examples.
2. Maybe we can double-check: if we can remove some previous notebooks / .cpp examples, as this is a good integration test.

georgii-tishenin (approved):

PR #576, opened 2026-07-17
Add automatic system matrix recomputation mode

georgii-tishenin:
System-matrix recomputation currently defaults to disabled, so users of variable (non-switch) components must know to enable it manually. Missing this setting leads to wrong simulation results.

This PR adds `Auto`, `Enabled,` and `Disabled` modes, with `Auto` as the new default. The MNA solver resolves `Auto` from the topology, logs the effective selection, and warns when recomputation is explicitly disabled for a component that may require it. The existing Boolean setter remains available as a compatibility wrapper and is marked as deprecated in the API documentation. 

Special case handling:
Some components implement `MNASwitchInterface` but cannot be represented completely by precomputed switch matrices. To support these cases, `MNASwitchInterface` now exposes whether precomputed matrices are sufficient. `varResSwitch` reports that recomputation is required because its resistance passes through intermediate values during switching.

A notebook verifies:

- `Auto == Enabled` for a non-switch variable component;
- `Auto == Disabled` for an ordinary switch;
- an explicit `Disabled` mode produces a warning for `varResSwitch`.

Suggested next steps:

1. Update existing examples to use the new mode API or rely on the Auto default instead of calling the old Boolean setter.
2. Remove the Boolean setter in a later breaking release.


leonardocarreras (changes requested):
Hi @georgii-tishenin, thanks for the PR. Looks like a really nice improvement so the user does not need to take care of this error-prone switch.

Just one comment that is more relevant... Could you check please that the SP_Ph1_varResSwitch.h also implements the override?

leonardocarreras (approved):
Looks good, thanks for addressing the edge cases and comments, lets merge

PR #577, opened 2026-07-18
grid-connected grid-forming SSN inverter + IEEE 9-bus mixed example

leonardocarreras:
- Opt-in grid-forming control extensions to `EMT::Ph3::SSN_GFM` (3 setters, all default to a no-op so the islanded model is unchanged):
  - `setGridCurrentFeedforward(scale)`: scale/disable the voltage-loop grid-current feedforward.
  - `setVirtualImpedance(Rv, Xv)`: virtual output impedance off the filter current.
  - `setReactivePowerDroop(Dq, cutoff)`: proportional Q-V droop replacing the integral excitation.
- New example `EMT_Ph3_IEEE9_SSN_InverterMix.cpp`: IEEE 9-bus with gen1 SG, gen2 to SSN_GFM (grid-forming), gen3 to averaged VSI (grid-following).

### DIFF
--- a/dpsim-models/include/dpsim-models/EMT/EMT_Ph3_SSN_GFM.h
+++ b/dpsim-models/include/dpsim-models/EMT/EMT_Ph3_SSN_GFM.h
@@ -181,9 +181,10 @@ public:
 
   /// \brief Configure the GFM model.
   ///
-  /// nominalVoltage is the nominal line-to-line RMS voltage. With the
-  /// power-invariant Park transformation used here, a balanced system's
-  /// dq voltage magnitude equals the line-to-line RMS voltage.
+  /// nominalVoltage is the nominal peak phase voltage. With the
+  /// amplitude-invariant (2/3) Park transformation used here, a balanced
+  /// system's dq voltage magnitude equals the peak phase voltage; pass
+  /// RMS3PH_TO_PEAK1PH * (line-to-line RMS voltage), as the example does.
   ///
   /// omegaN is the nominal angular frequency in rad/s.
   void setParameters(Real lf, Real cf, Real rf, Real rc, Real nominalVoltage,
--- a/dpsim-models/include/dpsim-models/EMT/EMT_Ph3_SSN_GFM.h
+++ b/dpsim-models/include/dpsim-models/EMT/EMT_Ph3_SSN_GFM.h
@@ -99,6 +99,19 @@ private:
   Real mPowerFilterCutoff;
   Real mDelayBandwidth;
 
+  // Virtual output impedance (opt-in). Zero recovers the islanded model.
+  Real mVirtualResistance;
+  Real mVirtualReactance;
+
+  // Scale on the grid-current feedforward in the voltage loop (1 = islanded).
+  Real mGridCurrentFeedforward;
+
+  // Proportional Q-V droop (opt-in). Cutoff > 0 selects droop over the integral
+  // excitation. Setpoint is captured at initialization.
+  Real mReactivePowerDroop;
+  Real mReactiveDroopCutoff;
+  Real mVoltageSetpoint;
+
   // Numerical linearization settings
   Real mJacobianRelativeStep;
   Real mJacobianAbsoluteStep;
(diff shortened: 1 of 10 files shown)


dpsim-models/src/EMT/EMT_Ph3_SSN_GFM.cpp
```diff
@@ -686,9 +724,16 @@ void EMT::Ph3::SSN_GFM::initializeFromNodesAndTerminals(Real frequency) {
...
 
-  // At a balanced steady operating point, the d-axis reference is the
-  // capacitor-voltage magnitude.
-  x0(VoltageMagnitude, 0) = voltageMagnitudeInitial;
+  // EMF carries the virtual-impedance drop so vc stays at the target.
+  const Real virtualDropD0 =
+      mVirtualResistance * iGridD0 - mVirtualReactance * iGridQ0;
+  const Real virtualReferenceQ0 =
+      -(mVirtualResistance * iGridQ0 + mVirtualReactance * iGridD0);
+  x0(VoltageMagnitude, 0) = voltageMagnitudeInitial + virtualDropD0;
```
georgii-tishenin: Can you double check if initialization equations are consistent with runtime equations for the case of non-zero virtual impedance? Basically can be tested by using non-zero virtual impedance in the example.

LLM flagged this:
"The runtime model applies the virtual impedance to ifD/ifQ, but initialization computes the compensating drop and q-axis reference from iGridD0/iGridQ0. Filter and grid currents differ by the capacitor current. This can cause a startup transient in case of non-zero virtual impedance."

georgii-tishenin (commented):
Hey @leonardocarreras. Thanks for the PR. I really like the new feature set. I reviewed with the help of LLM and marked two possible inconsistencies.

dpsim-models/src/EMT/EMT_Ph3_SSN_GFM.cpp
georgii-tishenin, line 55: Also here LLM flagged. Can you double check?
 "With virtual impedance enabled, v_ref_d and v_ref_q no longer correspond to the references used by the voltage controller. Could we calculate them here with the same ifD/ifQ expressions as in evaluateStateDerivative()?"

leonardocarreras: Fixed, you were right. Added a test.

dpsim-models/src/EMT/EMT_Ph3_SSN_GFM.cpp
leonardocarreras: Fixed.

matthiasmees (commented):
Looks good, please add:

- Examples.h equation-based struct on how the Controller parameters are calculated from
- Block diagram of new control structure for DPsim website
- Equations of Controllers

dpsim-models/src/EMT/EMT_Ph3_SSN_GFM.cpp
georgii-tishenin: Hey @leonardocarreras. Thanks for double checking and addressing. I asked the same LLM to re-check. Now it's happy about logging fix. But this one seems a partial solution. Here's what it says:

"Using the filter current here now matches the runtime implementation. However, the initialization is still not fully at equilibrium for nonzero virtual impedance.`theta0` is aligned with `vc`, so initially `vcQ0 = 0`, while the runtime q-axis voltage reference
`
-(Rv * ifQ0 + Xv * ifD0)`

is generally nonzero. This means that the initial derivative of `VoltageIntegratorQ` is still nonzero, even though its initial value is adjusted.

A possible fix would be to first form the internal EMF phasor as
`
E0 = Vc0 + Complex(Rv, Xv) * If0;`

and then initialize `theta0 `and the voltage magnitude from `arg(E0)` and `abs(E0)` before transforming the remaining quantities to dq."

To be completely sure can you double-check this?

dpsim-models/include/dpsim-models/EMT/EMT_Ph3_SSN_GFM.h
```diff
@@ -184,6 +197,18 @@ class SSN_GFM final : public TwoTerminalVTypeVariableSSNComp,
   void setNumericalLinearizationParameters(Real relativeStep,
                                            Real absoluteStep);
 
+  /// \brief Opt-in virtual output impedance Zv = Rv + jXv, subtracted from the
+  /// EMF reference across the grid current. Zero (default) is the islanded
```
georgii-tishenin: really minor thing. Should it be across filter current instead of grid current?

georgii-tishenin (dismissed):
I added a minor comment. From my point of view looks good. Let's still wait for what @matthiasmees says.

dpsim-models/include/dpsim-models/EMT/EMT_Ph3_SSN_GFM.h
leonardocarreras: Fixed

georgii-tishenin (dismissed):
Looks good to me. Thanks for addressing the minor comments.

matthiasmees (dismissed):

georgii-tishenin (approved):
From my point of view looks already fine. @leonardocarreras

leonardocarreras:
Rebasing. Most of the comments are not relevant (first iteration of the LLM reviewer).

PR #578, opened 2026-07-20
Add DP::Ph3 mixed variable-SSN base

leonardocarreras:
Adds `DP::Ph3::MixedVTypeVariableSSNComp`, a three-phase, two-terminal variable V-type SSN base carrying a mixed state: real control states plus per-phase complex envelope states, packed into one real vector. It is the Ph3 analogue of the merged `DP::Ph1::MixedVTypeVariableSSNComp` and the base that an averaged grid-forming inverter component will build on.

Abstract base only, no concrete component or example in this PR.


georgii-tishenin (approved):
Thanks @leonardocarreras. Looks consistent with single-phase DP. Let's continue!

PR #579, opened 2026-07-20
Add DP::Ph3::AvVoltSourceInverterStateSpace (averaged VSI SSN)

leonardocarreras:
Three-phase, per-phase-envelope port of `DP::Ph1::AvVoltSourceInverterStateSpace` onto the `DP::Ph3::MixedVTypeVariableSSNComp` base: 8 real control states (single PLL, single positive-sequence dq frame) plus 6 complex per-phase envelope states (`Vc_a/b/c`, `If_a/b/c`), relinearized every step.


georgii-tishenin (dismissed):
Thanks @leonardocarreras. Looks good to me!

georgii-tishenin (approved):

PR #580, opened 2026-07-20
add power-flow static VAR compensator in SP domain

leonardocarreras:
Voltage-controlling shunt reactive device modelled as a PQ bus, to be driven by the SVC outer control loop. No active power, no admittance stamp; participates via bus classification and reactive setpoint. Reuses the SynchronGenerator per-unit/attribute pattern (V_set_pu, Q_max_pu, Q_min_pu).

For real usage, #582 is the follow up.


leonardocarreras:
Superseded by #656. Moved into a stack with the solver loop so each pull request shows only its own diff.

PR #582, opened 2026-07-20
PFSolver: SVC voltage-control outer loop

leonardocarreras:
Adds the solver-side outer loop that holds each SVC bus at its voltage setpoint. Opt in via `set_pf_solver_enforce_svc_control`.

Needs #580


leonardocarreras:
Superseded by #657. Moved into a stack on top of #656, so this one now shows only its own 8 files instead of 16.

PR #583, opened 2026-07-20
 add DP::Ph3::Switch three-phase matrix switch

leonardocarreras:
Adds `DP::Ph3::Switch`, a 3x3 open/closed conductance switch for the DP domain


georgii-tishenin (approved):
Hi @leonardocarreras. Looks consistent to EMT Ph3 Switch. Let's go!

PR #585, opened 2026-07-20
Fix InterfaceVillasQueueless signal indexing, read handling, and start-of-sim sync

leonardocarreras:
- Add the missing index increment in the export signal loop so multi-signal exports map to the right sample slots.
- Read path: replace the exit-on-error behaviour with a single read that logs and returns the last sequence; the retry-on-zero loop is now gated on `blockOnRead`; guard a null allocation.
- Take the sequence number from the sample's `HAS_SEQUENCE` field instead of assuming the first attribute carries it.
- syncImports blocks only when an import is marked to sync at start, and that read always waits regardless of `blockOnRead`.
- Honour `blockOnRead`, which this interface accepted and then ignored: it always blocked. The default (`false`) is best-effort with reuse. It is per-import in the API but necessarily per-interface here, since one read fills the whole sample.
- Document the read and blocking behaviour, and that `InterfaceVillasQueueless` transfers on the simulation thread rather than through a queue.

### DIFF
--- a/dpsim-villas/src/InterfaceVillasQueueless.cpp
+++ b/dpsim-villas/src/InterfaceVillasQueueless.cpp
@@ -115,6 +115,7 @@ void InterfaceVillasQueueless::createSignals() {
     }
     nodeOutputSignals->push_back(std::make_shared<node::Signal>(
         "", "", stdTypeToNodeType(attr->getType())));
+    idx++;
   }
 
   node::SignalList::Ptr nodeInputSignals = mNode->getInputSignals(true);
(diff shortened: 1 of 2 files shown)


dpsim-villas/src/InterfaceVillasQueueless.cpp
```diff
@@ -218,20 +219,19 @@ Int InterfaceVillasQueueless::readFromVillas() {
...
-                            "Fatal error: failed to read sample from "
-                            "InterfaceVillas. Read returned code {}",
-                            ret);
-        close();
-        std::exit(1);
-      } else if (ret == 0) {
-        SPDLOG_LOGGER_WARN(mLog,
-                           "InterfaceVillas read returned 0. Retrying...");
-      }
+    if (sample == nullptr) {
```
pipeacosta, line 230: If I interpret this correctly, this change implies that the InterfaceVillas tries only once, what's the reason for this?

leonardocarreras: Hi @pipeacosta! Thanks for the question, is true that is a change on the way we do things. My way of thinking is that queued interface will work in a dedicated thread, so it can retry. And for the queueless, we read in the PreStep, so the simulation will block until we get data in the existing implementation. 
I was doing some experiments with the RT mode on DPsim. We gain robustness in some edge cases, reusing the last sample if we have one miss... For some RT experiments, it blocks too long for one single de-sync, or it exits with error... So basically one best effort in reading, if nothing is there, we continue with an old sample instead of exiting with error... It might clash with some behaviour that depends on "new sample arrives, then we run one step", so sometimes will reuse an old sample for the external pacing source...
An intermediate solution would be making this behaviour as selectable option, would that be better in your opinion? We can use the blockOnRead argument for that (this is not implemented in that way currently) so is possible to have either of them

pipeacosta: Thanks for the explanation. I would stay with the philosophy of trying to keep real-time execution with the queueless interface, and use previous samples if needed. It would be good to log a warning in this case. Having said that, it is not necessary to implement the intermediate solution you mentioned imo.

PR #586, opened 2026-07-20
Add DP::Ph3 averaged VSI unbalanced (SLG) validation

leonardocarreras:
DP::Ph3 averaged voltage source inverter validated against EMT under a moderate single-line-to-ground fault, the case a single-phase dynamic phasor model cannot represent. Adds a DP::Ph3::NetworkInjection slack primitive, a cxx example, and a cross-domain notebook (EMT + DP off one shared power flow, balanced + phase-A fault, three-tier tolerances). Also adds the missing DP::Ph3::PiLine Python binding.

Needs #579, #583


georgii-tishenin (dismissed):
Looks generally good. I know you've mentioned some minor difference between EMT and DP in the unbalanced case. I'd suggest merging and then re-checking with the help of eigenvalue analysis.

leonardocarreras:
@georgii-tishenin yes, there is a small mismatch due to the representation of the negative sequence and the controller, we can see how to deal with that in a follow-up with more information

georgii-tishenin (approved):

PR #587, opened 2026-07-21
Generic Ph1 two-terminal linear SSN components for SP domain

MarvinTollnitschRWTH:
Adds generalized components for the Ph1 SP domain which implement the State Space Nodal approach for any linear two-terminal component which can be described by state space matrices and does not include internal injections.

"SSNTypeV2T" implements the V-type group variant while "SSNTypeI2T" implements the I-type group variant.

As of now, powerflow is not supported by these components.

### DIFF
--- a/dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeI2T.h
+++ b/dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeI2T.h
@@ -1,0 +1,78 @@
+// SPDX-FileCopyrightText: 2026 Institute for Automation of Complex Power Systems, EONERC, RWTH Aachen University
+// SPDX-License-Identifier: MPL-2.0
+
+#pragma once
+
+#include <dpsim-models/MNASimPowerComp.h>
+#include <dpsim-models/Solver/MNATearInterface.h>
+#include <dpsim-models/Solver/PFSolverInterfaceBranch.h>
+
+namespace CPS {
+namespace SP {
+namespace Ph1 {
+/// \brief SSNTypeI2T
+/// Model for a one phase, two terminal I-type SSN component which can be represented using
+/// a state space equation system
+/// x' = A * x + B * u
+/// y = C * x + D * u
+/// with x: state vector, y: output vector, u: input vector,
+/// where u represents external current (mIntfCurrent),
+/// y represents external voltage (mIntfVoltage),
+/// x represents any component states.
+class SSNTypeI2T : public MNASimPowerComp<Complex>,
+                   public SharedFactory<SSNTypeI2T> {
+private:
+protected:
+  MatrixComp mA;
+  MatrixComp mB;
+  MatrixComp mC;
+  MatrixComp mD;
+
+  MatrixComp mX;
+  MatrixComp mU;
+  MatrixComp mUPrev;
+
+  Complex mAdmittance;
+
+public:
+  /// Defines UID, name, component parameters and logging level
+  SSNTypeI2T(String uid, String name,
+             Logger::Level logLevel = Logger::Level::off);
+  /// Defines name and logging level
+  SSNTypeI2T(String name, Logger::Level logLevel = Logger::Level::off)
+      : SSNTypeI2T(name, name, logLevel) {}
+
+  SimPowerComp<Complex>::Ptr clone(String name) override;
+
+  // #### General ####
+  void calculateAdmittance(Real omega);
+  void setParameters(const MatrixComp &A, const MatrixComp &B,
+                     const MatrixComp &C, const MatrixComp &D);
+
+  // #### Powerflow section ####
+  /// Initializes component from power flow data
+  void initializeFromNodesAndTerminals(Real frequency) override;
+
+  // #### MNA section ####
+  /// Initializes internal variables of the component
+  void mnaCompInitialize(Real omega, Real timeStep,
+                         Attribute<Matrix>::Ptr leftVector) override;
+  /// Stamps system matrix
+  void mnaCompApplySystemMatrixStamp(SparseMatrixRow &systemMatrix) override;
+  /// Update interface voltage from MNA system result
+  void mnaCompUpdateVoltage(const Matrix &leftVector) override;
+  /// Update interface current from MNA system result
+  void mnaCompUpdateCurrent(const Matrix &leftVector) override;
+
+  void mnaCompPostStep(Real time, Int timeStepCount,
+                       Attribute<Matrix>::Ptr &leftVector) override;
+  /// Add MNA post step dependencies
+  void
+  mnaCompAddPostStepDependencies(AttributeBase::List &prevStepDependencies,
+                                 AttributeBase::List &attributeDependencies,
+                                 AttributeBase::List &modifiedAttributes,
+                                 Attribute<Matrix>::Ptr &leftVector) override;
+};
+} // namespace Ph1
+} // namespace SP
+} // namespace CPS
--- a/dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeV2T.h
+++ b/dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeV2T.h
@@ -1,0 +1,75 @@
+// SPDX-FileCopyrightText: 2026 Institute for Automation of Complex Power Systems, EONERC, RWTH Aachen University
+// SPDX-License-Identifier: MPL-2.0
+
+#pragma once
+
+#include <dpsim-models/MNASimPowerComp.h>
+#include <dpsim-models/Solver/MNATearInterface.h>
+
+namespace CPS {
+namespace SP {
+namespace Ph1 {
+/// \brief SSNTypeV2T
+/// Model for a one phase, two terminal V-type SSN component which can be represented using
+/// a state space equation system
+/// x' = A * x + B * u
+/// y = C * x + D * u
+/// with x: state vector, y: output vector, u: input vector,
+/// where u represents external voltage (mIntfVoltage),
+/// y represents external current (mIntfCurrent),
+/// x represents any component states.
+class SSNTypeV2T : public MNASimPowerComp<Complex>,
+                   public SharedFactory<SSNTypeV2T> {
+private:
+protected:
+  MatrixComp mA;
+  MatrixComp mB;
+  MatrixComp mC;
+  MatrixComp mD;
+
+  MatrixComp mX;
+  MatrixComp mU;
+  MatrixComp mUPrev;
+
+  Complex mAdmittance;
+
+public:
+  /// Defines UID, name, component parameters and logging level
+  SSNTypeV2T(String uid, String name,
+             Logger::Level logLevel = Logger::Level::off);
+  /// Defines name and logging level
+  SSNTypeV2T(String name, Logger::Level logLevel = Logger::Level::off)
+      : SSNTypeV2T(name, name, logLevel) {}
+
+  SimPowerComp<Complex>::Ptr clone(String name) override;
+
+  // #### General ####
+  void calculateAdmittance(Real omega);
+  void setParameters(const MatrixComp &A, const MatrixComp &B,
+                     const MatrixComp &C, const MatrixComp &D);
+  /// Initializes component from power flow data
+  void initializeFromNodesAndTerminals(Real frequency) override;
+
+  // #### MNA section ####
+  /// Initializes internal variables of the component
+  void mnaCompInitialize(Real omega, Real timeStep,
+                         Attribute<Matrix>::Ptr leftVector) override;
+  /// Stamps system matrix
+  void mnaCompApplySystemMatrixStamp(SparseMatrixRow &systemMatrix) override;
+  /// Update interface voltage from MNA system result
+  void mnaCompUpdateVoltage(const Matrix &leftVector) override;
+  /// Update interface current from MNA system result
+  void mnaCompUpdateCurrent(const Matrix &leftVector) override;
+
+  void mnaCompPostStep(Real time, Int timeStepCount,
+                       Attribute<Matrix>::Ptr &leftVector) override;
+  /// Add MNA post step dependencies
+  void
+  mnaCompAddPostStepDependencies(AttributeBase::List &prevStepDependencies,
+                                 AttributeBase::List &attributeDependencies,
+                                 AttributeBase::List &modifiedAttributes,
+                                 Attribute<Matrix>::Ptr &leftVector) override;
+};
+} // namespace Ph1
+} // namespace SP
+} // namespace CPS
--- a/dpsim-models/src/SP/SP_Ph1_SSNTypeI2T.cpp
+++ b/dpsim-models/src/SP/SP_Ph1_SSNTypeI2T.cpp
@@ -1,0 +1,182 @@
+// SPDX-FileCopyrightText: 2026 Institute for Automation of Complex Power Systems, EONERC, RWTH Aachen University
+// SPDX-License-Identifier: MPL-2.0
+
+#include <dpsim-models/SP/SP_Ph1_SSNTypeI2T.h>
+
+using namespace CPS;
+
+SP::Ph1::SSNTypeI2T::SSNTypeI2T(String uid, String name, Logger::Level logLevel)
+    : MNASimPowerComp<Complex>(uid, name, false, true, logLevel) {
+  setTerminalNumber(2);
+  **mIntfVoltage = Matrix::Zero(1, 1);
+  **mIntfCurrent = Matrix::Zero(1, 1);
+  mParametersSet = false;
+}
+
+void SP::Ph1::SSNTypeI2T::setParameters(const MatrixComp &A,
+                                        const MatrixComp &B,
+                                        const MatrixComp &C,
+                                        const MatrixComp &D) {
+
+  if (A.cols() != A.rows())
+    throw std::invalid_argument("A needs to be square.");
+
+  if (B.rows() != A.cols())
+    throw std::invalid_argument(
+        "Rows of B do not match columns and rows of A.");
+
+  if (B.cols() != 1)
+    throw std::invalid_argument(
+        "Two terminal I-type SSN component only has one "
+        "external current, so column number of B must be one.");
+
+  if (C.rows() != 1)
+    throw std::invalid_argument(
+        "A two terminal I type SSN component only has"
+        "one output voltage. C must have exactly one row.");
+
+  if (C.cols() != A.rows())
+    throw std::invalid_argument("Columns of C do not match rows of A.");
+
+  if (D.rows() != 1)
+    throw std::invalid_argument(
+        "A two terminal I type SSN component only has"
+        "one output voltage. D must have exactly one row.");
+
+  if (D.cols() != B.cols())
+    throw std::invalid_argument(
+        "Columns of D do not match columns of B and rows"
+        "of input vector u, which need to be one.");
+
+  mA = A;
+  mB = B;
+  mC = C;
+  mD = D;
+
+  mX = Matrix::Zero(mA.rows(), 1);
+  mU = Matrix::Zero(mB.cols(), 1);
+  mUPrev = Matrix::Zero(mB.cols(), 1);
+
+  mParametersSet = true;
+}
+
+SimPowerComp<Complex>::Ptr SP::Ph1::SSNTypeI2T::clone(String name) {
+  auto copy = SSNTypeI2T::make(name, mLogLevel);
+  copy->setParameters(mA, mB, mC, mD);
+  return copy;
+}
+
+void SP::Ph1::SSNTypeI2T::calculateAdmittance(Real omega) {
+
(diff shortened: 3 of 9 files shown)


dpsim-models/src/SP/SP_Ph1_SSNTypeI2T.cpp
```diff
@@ -0,0 +1,293 @@
...
+                     Logger::phasorToString((**mIntfCurrent)(0, 0)),
+                     Logger::phasorToString(initialSingleVoltage(0)),
+                     Logger::phasorToString(initialSingleVoltage(1)));
+}
+
+void SP::Ph1::SSNTypeI2T::mnaCompInitialize(Real omega, Real timeStep,
+                                            Attribute<Matrix>::Ptr leftVector) {
+  updateMatrixNodeIndices();
+
+  MatrixComp H_inv =
```
MarvinTollnitschRWTH: This is duplicate to "initializeFormNodesAndTerminals". Other components do not implement this here, but I could not see this being executed if "doInitFormNodesAndTerminals" is false. I don't know if it's just not needed then? Same goes for the V-type component.

matthiasmees (dismissed):
looks good, please correct override error and declare override

dpsim-models/src/SP/SP_Ph1_SSNTypeI2T.cpp
```diff
@@ -0,0 +1,293 @@
...
+                     "\nTerminal 0 voltage: {:s}"
+                     "\nTerminal 1 voltage: {:s}"
+                     "\n--- Initialization from powerflow finished ---",
+                     Logger::phasorToString((**mIntfVoltage)(0, 0)),
+                     Logger::phasorToString((**mIntfCurrent)(0, 0)),
+                     Logger::phasorToString(initialSingleVoltage(0)),
+                     Logger::phasorToString(initialSingleVoltage(1)));
+}
+
+void SP::Ph1::SSNTypeI2T::mnaCompInitialize(Real omega, Real timeStep,
```
leonardocarreras, line 121: omega is not used in the function

dpsim-models/src/SP/SP_Ph1_SSNTypeI2T.cpp
```diff
@@ -0,0 +1,293 @@
...
+
+void CPS::SP::Ph1::SSNTypeI2T::setSSNMatricesToZero() {
+  **mA = Matrix::Zero(1, 1);
+  **mB = Matrix::Zero(1, 1);
+  **mC = Matrix::Zero(1, 1);
+  **mD = Matrix::Zero(1, 1);
+
+  mX = Matrix::Zero(1, 1);
+  mU = Matrix::Zero(1, 1);
+  mUOld = Matrix::Zero(1, 1);
```
leonardocarreras: mUPrev? we are in general using Prev for the y[k-1] term

dpsim-models/src/SP/SP_Ph1_SSNTypeI2T.cpp
```diff
@@ -0,0 +1,293 @@
...
+                     Logger::complexToString(mImpedancePerUnit),
+                     Logger::complexToString(mAdmittancePerUnit));
+}
+
+void SP::Ph1::SSNTypeI2T::pfApplyAdmittanceMatrixStamp(SparseMatrixCompRow &Y) {
+  int bus1 = this->matrixNodeIndex(0);
+
+  if (std::isinf(mAdmittancePerUnit.real()) ||
+      std::isinf(mAdmittancePerUnit.imag())) {
+    std::cout << "Y:" << mAdmittancePerUnit << std::endl;
```
leonardocarreras: please use a logger or throw as part of the exception message

dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeI2T.h
```diff
@@ -0,0 +1,116 @@
+// SPDX-FileCopyrightText: 2025 Institute for Automation of Complex Power Systems, EONERC, RWTH Aachen University
```
leonardocarreras: 2026?

dpsim/examples/cxx/CMakeLists.txt
leonardocarreras, line 167: Circuits/SP_Ph1_General2TerminalSSN.cpp should be added here for the notebook to work in the CI

examples/Notebooks/Circuits/SP_Ph1_generalizedSSN.ipynb
```diff
@@ -0,0 +1,213 @@
...
+    "## Assert"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "epsilon = 1e-12\n",
```
leonardocarreras, line 145: Could we use the DOUBLE_EPSILON from Definitions.h instead?

dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeV2T.h
```diff
@@ -0,0 +1,85 @@
...
+  MatrixComp mU;
+  MatrixComp mUOld;
+  MatrixComp mW;
+  MatrixComp mYHist;
+  Real mOmega;
+  /// susceptance [S]
+  Complex mSusceptance;
+
+public:
+  const CPS::Attribute<MatrixComp>::Ptr mA;
```
leonardocarreras: do mA, mB, mC, mD need to be attributes?

dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeI2T.h
```diff
@@ -0,0 +1,116 @@
...
+
+  /// Impedance [Ohm]
+  Complex mImpedance;
+  /// Admittance [S]
+  Complex mAdmittance;
+  /// Susceptance [S]
+  Complex mSusceptance;
+
+public:
+  const CPS::Attribute<MatrixComp>::Ptr mA;
```
leonardocarreras: do mA, mB, mC, mD need to be attributes?

dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeI2T.h
```diff
@@ -0,0 +1,116 @@
...
+                   public MNATearInterface,
+                   public PFSolverInterfaceBranch {
+private:
+  void ssnUpdateState();
+  void setSSNMatricesToZero();
+
+protected:
+  MatrixComp mX;
+  MatrixComp mU;
+  MatrixComp mUOld;
```
leonardocarreras: could we use mUPrev mU_prev or similar, like in the other files?

dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeV2T.h
```diff
@@ -0,0 +1,85 @@
...
+                   public SharedFactory<SSNTypeV2T>,
+                   public MNATearInterface {
+private:
+  void ssnUpdateState();
+  void setSSNMatricesToZero();
+
+protected:
+  MatrixComp mX;
+  MatrixComp mU;
+  MatrixComp mUOld;
```
leonardocarreras: could we use mUPrev mU_prev or similar, like in the other files?

dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeV2T.h
```diff
@@ -0,0 +1,85 @@
...
+/// y = C * x + D * u
+/// with x: state vector, y: output vector, u: input vector,
+/// where u represents external voltage (mIntfVoltage),
+/// y represents external current (mIntfCurrent),
+/// x represents any component states.
+class SSNTypeV2T : public MNASimPowerComp<Complex>,
+                   public SharedFactory<SSNTypeV2T>,
+                   public MNATearInterface {
+private:
+  void ssnUpdateState();
```
leonardocarreras: do we need this?

dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeI2T.h
```diff
@@ -0,0 +1,116 @@
...
+/// with x: state vector, y: output vector, u: input vector,
+/// where u represents external current (mIntfCurrent),
+/// y represents external voltage (mIntfVoltage),
+/// x represents any component states.
+class SSNTypeI2T : public MNASimPowerComp<Complex>,
+                   public SharedFactory<SSNTypeI2T>,
+                   public MNATearInterface,
+                   public PFSolverInterfaceBranch {
+private:
+  void ssnUpdateState();
```
leonardocarreras: do we need this? admittance changes?

dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeI2T.h
```diff
@@ -0,0 +1,116 @@
...
+
+  void mnaCompPostStep(Real time, Int timeStepCount,
+                       Attribute<Matrix>::Ptr &leftVector) override;
+  /// Add MNA post step dependencies
+  void
+  mnaCompAddPostStepDependencies(AttributeBase::List &prevStepDependencies,
+                                 AttributeBase::List &attributeDependencies,
+                                 AttributeBase::List &modifiedAttributes,
+                                 Attribute<Matrix>::Ptr &leftVector) override;
+  void mnaTearApplyMatrixStamp(SparseMatrixRow &tearMatrix);
```
leonardocarreras: override missing

dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeV2T.h
```diff
@@ -0,0 +1,85 @@
...
+
+  void mnaCompPostStep(Real time, Int timeStepCount,
+                       Attribute<Matrix>::Ptr &leftVector) override;
+  /// Add MNA post step dependencies
+  void
+  mnaCompAddPostStepDependencies(AttributeBase::List &prevStepDependencies,
+                                 AttributeBase::List &attributeDependencies,
+                                 AttributeBase::List &modifiedAttributes,
+                                 Attribute<Matrix>::Ptr &leftVector) override;
+  void mnaTearApplyMatrixStamp(SparseMatrixRow &tearMatrix);
```
leonardocarreras: override missing

dpsim-models/src/SP/SP_Ph1_SSNTypeI2T.cpp
```diff
@@ -0,0 +1,293 @@
...
+  mAdmittancePerUnit = 1. / mImpedancePerUnit;
+  SPDLOG_LOGGER_INFO(mSLog, "Impedance={} [pu]  Admittance={} [pu]",
+                     Logger::complexToString(mImpedancePerUnit),
+                     Logger::complexToString(mAdmittancePerUnit));
+}
+
+void SP::Ph1::SSNTypeI2T::pfApplyAdmittanceMatrixStamp(SparseMatrixCompRow &Y) {
+  int bus1 = this->matrixNodeIndex(0);
+
+  if (std::isinf(mAdmittancePerUnit.real()) ||
```
leonardocarreras: use CPS::Math::isFinite(Complex number)

leonardocarreras (changes requested):
Hi @MarvinTollnitschRWTH! Thanks for this nice contributions. I left some comments, the short summary is:
- CI fails because TEST_SOURCES did not get to compile for the CI the circuit for running the notebook. It may even be great to do it all with dpsimpy, so we do not need the C source compiled in the CI, but this depends on you.
- PF uses std::inf, instead of Math::isFinite
- Log or throw, do not use std::cout
- I think you may want to check the uses of omega/mOmega, it might be the case is not initialized in some uses
- A/B/C/D are not logged or changed in runtime, do we need them as Attributes?
- I-Type setParameter throw strings say V-type SSN (just typo)

dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeI2T.h
```diff
@@ -0,0 +1,116 @@
...
+/// x' = A * x + B * u
+/// y = C * x + D * u
+/// with x: state vector, y: output vector, u: input vector,
+/// where u represents external current (mIntfCurrent),
+/// y represents external voltage (mIntfVoltage),
+/// x represents any component states.
+class SSNTypeI2T : public MNASimPowerComp<Complex>,
+                   public SharedFactory<SSNTypeI2T>,
+                   public MNATearInterface,
+                   public PFSolverInterfaceBranch {
```
georgii-tishenin: This is related to comments about missing `override`. An alternative is to not implement `MNATearInterface `and remove function `mnaTearApplyMatrixStamp`. If needed can be implemented in the future.

I would also propose the same for `PFSolverInterfaceBranch`: not to inherit from it in the scope of this PR, as we're not implementing PF suport yet.

dpsim-models/src/SP/SP_Ph1_SSNTypeI2T.cpp
```diff
@@ -0,0 +1,293 @@
...
+      mD(mAttributes->create<MatrixComp>("D")) {
+  setTerminalNumber(2);
+  **mIntfVoltage = Matrix::Zero(1, 1);
+  **mIntfCurrent = Matrix::Zero(1, 1);
+  mParametersSet = false;
+}
+
+void SP::Ph1::SSNTypeI2T::setParameters(const MatrixComp A, const MatrixComp B,
+                                        const MatrixComp C,
+                                        const MatrixComp D) {
```
georgii-tishenin: For consistency, I'd suggest doing parameter checks and setting similar to `setParameters` function here: https://github.com/sogno-platform/dpsim/blob/master/dpsim-models/src/DP/DP_SSNComp.cpp

(we also do it in similar way in EMT)

dpsim-models/src/SP/SP_Ph1_SSNTypeV2T.cpp
```diff
@@ -0,0 +1,238 @@
...
+      mB(mAttributes->create<MatrixComp>("B")),
+      mC(mAttributes->create<MatrixComp>("C")),
+      mD(mAttributes->create<MatrixComp>("D")) {
+  setTerminalNumber(2);
+  **mIntfVoltage = Matrix::Zero(1, 1);
+  **mIntfCurrent = Matrix::Zero(1, 1);
+  mParametersSet = false;
+}
+
+void SP::Ph1::SSNTypeV2T::setParameters(const MatrixComp A, const MatrixComp B,
```
georgii-tishenin: For consistency, I'd suggest doing parameter checks and setting similar to `setParameters` function here: https://github.com/sogno-platform/dpsim/blob/master/dpsim-models/src/DP/DP_SSNComp.cpp

(we also do it in similar way in EMT)

dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeI2T.h
```diff
@@ -0,0 +1,116 @@
...
+  /// Defines UID, name, component parameters and logging level
+  SSNTypeI2T(String uid, String name,
+             Logger::Level logLevel = Logger::Level::off);
+  /// Defines name and logging level
+  SSNTypeI2T(String name, Logger::Level logLevel = Logger::Level::off)
+      : SSNTypeI2T(name, name, logLevel) {}
+
+  SimPowerComp<Complex>::Ptr clone(String name) override;
+
+  void manualInit(Matrix initialState, Matrix initialInput,
```
georgii-tishenin: I would suggest removing `manualInit()`. It's not tested in the example, also not eventually the way we would like to initialize the component.

dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeV2T.h
```diff
@@ -0,0 +1,85 @@
...
+  SSNTypeV2T(String uid, String name,
+             Logger::Level logLevel = Logger::Level::off);
+  /// Defines name and logging level
+  SSNTypeV2T(String name, Logger::Level logLevel = Logger::Level::off)
+      : SSNTypeV2T(name, name, logLevel) {}
+
+  SimPowerComp<Complex>::Ptr clone(String name) override;
+
+  void manualInit(Matrix initialState, Matrix initialInput,
+                  Matrix initialOldInput, Real initCurrent, Real initVoltage);
```
georgii-tishenin: I would suggest removing `manualInit()`. It's not tested in the example, also not eventually the way we would like to initialize the component.

dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeV2T.h
```diff
@@ -0,0 +1,85 @@
...
+/// with x: state vector, y: output vector, u: input vector,
+/// where u represents external voltage (mIntfVoltage),
+/// y represents external current (mIntfCurrent),
+/// x represents any component states.
+class SSNTypeV2T : public MNASimPowerComp<Complex>,
+                   public SharedFactory<SSNTypeV2T>,
+                   public MNATearInterface {
+private:
+  void ssnUpdateState();
+  void setSSNMatricesToZero();
```
georgii-tishenin: I'd suggest removing `setSSNMatricesToZero() `completely. It's now used only in `setParameters()` for setting matrices to zero if the parameter matrices have unexpected shape. I think throwing an unhandled exception with clear message is better, because it then terminates the program that would otherwise run incorrectly and log error.

dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeI2T.h
```diff
@@ -0,0 +1,116 @@
...
+/// where u represents external current (mIntfCurrent),
+/// y represents external voltage (mIntfVoltage),
+/// x represents any component states.
+class SSNTypeI2T : public MNASimPowerComp<Complex>,
+                   public SharedFactory<SSNTypeI2T>,
+                   public MNATearInterface,
+                   public PFSolverInterfaceBranch {
+private:
+  void ssnUpdateState();
+  void setSSNMatricesToZero();
```
georgii-tishenin: I'd suggest removing `setSSNMatricesToZero() `completely. It's now used only in `setParameters()` for setting matrices to zero if the parameter matrices have unexpected shape. I think throwing an unhandled exception with clear message is better, because it then terminates the program that would otherwise run incorrectly and log error.

georgii-tishenin (changes requested):

dpsim-models/src/SP/SP_Ph1_SSNTypeI2T.cpp
georgii-tishenin: Yes, calculating the frequency-domain equivalent in both methods is required: initializeFromNodesAndTerminals() initializes the operating point, while mnaCompInitialize() prepares the MNA stamp and is still called when node-based initialization is disabled. The duplication could be moved into a helper, but in mnaCompInitialize() it should use the passed omega rather than mOmega, because mOmega is only set by the optional node initialization.

dpsim-models/include/dpsim-models/SP/SP_Ph1_SSNTypeI2T.h
MarvinTollnitschRWTH: They don't need to. I create them as attributes since I thought that for any trait special to a component which one may want to check or retrieve it would make sense to be implemented as an attribute. But I can change them back into normal members and just add getters like you did in DP_SSNComp, for example, with "getDiscreteA()" or "getC". Your call!

leonardocarreras: Hi, sorry I missed this... I think is better if they are all the same and consistent, so would prefer the normal members unless there is a strong opinion that they should be kept

examples/Notebooks/Circuits/SP_Ph1_generalizedSSN.ipynb
MarvinTollnitschRWTH: Yes, but according to Georgii we'd need to pybind it and he suggests doing it in another PR.

leonardocarreras:
@MarvinTollnitschRWTH thanks for the changes, whenever is convenient for you please rebase

georgii-tishenin (approved):
@MarvinTollnitschRWTH thanks for addressing the comments. Looks good to me.

leonardocarreras (approved):
Thanks for addressing the comments.

There are still some small nits, but we can address later in a follow-up (or tell me later if I am not correct):
- Drop dead members `mX`, `mU`, `mUPrev` (and their allocs in `setParameters`), seem that static admittance never integrates state.
- Remove unused includes: `MNATearInterface.h` (both), `PFSolverInterfaceBranch.h` (I-type only), because looks neither is inherited.
- In `calculateAdmittance`, `H_inv` actually holds `(jωI−A)` before inversion, is possible collapse to a single `H = (…).inverse().eval()` and fix the name.

PR #588, opened 2026-07-21
Feat: Implemented SSN_SynchronGenerator

matthiasmees:
- Implemented SynchronGenerator for SSN according to model from https://ieeexplore.ieee.org/abstract/document/8536236 
- Implemented cxx example and notebook to test Grid-connected LoadStep


dpsim/examples/cxx/Components/EMT_SSN_SynchronousGenerator_Test.cpp
```diff
@@ -0,0 +1,375 @@
...
+  auto systemEMT = SystemTopology(
+      nominalFrequency, SystemNodeList{nGrid, nMachine, nDisturbance},
+      SystemComponentList{slack, gridConnection, generator, loadSwitch,
+                          disturbanceLoad});
+
+  systemEMT.initWithPowerflow(systemPF, Domain::EMT);
+
+  // Preserve the initial operating-point metadata used by the component.
+  generator->terminal(1)->setPower(
+      Complex(-generatedActivePower, -generatedReactivePower));
```
georgii-tishenin, line 317: Just to take note (does not need to be solved in this PR). 
We need to harmonize the SSN generator and `SP::Ph1::SynchronGenerator`, so that we can initialize SSN generator, like in this example: https://github.com/sogno-platform/dpsim/blob/master/dpsim/examples/cxx/Circuits/DP_Ph1_IEEE9_4Order.cpp

As I understand the problem, the `SystemTopology::initWithPowerflow()` finds the dynamic component with the same name and always writes the solved generator power to `comp->terminals()[0]`
While we connect the new generator as `generator->connect({EMT::SimNode::GND, nMachine});`

And the initialization of SSN generator uses `const Complex targetPower = terminal(1)->singlePower();`

dpsim-models/src/EMT/EMT_Ph3_SSN_SynchronousGenerator.cpp
```diff
@@ -0,0 +1,639 @@
...
+
+  mResistanceMatrix.setZero();
+  mResistanceMatrix.diagonal() << mStatorResistance, mStatorResistance,
+      mFieldResistance, mDamperResistanceD, mDamperResistanceQ1,
+      mDamperResistanceQ2;
+}
+
+void EMT::Ph3::SSN_SynchronousGenerator::setFieldVoltage(Real fieldVoltage) {
+  mFieldVoltage = fieldVoltage;
+}
```
georgii-tishenin, line 165: Seems to be set in `setParameters().` Is there a need for setting this separately? 

Also this field voltage is more like initial guess for the iterative initialization, right? Then gets overwritten in `initializeFromNodesAndTerminals()`? Maybe we can point that out somehow, e.g. name the argument `fieldVoltageInitialGuess`?

dpsim-models/src/EMT/EMT_Ph3_SSN_SynchronousGenerator.cpp
```diff
@@ -0,0 +1,639 @@
...
+
+void EMT::Ph3::SSN_SynchronousGenerator::setMechanicalTorque(
+    Real mechanicalTorque) {
+  mMechanicalTorque = mechanicalTorque;
+  mAutoInitializeMechanicalTorque = false;
+}
```
georgii-tishenin, line 171: same here: both parameters are set in `setParameters()`. Do we need an extra function for that?

georgii-tishenin (commented):
Hi @matthiasmees. Really nice work!
I left a few comments I wanted to discuss before merging (does not mean that needs to be addressed in this PR)

PR #589, opened 2026-07-21
matpower reader: map generators on PQ-labeled and multi-generator buses

leonardocarreras:
The matpower reader dropped generation in three cases:

1. Generators on a bus labeled type 1 (PQ) were never mapped, so their generation silently disappeared. Some exports label a generator's bus PQ even when it carries real generation.
2. Multiple generators sharing a bus collapsed into a single `SynchronGenerator`, keeping only the first generator's rating.
3. Offline generators (`status == 0`) were still instantiated and injected power as if online.

### DIFF
--- a/python/src/dpsim/matpower.py
+++ b/python/src/dpsim/matpower.py
@@ -732,9 +732,6 @@ class Reader:
                 q_limit_min=gen_q_min,
             )
             gen.set_base_voltage(gen_baseV)
-            # modifyPowerFlowBusType throws for PQ, set_parameters already stored it
-            if bus_type != dpsimpy.PowerflowBusType.PQ:
-                gen.modify_power_flow_bus_type(bus_type)
         else:
             # get dynamic data of the generator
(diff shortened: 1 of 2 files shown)


python/src/dpsim/matpower.py
```diff
@@ -636,40 +659,72 @@ def get_node_name(self, node_index):
...
         with_tg=True,
         with_avr=True,
     ):
-        #
-        gen_name = "Gen_N" + str(bus_index)
-
-        # relevant data from self.mpc_gen_data. Identification with bus number available in mpc_bus_data and mpc_gen_data
-        gen_data = self.mpc_gen_data.loc[
+        # A bus may carry more than one physical MATPOWER generator (e.g.
+        # buses 921, 41826, 42002, 2065, 123499, 123520 in
```
pipeacosta: The example buses and example grid should be removed from this comment.

pipeacosta (commented):
Hi @leonardocarreras, I have just one small observation for now.

PR #593, opened 2026-07-22
Analyze fork PRs with SonarCloud without exposing secrets

leonardocarreras:
The SonarCloud workflow used `pull_request_target` and checked out the PR head SHA. `actions/checkout@v6` now refuses this by default (the "pwn request" guard), so the check fails on fork PRs.

Split the analysis into two stages so fork PRs are analyzed again, without running untrusted code with secrets in scope.

Need to be merged to verify if it works.

### DIFF
--- a/.github/workflows/sonar_cloud_scan.yaml
+++ b/.github/workflows/sonar_cloud_scan.yaml
@@ -1,0 +1,115 @@
+name: Sonar Cloud Scan
+
+# Privileged stage: triggered by the completion of "Sonar Cloud Build". It runs
+# in the base-repo context with SONAR_TOKEN available, but only consumes the
+# build artifact and runs sonar-scanner (which parses sources, it does not build
+# or execute them). No untrusted PR code is executed here, so a fork PR cannot
+# reach the secret. projectKey/organization are pinned so a fork cannot redirect
+# the analysis to another project.
+
+on:
+  workflow_run:
+    workflows: ["Sonar Cloud Build"]
+    types:
+      - completed
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.workflow_run.head_branch }}
+  cancel-in-progress: true
+
+jobs:
+  scan:
+    name: Prepare and run Sonar Scan
+    runs-on: ubuntu-latest
+    container: sogno/dpsim:dev
+    if: github.event.workflow_run.conclusion == 'success'
+    permissions:
+      contents: read
+      actions: read
+    env:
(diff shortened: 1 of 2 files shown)


.github/workflows/sonar_cloud_scan.yaml
```diff
@@ -0,0 +1,91 @@
...
+          path: /tmp/sonar-dl
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Extract workspace
+        run: tar xzf /tmp/sonar-dl/sonar-workspace.tgz -C "$GITHUB_WORKSPACE"
+
+      - name: Load analysis metadata
+        id: meta
+        run: cat "$GITHUB_WORKSPACE/sonar-metadata.env" >> "$GITHUB_OUTPUT"
```
georgii-tishenin: Comment from LLM:
"This metadata comes from the PR-controlled artifact but later affects the scan mode and arguments. Could we derive it from github.event.workflow_run or the GitHub API instead, so the privileged scan is not controlled by untrusted input?"

.github/workflows/sonar_cloud_scan.yaml
```diff
@@ -0,0 +1,91 @@
...
+        with:
+          node-version: 24
+
+      - name: Setup sonar cache
+        uses: actions/cache@v5
+        with:
+          path: |
+            .scannerwork
+            sonar-cache
+          key: sonar-cache-${{ steps.meta.outputs.head_sha }}
```
georgii-tishenin, line 83: Comment from LLM:
"Could we disable this cache for PR runs? The artifact may already contain these directories, so the privileged workflow could otherwise persist PR-controlled files and reuse them in later trusted runs."

georgii-tishenin (commented):
Hi @leonardocarreras. Thanks for keeping the CI afloat😊. As github workflows is not really my strong side, I reviewed with the help of LLM and left two comments for you to double-check before merging.

.github/workflows/sonar_cloud_scan.yaml
leonardocarreras: Fixed, now resolved via API.

.github/workflows/sonar_cloud_scan.yaml
leonardocarreras: Skipped in PRs

georgii-tishenin (approved):

PR #595, opened 2026-07-23
ci: tier LLM reviewer models across finders and escalating verify gates

leonardocarreras:
## Why

The reviewer currently runs find and verify on a single model, so a weak model's confident-but-wrong findings survive to the review. On a real PR this produced a wall of high-severity false positives, including control-theory sign-convention claims that were simply misreadings of the code.

## What

Split the pipeline into tiers, all driven by env vars and fully optional (leave them unset and behaviour is identical to today's single model):

- **Finders** (`LLM_MODEL`, optional `LLM_MODEL_2`): cheap/high-throughput models that deliberately over-generate candidates. Cross-model agreement is recorded during de-duplication.
- **Mid-tier verify** (`LLM_VERIFY_MODEL`): grounds every candidate against the real source, one focused call per changed file, uncapped so no file is skipped. This does the real culling.
- **Final gate** (`LLM_FINAL_MODEL`): re-checks only the survivors in a single batched call, so the strongest (and most rate-limited) model spends the least quota. A survivor a stronger tier confirms sheds its tentative flag.

### DIFF
--- a/.github/llm-review/review.py
+++ b/.github/llm-review/review.py
@@ -5,15 +5,23 @@
 Multi-stage LLM PR reviewer for DPsim, driven by an OpenAI-compatible endpoint
 (KI:connect / RWTHgpt). Pure stdlib, no pip installs needed on the runner.
 
-Flow: diff + full source of changed files -> STAGES (finders) -> dedup ->
-per-file verification against the real code -> synthesis -> one non-blocking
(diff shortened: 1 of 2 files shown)


georgii-tishenin (approved):
Thanks @leonardocarreras. Let's try this out.

PR #600, opened 2026-07-27
Feat: SSN_DC domain components

matthiasmees:
## Summary

This PR introduces initial support for scalar DC networks in the DPsim EMT
domain.

The implementation allows DC nodes and basic DC components to be used directly
inside EMT simulations without representing the DC system as an artificial
single-phase AC network and without requiring a separate DC power-flow solver.

The main objective is to provide a clean foundation for future mixed AC/DC
models, including EMT::Ph3 converter models connected to an explicitly modeled
DC grid.

### Scalar DC phase type

- Added support for `PhaseType::DC`.
- DC nodes use one real-valued electrical quantity represented through the
  existing EMT node and MNA infrastructure.
- Added the corresponding Python enum binding as `dpsimpy.PhaseType.DC`.

### DC component namespace

Added the following ideal source components under `EMT::DC`:

- `EMT::DC::VoltageSource`
- `EMT::DC::CurrentSource`

The components use scalar DC voltages and currents and follow the passive sign
convention used by the existing DPsim component interfaces.

### DC SSN components

Added the following State-Space Nodal components under `EMT::DC::SSN`:

- `Resistor`
- `Capacitor`
- `Inductor`
- `PiLine`

The components are integrated into the existing EMT MNA/SSN solution process.
No additional DC solver is introduced.

The DC Pi-line contains:

- a series resistance;
- a series inductance;
- sending-end and receiving-end shunt capacitances;
- optional shunt conductance;
- configurable initial series current.

### Initialization

The dynamic components support explicit initial conditions, including:

- initial node voltages;
- initial inductor current;
- initial Pi-line series current.

This allows DC systems to start from a calculated steady-state operating point
instead of producing an artificial energization transient from zero initial
conditions.

### Python bindings

Added pybind bindings for:

```text
dpsimpy.PhaseType.DC

dpsimpy.emt.dc.VoltageSource
dpsimpy.emt.dc.CurrentSource

dpsimpy.emt.dc.ssn.Resistor
dpsimpy.emt.dc.ssn.Capacitor
dpsimpy.emt.dc.ssn.Inductor
dpsimpy.emt.dc.ssn.PiLine


leonardocarreras:
Hi Matthias! Some small requests before checking with more detail:

- Can you please rebase this on top of current master and not include the #588 inside? Or do you need something from it here? In that case mark that the other is needed.
- Can you rewrite the history so your dev-tool data is not in the commits (markdown planning file, gitignore changes, .codex folder, etc)
- Can you add the c++ files to TEST_SOURCES if you plan to have a notebook to use them as tests? We do not want a cmake run in the repo notebooks
- Can you take care of signing the commits again (some DCO problem with one of them)

And then a more general question about what difference would we have/what we gain on top of EMT::Ph1 if we do things the way you propose here

Would also be good to know what does @georgii-tishenin thinks about it

PR #601, opened 2026-07-28
Feat: Add option to map synchronous machine to ideal voltage source in matpower

pipeacosta:
This PR adds an option to choose the synchronous machine model to map to when importing a dynamic grid in matpower, and adds support to map it as an ideal voltage source.

### DIFF
--- a/python/src/dpsim/matpower.py
+++ b/python/src/dpsim/matpower.py
@@ -291,6 +291,7 @@ class Reader:
         with_avr=True,
         with_tg=True,
         filter_out_of_service=False,
+        generator_model=None,
     ):
         """
         Create dpsim objects with the data contained in the mpc files.
@@ -411,6 +412,7 @@ class Reader:
                     with_pss=with_pss,
                     with_tg=with_tg,
                     with_avr=with_avr,
+                    gen_model=generator_model,
                 )
 
                 # check if there is a load connected to PV bus (and create it)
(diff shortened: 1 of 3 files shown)


python/src/dpsim/matpower.py
```diff
@@ -588,24 +590,26 @@ def create_dpsim_objects(
 
                 # create dpsim component
                 trafo = self.dpsimpy_components.Transformer(transf_name, log_level)
+
+                # The transformer snubbers are sized as a fraction of rated power, so
```
leonardocarreras: Can this comment be summarized a bit and explained the difference with the previous way of doing it?

python/src/dpsim/matpower.py
```diff
@@ -695,106 +700,112 @@ def map_synchronous_machine(
...
-                    Tq0_s=Tq0_s,
-                    Taa=0,
-                )
-            elif gen_model == 6:
-                gen = self.dpsimpy_components.SynchronGenerator6bOrderVBR(
+            gen_model_data = self.mpc_dyn_gen_data["model"][gen_dyn_row_idx]
+            if gen_model is None:
+                gen_model = int(gen_model_data)
+
+            if gen_model == 1:
```
leonardocarreras: can we have an enum for this, or something more verbose?

leonardocarreras (commented):
Hi Andres! I have just small questions/comments

python/src/dpsim/matpower.py
pipeacosta: Done as an enum.

python/src/dpsim/matpower.py
pipeacosta: The comment was moved from the `else` part because now the `set_parameters` of the EMT transformer requires to specify `rated_power`. Since the "rateA" parameter is taken from static data, it applies to every dynamic simulation domain. In addition, it is still the case that we must avoid `rated_power=0`, since that would lead to a NaN snubber resistance, see e.g. https://github.com/sogno-platform/dpsim/blob/master/dpsim-models/src/EMT/EMT_Ph3_Transformer.cpp#L60. The comment will be slightly changed for more clarity and consiceness.

leonardocarreras (approved):
Thanks Andres for the work! Two things that will be nice to edit in the description: this also fixes the EMT transformer call (master passed 6 args to a 7-arg binding), and the model column from the file is now used instead of being forced to 4th order. Please resolve the comments with a short comment and feel free to merge if all passes in the CI.

PR #605, opened 2026-07-30
chore: strip saved outputs and backfill cell ids across all notebooks

leonardocarreras:
Mechanical pass, no notebook content changes: outputs cleared and missing cell ids backfilled across 113 notebooks. Split out of #509 so the contribution policy there stays reviewable. The pre-commit hook and CI guard that keep this true land with #509 and scope themselves to notebooks a pull request touches, so this can merge in either order. Regenerated rather than rebased if it goes stale.


pipeacosta (approved):
Thanks for the PR @leonardocarreras. Everything looks fine to me.

PR #607, opened 2026-07-30
ci: enforce conventional commits for commit messages

leonardocarreras:
Adds a commitlint config, a `commit-msg` hook and a CI check over every commit in a pull request. Pull request titles are not linted; linting both is what produced the duplicate changelog entries in the #608 dry run.

217 of the 326 commits since v1.2.1 do not parse today. This is a prerequisite for #608, which derives the version and the changelog from these messages.

The check is a job in `checks.yaml`, so it runs through the entry point from #661 and the fork pipeline from #674.


georgii-tishenin (approved):
Thanks @leonardocarreras, let's try that out.

PR #608, opened 2026-07-30
ci: manage releases and version bumps with release-please

leonardocarreras:
Replaces manual version bumps with release-please. It opens a pull request that bumps CMakeLists.txt and pyproject.toml and maintains CHANGELOG.md, and merging that pull request is the release decision. The manifest starts at 1.2.1, matching both files and the last tag.

CHANGELOG.md is kept, not replaced. changelog-sections maps the conventional commit types onto the Keep a Changelog headings the file already uses. The version header line cannot be matched and will differ from the existing entries.

A tag pushed with GITHUB_TOKEN starts no workflow run, so the workflow dispatches publish_to_pypi.yaml at the new tag instead. That workflow gains a workflow_dispatch trigger; since #661 it no longer runs on pull requests, so no event gating is needed on the publish jobs. Publishing to PyPI still goes through the pypi environment and its reviewer approval. No personal access token is needed.

target-branch is pinned to master so a manual dispatch cannot release unmerged code. The generated changelog trips MD012, so it is excluded from the markdownlint hook. SPDX for it belongs with the REUSE rollout in #602.

Depends on the conventional commits enforcement pull request, which has to land first: without it the version is derived from only the commits that happen to parse.


.github/workflows/release-please.yaml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+      actions: write
+    steps:
+      - name: Open or update the release pull request
+        id: release
+        uses: googleapis/release-please-action@v4
+        with:
```
georgii-tishenin, line 27: I did some additional AI review. It suggested this:
```
with:
  token: ${{ secrets.GITHUB_TOKEN }}
  target-branch: master
```
Motivation: release-please currently uses `secrets.RELEASE_PLEASE_TOKEN || secrets.GITHUB_TOKEN`, but the workflow always manually dispatches the PyPI publishing workflow after creating a release. If `RELEASE_PLEASE_TOKEN` is configured as a PAT/App token, the tag push may trigger the publish workflow, and the explicit workflow_dispatch can trigger it again.

georgii-tishenin (approved):
Hi @leonardocarreras, looks good. I added an optional comment, just resolve it if does not make sense.

.github/workflows/release-please.yaml
leonardocarreras: `RELEASE_PLEASE_TOKEN` isn't set, so it already falls back to `GITHUB_TOKEN` at the moment. I think we can see if it ever happen the situation.

PR #610, opened 2026-07-30
fix(docker): build libcimpp from source on ARM64/macOS

### LINKED ISSUE #609: Fix Docker build failure on macOS (ARM64) for libcimpp
### Description
Building the Docker container on macOS (Apple Silicon / ARM64 architecture) fails. The current Dockerfile attempts to install a precompiled `libcimpp` RPM package, which is compiled specifically for `x86_64` Linux. Because Docker on macOS runs on ARM64, the precompiled binary is incompatible.

### Expected Behavior
The Docker container should build successfully regardless of whether the host system is running on `x86_64` or `ARM64` (macOS).

### Current Behavior
The Docker build fails during the installation of the precompiled `libcimpp` RPM package on ARM64 macOS hosts.

### Proposed Solution
Modify the `Dockerfile` to detect the host architecture/platform:
- On `x86_64` Linux hosts: Continue using the precompiled RPM package.
- On macOS / `ARM64` hosts: Build `libcimpp` directly from source instead of installing the precompiled package.

---

*Note: A Pull Request (PR) implementing this fix will be submitted shortly.*

AnasBahr:
## Description
Resolves issue regarding Docker build failures on macOS/ARM64 architectures. 

The current build process installs a precompiled `libcimpp` RPM package target built for `x86_64` Linux systems. On ARM64 macOS hosts, this binary is architecture-incompatible and breaks the image build.

This PR adds architecture detection inside the `Dockerfile` to dynamically build `libcimpp` from source when running on ARM64 / macOS hosts, while preserving the precompiled package installation for standard `x86_64` targets.

## Changes Made
- Updated `Dockerfile` to detect host platform/architecture (`x86_64` vs. `ARM64`).
- Added conditional build steps to compile `libcimpp` from source on `ARM64` environments.

## Related Issue'
Closes #609

### DIFF
--- a/packaging/Docker/Dockerfile.dev
+++ b/packaging/Docker/Dockerfile.dev
@@ -72,19 +72,29 @@ RUN pip3 install --no-cache-dir --upgrade wheel build setuptools packaging && \
 	rm /tmp/requirements.txt && \
 	pip3 cache purge
 
-# Install libcimpp from prebuilt RPMs
+# Install libcimpp from prebuilt RPMs (x86_64 only) or build from source
 RUN set -eux; \
 	case "${CIM_VERSION}" in \
 		CGMES_2.4.15_16FEB2016|CGMES_2.4.15_27JAN2020|CGMES_2.4.13_18DEC2013) \
-			libcimpp_pkg="libcimpp_${CIM_VERSION}-${LIBCIMPP_VERSION}-Linux.rpm" ;; \
+			libcimpp_pkg="libcimpp_${CIM_VERSION}-${LIBCIMPP_VERSION}-Linux.rpm" \
+			libcimpp_url="https://github.com/sogno-platform/libcimpp/releases/download/release%2Fv${LIBCIMPP_VERSION}/${libcimpp_pkg}" ;; \
 		*) \
-			echo "Unsupported CIM_VERSION '${CIM_VERSION}' for prebuilt libcimpp package" >&2; \
+			echo "Unsupported CIM_VERSION '${CIM_VERSION}' for libcimpp package" >&2; \
 			exit 1 ;; \
 	esac; \
-	libcimpp_url="https://github.com/sogno-platform/libcimpp/releases/download/release%2Fv${LIBCIMPP_VERSION}/${libcimpp_pkg}"; \
-	curl -sSL "${libcimpp_url}" -o /tmp/libcimpp.rpm; \
-	dnf -y install /tmp/libcimpp.rpm; \
-	rm -f /tmp/libcimpp.rpm; \
+	if [ "$(uname -m)" = "x86_64" ]; then \
+		curl -sSL -L --max-time 300 --retry 3 --retry-delay 5 "${libcimpp_url}" -o /tmp/libcimpp.rpm || (echo "Failed to download ${libcimpp_url}" >&2; exit 1); \
+		dnf -y install /tmp/libcimpp.rpm; \
+		rm -f /tmp/libcimpp.rpm; \
+	else \
+		echo "Building libcimpp from source for $(uname -m)"; \
+		dnf -y install graphviz-devel libxml2-devel; \
+		cd /tmp && git clone --branch "v${LIBCIMPP_VERSION}" --recurse-submodules --depth 1 https://github.com/sogno-platform/libcimpp.git && \
+		mkdir -p libcimpp/build && cd libcimpp/build && \
+		cmake ${CMAKE_OPTS} .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_LIBDIR=/usr/local/lib64 && \
+		make ${MAKE_OPTS} install && \
+		rm -rf /tmp/libcimpp; \
+	fi; \
 	ldconfig; \
 	dnf clean all; \
 	rm -rf /var/cache/dnf /var/cache/yum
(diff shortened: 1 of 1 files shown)


packaging/Docker/Dockerfile.dev
```diff
@@ -72,19 +72,29 @@ RUN pip3 install --no-cache-dir --upgrade wheel build setuptools packaging && \
...
+		rm -f /tmp/libcimpp.rpm; \
+	else \
+		echo "Building libcimpp from source for $(uname -m)"; \
+		dnf -y install graphviz-devel libxml2-devel; \
+		cd /tmp && git clone --branch "v${LIBCIMPP_VERSION}" --recurse-submodules --depth 1 https://github.com/sogno-platform/libcimpp.git && \
+		mkdir -p libcimpp/build && cd libcimpp/build && \
+		cmake ${CMAKE_OPTS} .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_LIBDIR=/usr/local/lib64 && \
+		make ${MAKE_OPTS} install && \
+		rm -rf /tmp/libcimpp; \
+	fi; \
```
leonardocarreras, line 97:
```suggestion
else \
		cd /tmp && git clone --branch "release/v${LIBCIMPP_VERSION}" --recurse-submodules --depth 1 https://github.com/sogno-platform/libcimpp.git && \
		mkdir -p libcimpp/build && cd libcimpp/build && \
		cmake ${CMAKE_OPTS} .. \
			-DCMAKE_BUILD_TYPE=Release \
			-DCMAKE_INSTALL_LIBDIR=/usr/local/lib64 \
			-DUSE_CIM_VERSION=${CIM_VERSION} \
			-DBUILD_SHARED_LIBS=ON \
			-DBUILD_ARABICA_EXAMPLES=OFF \
			-DCIMPP_BUILD_DOC=OFF && \
		make ${MAKE_OPTS} install && \
		cd / && rm -rf /tmp/libcimpp; \
fi; \
```

packaging/Docker/Dockerfile.dev
```diff
@@ -72,19 +72,29 @@ RUN pip3 install --no-cache-dir --upgrade wheel build setuptools packaging && \
...
-			echo "Unsupported CIM_VERSION '${CIM_VERSION}' for prebuilt libcimpp package" >&2; \
+			echo "Unsupported CIM_VERSION '${CIM_VERSION}' for libcimpp package" >&2; \
 			exit 1 ;; \
 	esac; \
-	libcimpp_url="https://github.com/sogno-platform/libcimpp/releases/download/release%2Fv${LIBCIMPP_VERSION}/${libcimpp_pkg}"; \
-	curl -sSL "${libcimpp_url}" -o /tmp/libcimpp.rpm; \
-	dnf -y install /tmp/libcimpp.rpm; \
-	rm -f /tmp/libcimpp.rpm; \
+	if [ "$(uname -m)" = "x86_64" ]; then \
+		curl -sSL -L --max-time 300 --retry 3 --retry-delay 5 "${libcimpp_url}" -o /tmp/libcimpp.rpm || (echo "Failed to download ${libcimpp_url}" >&2; exit 1); \
```
leonardocarreras: sSL never fails
```suggestion
		curl -fsSL --max-time 300 --retry 3 --retry-delay 5 "${libcimpp_url}" -o /tmp/libcimpp.rpm || (echo "Failed to download ${libcimpp_url}" >&2; exit 1); \
```

packaging/Docker/Dockerfile.dev
leonardocarreras: no need for the dnf line, this is part of the base image already as i understand
https://github.com/sogno-platform/dpsim/blob/9e864daa8458a08bb7d59bfe818faa51f41c88f7/packaging/Docker/Dockerfile.dev#L36-L37

packaging/Docker/Dockerfile.dev
```diff
@@ -72,19 +72,29 @@ RUN pip3 install --no-cache-dir --upgrade wheel build setuptools packaging && \
 	rm /tmp/requirements.txt && \
 	pip3 cache purge
 
-# Install libcimpp from prebuilt RPMs
+# Install libcimpp from prebuilt RPMs (x86_64 only) or build from source
 RUN set -eux; \
 	case "${CIM_VERSION}" in \
 		CGMES_2.4.15_16FEB2016|CGMES_2.4.15_27JAN2020|CGMES_2.4.13_18DEC2013) \
```
leonardocarreras, line 78: libcimpp_CGMES_3.0.0-2.2.0-Linux.rpm exists, but we have not tested in DPsim

leonardocarreras (changes requested):
Hi @AnasBahr, i think this does not solve all what you mention in #609 as cases (it solves only for one container), as for many of the containers is done this way (ubuntu, rocky, etc). However, that can be a follow up.
Some comments:
- tag for `v2.2.0` seems to be `release/v2.2.0`, maybe that is why it does not work, that might be pre-existing (does it work on the branch)
Image

- when compiling, is important to keep `CIM_VERSION` to select the CGMES

packaging/Docker/Dockerfile.dev
AnasBahr: Hi Leo, thanks for your comments. I have applied your previous suggestion. What would be the next step?

AnasBahr: I didn't add these lines. They already existed. Do you think it is a good idea to remove them?

leonardocarreras: Hi Anas, I think we can leave it for a follow-up

leonardocarreras (approved):

leonardocarreras:
@AnasBahr I will merge, please test and if not, make a follow-up

PR #611, opened 2026-07-31
Fix EMT::Ph3 averaged VSI without a connection transformer (#604)

### LINKED ISSUE #604: EMT::Ph3::AvVoltageSourceInverterDQ segfaults with withTrafo=false: mVirtualNodes[3] dereferenced when only 3 are allocated
### Summary

`EMT::Ph3::AvVoltageSourceInverterDQ` constructed with `withTrafo = false` segfaults during
initialisation. On that branch the constructor allocates **three** virtual nodes, but four sites
dereference `mVirtualNodes[3]` unconditionally.

Tested at `1b063c54298a46645c04c5ca156ad8178e37f486` (main, 2026-07-21), built from source with
Python bindings, EMT Ph3 / MNA, `dt = 100 us`.

### Reproducer

`examples/Notebooks/Circuits/EMT_Slack_PiLine_VSI_with_PF_Init.ipynb` runs clean as committed. Change
only the `with_trafo` argument:

```python
pv = dpsimpy.emt.ph3.AvVoltageSourceInverterDQ(
    "pv", "pv", dpsimpy.LogLevel.debug, with_trafo=False   # was True
)
```

and remove the now-inapplicable `set_transformer_parameters(...)` call. The process dies with SIGSEGV
after `MnaSolverFactory ... creating KLUAdapter solver implementation`, i.e. inside
`initializeFromNodesAndTerminals`, before the first step. Nothing else in the notebook is touched.

### Cause

`dpsim-models/src/EMT/EMT_Ph3_AvVoltageSourceInverterDQ.cpp`, constructor:

```cpp
if (withTrafo) {
  setVirtualNodeNumber(4);
  ...
} else {
  setVirtualNodeNumber(3);
}
```

With the transformer, index 3 is the filter-interface node. Without it, index 3 is one past the end.
These four dereferences are not guarded by `mWithConnectionTransformer`:

- `L277`  `Real theta = std::arg(mVirtualNodes[3]->initialSingleVoltage());`
- `L291`  `matrixStateInit(0, 0) = std::arg(mVirtualNodes[3]->initialSingleVoltage());`
- `L292`  `matrixOutputInit(0, 0) = std::arg(mVirtualNodes[3]->initialSingleVoltage());`
- `L413`  `vcdq = parkTransformPowerInvariant(theta, **mVirtualNodes[3]->mVoltage);`

The first three are in `initializeFromNodesAndTerminals` (where the crash lands); `L413` is in the
per-step control update, so it would fire every step if initialisation were survived.

Note the surrounding code already handles the no-transformer case correctly — e.g. `L263-266` picks
the terminal node instead of `mVirtualNodes[3]` for `mSubResistorC`, and `L217-236` branches properly
on the initial-voltage calculation. It appears only these four references were missed.

### Suggested fix

In each of the four places, use the node that is electrically equivalent when there is no transformer,
which is the one `mSubResistorC` is already connected to on that branch — `mTerminals[0]->node()`.
Something of the shape:

```cpp
auto filterIfNode = mWithConnectionTransformer
                        ? mVirtualNodes[3]
                        : mTerminals[0]->node();
```

then use `filterIfNode` at all four sites. That keeps the with-transformer path bit-identical.

### Why it matters

For a low-voltage island there is no step-up transformer to model, so `withTrafo = false` is the
physically correct construction, and it is reachable from the Python bindings with no warning. A
guard, or an explicit "not supported" throw, would at least turn a segfault into something
diagnosable.

Happy to open a PR if the suggested shape looks right.

ysrji:
Fixes #604.

### The segfault

`AvVoltageSourceInverterDQ` allocates three virtual nodes when `withTrafo` is false (`setVirtualNodeNumber(3)`, leaving indices 0..2) while four sites dereference `mVirtualNodes[3]` unconditionally — three in `initializeParentFromNodesAndTerminals`, one in `controlStep`.

This resolves the control's measurement node once during initialization instead: virtual node 3 with a connection transformer, `mTerminals[0]->node()` without one. Without a transformer the filter interface *is* the component terminal, which is already how `mSubResistorC` is connected a few lines above. `EMT_Ph3_VSIVoltageControlVCO` was the reference, as suggested.

### A second defect on the same branch

While confirming the fix I found that `mnaCompUpdateCurrent` publishes `mSubResistorC->mIntfCurrent` unnegated:

```cpp
if (mWithConnectionTransformer)
  **mIntfCurrent = mConnectionTransformer->mIntfCurrent->get();
else
  **mIntfCurrent = mSubResistorC->mIntfCurrent->get();
```

The transformer is connected `{terminal 0, virtual node 3}` and Rc is connected `{virtual node 2, terminal 0}` — opposite orientations — so the two branches report opposite signs for the same physical current. `controlStep()` already applies exactly this negation when it feeds the same attribute to the Park transform, which I took as the intended convention.

I've included the one-line fix here since it is the same root cause (the branch was never exercised) and within the EMT::Ph3 scope you set, but I'm happy to split it out if you'd rather review it separately.

### Verification

Both paths on the same network — 400 V, 12 kW load, inverter at a +4 kW setpoint behind a short LV feeder — run for 2000 steps at `dt = 100 us`:

| | before | after |
|---|---|---|
| `with_trafo=True` | +4472.5 W, load bus 252.2 V | unchanged |
| `with_trafo=False` | segfault | +4480.4 W, load bus 252.2 V |

The load bus voltage agreeing to within a volt across the two paths is what says the sign difference was in the published attribute and not in the network solution.

### Scope

EMT::Ph3 only, as discussed. SP and DP have the same node-count split and can follow as separate PRs — happy to do those next if this shape looks right to you.

### DIFF
--- a/dpsim-models/include/dpsim-models/EMT/EMT_Ph3_AvVoltageSourceInverterDQ.h
+++ b/dpsim-models/include/dpsim-models/EMT/EMT_Ph3_AvVoltageSourceInverterDQ.h
@@ -55,6 +55,12 @@ protected:
   std::shared_ptr<EMT::Ph3::Resistor> mSubResistorC;
   /// Optional connection transformer
   std::shared_ptr<EMT::Ph3::Transformer> mConnectionTransformer;
+  /// Node the control measures: the transformer's low-voltage virtual node
+  /// when a connection transformer is present, and the component's own
+  /// terminal when it is not. Without a transformer there is no virtual node 3
+  /// -- setVirtualNodeNumber(3) leaves indices 0..2 -- so the control cannot
+  /// name it. Resolved once during initialization and used by controlStep().
+  SimNode::Ptr mControlNode;
 
   /// Flag for connection transformer usage
   Bool mWithConnectionTransformer = false;
--- a/dpsim-models/src/EMT/EMT_Ph3_AvVoltageSourceInverterDQ.cpp
+++ b/dpsim-models/src/EMT/EMT_Ph3_AvVoltageSourceInverterDQ.cpp
@@ -265,6 +265,13 @@ void EMT::Ph3::AvVoltageSourceInverterDQ::initializeParentFromNodesAndTerminals(
   else
     mSubResistorC->connect({mVirtualNodes[2], mTerminals[0]->node()});
 
+  // The control measures the filter interface. With a connection transformer
+  // that is virtual node 3; without one the filter interface IS the component
+  // terminal, and virtual node 3 does not exist -- setVirtualNodeNumber(3)
+  // above leaves only indices 0..2, so naming it is out-of-range access.
+  mControlNode = mWithConnectionTransformer ? mVirtualNodes[3]
+                                            : mTerminals[0]->node();
+
   // Initialize electrical subcomponents
   for (auto subcomp : mSubComponents) {
     subcomp->initialize(mFrequencies);
@@ -274,7 +281,7 @@ void EMT::Ph3::AvVoltageSourceInverterDQ::initializeParentFromNodesAndTerminals(
   // Initialize control subcomponents
   // current and voltage inputs to PLL and power controller
   Matrix vcdq, ircdq;
-  Real theta = std::arg(mVirtualNodes[3]->initialSingleVoltage());
+  Real theta = std::arg(mControlNode->initialSingleVoltage());
   vcdq =
       parkTransformPowerInvariant(theta, filterInterfaceInitialVoltage.real());
   ircdq = parkTransformPowerInvariant(
@@ -288,8 +295,8 @@ void EMT::Ph3::AvVoltageSourceInverterDQ::initializeParentFromNodesAndTerminals(
   // angle input
   Matrix matrixStateInit = Matrix::Zero(2, 1);
   Matrix matrixOutputInit = Matrix::Zero(2, 1);
-  matrixStateInit(0, 0) = std::arg(mVirtualNodes[3]->initialSingleVoltage());
-  matrixOutputInit(0, 0) = std::arg(mVirtualNodes[3]->initialSingleVoltage());
+  matrixStateInit(0, 0) = theta;
+  matrixOutputInit(0, 0) = theta;
   mPLL->setInitialValues(**mVcq, matrixStateInit, matrixOutputInit);
 
   SPDLOG_LOGGER_INFO(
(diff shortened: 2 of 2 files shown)


leonardocarreras:
Hi @ysrji, if you agree to contribute, you should sign-off (this is to attest that you agree with the licensing and the terms of how dpsim works). The DCO is [missing](https://github.com/sogno-platform/dpsim/pull/611/checks?check_run_id=91057971212).

dpsim-models/include/dpsim-models/EMT/EMT_Ph3_AvVoltageSourceInverterDQ.h
```diff
@@ -55,6 +55,12 @@ class AvVoltageSourceInverterDQ
   std::shared_ptr<EMT::Ph3::Resistor> mSubResistorC;
   /// Optional connection transformer
   std::shared_ptr<EMT::Ph3::Transformer> mConnectionTransformer;
+  /// Node the control measures: the transformer's low-voltage virtual node
```
leonardocarreras, line 58: This comment is too long, and additionally does not help to the documentation. Here it should be 
```cpp
/// Help text of the thing below
VariableOrClass declaratedVarOrClass; 
```

dpsim-models/src/EMT/EMT_Ph3_AvVoltageSourceInverterDQ.cpp
```diff
@@ -265,6 +265,13 @@ void EMT::Ph3::AvVoltageSourceInverterDQ::initializeParentFromNodesAndTerminals(
   else
     mSubResistorC->connect({mVirtualNodes[2], mTerminals[0]->node()});
 
+  // The control measures the filter interface. With a connection transformer
```
leonardocarreras, line 268: A shorter comment is preferable

dpsim-models/src/EMT/EMT_Ph3_AvVoltageSourceInverterDQ.cpp
```diff
@@ -470,10 +477,20 @@ void EMT::Ph3::AvVoltageSourceInverterDQ::mnaParentPostStep(
...
     const Matrix &leftvector) {
-  if (mWithConnectionTransformer)
+  if (mWithConnectionTransformer) {
+    // Transformer is connected {terminal 0, virtual node 3}, so its interface
+    // current already flows towards the terminal.
     **mIntfCurrent = mConnectionTransformer->mIntfCurrent->get();
-  else
-    **mIntfCurrent = mSubResistorC->mIntfCurrent->get();
+  } else {
+    // Rc is connected the other way round, {virtual node 2, terminal 0}, so its
```
leonardocarreras, line 485: This comment is too long

leonardocarreras (changes requested):
Hi @ysrji ! Thanks for the interest in contributing.

Here just some small requests: sign-off with `git commit -s `(for the DCO), and use the `pre-commit`. Additionally, about comments is always better if the code is clear and we do not need them. If we do, please keep the comments short (1-2 lines) related to the code, not to the tests done or the experiments to discover, or the way is solved.

About the sign of the current, that change seems correct. Thanks for that fix too.

leonardocarreras:
Hi @ysrji @3shn  this is a friendly ping

Particularly DCO is required for us, but if you can also address the comments is even better :)

PR #612, opened 2026-07-31
docs: fix site config and publish the API reference to both mirrors

leonardocarreras:
The commented-out github_subdir made the edit links 404 on every published page, the baseURL pointed at a host that now redirects, and the feedback widget still had the theme placeholder URLs. All three are fixed.

Only the GitHub Pages deploy copied the Sphinx and Doxygen output, so dpsim.fein-aachen.org returned 404 for /sphinx and /doxygen while the README pointed readers there as the main documentation. The two workflows are merged so one reference build feeds both deploys, and the main documentation links now point at sogno.energy/dpsim.

One page also loaded a script from polyfill.io, a dead domain that previously served malicious code. It duplicated KaTeX, which is already enabled site wide, so it is removed.

Verified by building the site locally with the pinned Hugo version for both baseURLs.


leonardocarreras:

PR #613, opened 2026-07-31
docs: document the Ph3 Python API and modernise the Doxygen config

leonardocarreras:
Reference.rst listed ph1 twice for each domain where the second entry was meant to be ph3, so the whole three-phase Python API was missing and the duplicated blocks produced warnings. The ph3 modules are now listed, along with base, signal and event, all confirmed to exist at runtime.

On the C++ side, EXTRACT_ALL was off, so Doxygen only documented entities carrying a doc comment. The CPS namespaces have none, so the generated reference showed a single namespace and no file pages at all, and the component namespaces were not browsable. Enabling it takes the output from 2 to 40 namespaces and from 0 to 571 file pages, with the class count unchanged at 830.

The Doxygen config was also still in 1.8.12 format with WARN_IF_UNDOCUMENTED on, which buried real problems. It is migrated to the current format with the navigation tree enabled. The build page claimed the output lands in Documentation/html, which is corrected.

Measured locally: Sphinx 248 warnings to 0 and 45 three-phase classes now documented, Doxygen 16727 warnings to 49. The remaining Doxygen warnings are genuine signature mismatches, so they stay.


leonardocarreras:

Issue #501, opened 2026-06-17
Docs: How to build in python for windows


leonardocarreras:
**Is your feature request related to a problem? Please describe.**
For Windows is not available to have a wheel yet.

**Describe the solution you'd like**
Document how to pip install directly from master.

**Describe alternatives you've considered**
Clean build of the wheel like  #368 

**Additional context**
N/A

## Status after #619

- [x] Document a source build on Windows. `Developer Guide > Build > CMake for Windows` mirrors the `build_test_windows.yaml` CI recipe, including `pip install pybind11[global]` for the bindings, and states that VILLAS and CIM are not available there.
- [ ] Document `pip install` directly from master. No `pip install .` or `pip install git+https://github.com/sogno-platform/dpsim` appears anywhere in the docs; `pip install dpsim` is the only pip path shown, and `pip install -e .` is explicitly discouraged.

leonardocarreras:
wheel is now available in PyPI, closing

PR #614, opened 2026-07-31
docs: rewrite the install and build pages against the current toolchain

leonardocarreras:
The install and build pages had drifted a long way from the repository: Visual Studio 2017, setup.py bdist_wheel, conda develop against paths that no longer exist, a Homebrew URL that returns 404, Sundials v3.1.1 where the Dockerfile builds v3.2.1, and a dependency link pinned to an old commit.

Windows now mirrors what build_test_windows.yaml runs, which also covers #501. macOS is marked as not covered by CI and points at the Apple Silicon libcimpp failure in #609. The install page leads with a venv, states that only Linux wheels are published and for which Python versions, and notes that the Docker image pins a fixed Jupyter token. The PYTHONPATH workflow contributors actually use was undocumented and is now written down.

Every path, target and option referenced was checked to exist, all external links return 200, and the site builds with the pinned Hugo version.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -48,3 +48,11 @@ repos:
     hooks:
       - id: markdownlint
         args: [-r, "~MD013,~MD033,~MD024"]
+        exclude: ^docs/hugo/content/
+      # Hugo emits the <h1> from the page front matter, so a top level header in the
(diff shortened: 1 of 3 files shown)


leonardocarreras:

PR #615, opened 2026-07-31
docs: add copy buttons to code blocks and compact the sidebar

leonardocarreras:
The docs are full of shell commands readers are meant to run, with no way to copy them. Docsy ships this already through Prism, but prism_syntax_highlighting was set to false. The sidebar is also switched to compact so it opens to the current section rather than the whole tree.

Verified against the pinned Hugo version: the Prism assets are linked, the bundled prism.js already carries the copy-to-clipboard plugin so no new dependency is needed, and the KaTeX formulas on the model and concept pages still render.


leonardocarreras:

PR #616, opened 2026-07-31
docs: correct the model inventory and contributor pages

leonardocarreras:
The models index listed a Bergeron line that exists nowhere in the tree and omitted the SP domain, all Ph3 models, the reduced-order generators and the SSN components. It is regenerated from the headers as a per-domain availability table. add-model.md pointed at a source layout that has not existed for years and now describes the real dpsim-models tree and the three registration points.

Also deletes induction-machine.md, since no such model exists and history shows the page is an orphan, fills the empty branches.md, and scopes markdownlint MD002 to non-Hugo files, since Docsy emits the h1 from front matter. Verified with pre-commit run --all-files and a Hugo build.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -48,3 +48,11 @@ repos:
     hooks:
       - id: markdownlint
         args: [-r, "~MD013,~MD033,~MD024"]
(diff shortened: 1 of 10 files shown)


leonardocarreras:

PR #618, opened 2026-07-31
docs: explain how shifted frequency analysis relates to dynamic phasors

leonardocarreras:
The dynamic phasors concept page named shifted frequency analysis in a single line and left the relationship unexplained, then promised a derivation starting from bandpass signals that the page never delivered. The Sphinx reference already titles its section "Dynamic Phasor / SFA Models", so the two terms were being used interchangeably with nothing anywhere saying why.

This gives both derivations, the analytic-signal frequency shift and the sliding-window Fourier averaging, and shows that retaining the first coefficient makes them the same envelope model, so the difference is emphasis and lineage rather than mathematics. A closing section covers the relationship to electromagnetic transient models, where the larger timestep comes from, and the carrier and phase-reference conditions that a mixed-domain interface has to respect. The page stays free of implementation detail and adds no citations beyond those already present.


leonardocarreras:

PR #423, opened 2025-11-21
Decoupling ITM

pipeacosta:
This PR implements new decoupling elements based on the Ideal Transformer Model (ITM) for exemplifying co-simulation scenarios. 

This PR renames the Python API (dpsimpy) function `add` to `add_component`.

PR #621, opened 2026-08-02
SystemTopology::removeNode: remove connected components too

### LINKED ISSUE #529: Finish `SystemTopology::removeNode`
Follow-up to #423.

`removeNode` only removes the node today and has a `TODO` saying the components connected to that node should be removed too. For now the caller has to do that by hand, which is easy to get wrong.

We should either make `removeNode` clean up the connected components, or provide a helper that does.

## Done when

- [ ] `removeNode` cleans up connected components, or there's a documented helper that does

barismuutlu:
Closes #529. Follow-up to #423.

`removeNode` now cleans up the components connected to the removed node instead of leaving that to the caller:

- Components whose terminals reference the removed node are removed from `mComponents`. Connectivity is checked through `topologicalTerminals()`; terminals that are not connected yet (`nullptr`) are skipped, so partially-wired components cannot crash the check.
- The removed node's entry in `mComponentsAtNode` is erased, and removed components are also purged from the remaining per-node lists (a two-terminal component disappears from the lists of both of its nodes).
- `SystemTopology.remove_node` is now exposed in the Python bindings next to `remove_component`.

Existing callers (e.g. the `*_split_decoupled_node` examples) remove the connected components by hand *before* calling `removeNode`; for them the new cleanup finds nothing left to remove, so their behaviour is unchanged.

One observation kept out of this PR to keep it small: `removeComponent` also leaves stale entries in `mComponentsAtNode` today. Happy to address that in a follow-up if you would like it handled the same way.

Verified with a self-contained Python check (built with the Python bindings, `PYTHONPATH=build`):

```python
import dpsimpy

gnd = dpsimpy.emt.SimNode.gnd
n1, n2 = dpsimpy.emt.SimNode("n1"), dpsimpy.emt.SimNode("n2")
vs = dpsimpy.emt.ph1.VoltageSource("vs"); vs.set_parameters(V_ref=10+0j, f_src=50)
r1 = dpsimpy.emt.ph1.Resistor("r1"); r1.set_parameters(R=1)
c1 = dpsimpy.emt.ph1.Capacitor("c1"); c1.set_parameters(C=1e-3)
vs.connect([gnd, n1]); r1.connect([n1, n2]); c1.connect([n2, gnd])

system = dpsimpy.SystemTopology(50, [gnd, n1, n2], [vs, r1, c1])
system.remove_node("n2")

assert sorted(c.name() for c in system.components) == ["vs"]
assert all(n.name() != "n2" for n in system.nodes)
assert all(n.name() != "n2" for n in system.components_at_node)
```

Also checked: removing a non-existent node stays a no-op, the old caller pattern (remove components first, then the node) is unaffected, and a component that only shares another node with the removed component keeps its entry in `components_at_node`.

### DIFF
--- a/dpsim-models/include/dpsim-models/SystemTopology.h
+++ b/dpsim-models/include/dpsim-models/SystemTopology.h
@@ -151,6 +151,8 @@ public:
   /// @param name name of the component
   void removeComponent(const String &name);
 
+  /// @brief Remove node and all components connected to it
+  /// @param name name of the node
   void removeNode(const String &name);
 
   ///
@@ -170,5 +172,8 @@ public:
 
 private:
   template <typename VarType> void multiplyPowerComps(Int numberCopies);
+  /// Removes all components connected to the given node from the
+  /// component list and from the per-node component map
+  void removeComponentsConnectedTo(const TopologicalNode::Ptr &node);
 };
 } // namespace CPS
--- a/dpsim-models/src/SystemTopology.cpp
+++ b/dpsim-models/src/SystemTopology.cpp
@@ -267,12 +267,12 @@ static bool isConnectedTo(const IdentifiedObject::Ptr &comp,
   if (!powerComp)
     return false;
 
-  for (auto terminal : powerComp->topologicalTerminals()) {
-    // Terminals may not be connected yet
-    if (terminal && terminal->topologicalNodes() == node)
-      return true;
-  }
-  return false;
+  // Terminals may not be connected yet
+  const auto terminals = powerComp->topologicalTerminals();
+  return std::any_of(terminals.begin(), terminals.end(),
+                     [&node](const TopologicalTerminal::Ptr &terminal) {
+                       return terminal && terminal->topologicalNodes() == node;
+                     });
 }
 
 void SystemTopology::removeComponentsConnectedTo(
--- a/dpsim-models/src/SystemTopology.cpp
+++ b/dpsim-models/src/SystemTopology.cpp
@@ -296,11 +296,18 @@ void SystemTopology::removeComponentsConnectedTo(
                   comps.end());
     }
   }
+
+  for (const auto &removed : removedComponents) {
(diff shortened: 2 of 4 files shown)


barismuutlu:
On the automated review's remaining points: happy to add a short docs section and a test for `removeNode` if maintainers want them — the suggested paths (`tests/python/…`, `docs/…/Models/SystemTopology.md`) don't currently exist in the repo, so a pointer to the preferred location would help. On the thread-safety note: `removeNode` follows the same single-threaded-setup assumption as the other `SystemTopology` mutators (`addNode`, `removeComponent`).

barismuutlu:
Codecov flagged the new lines as untested, so I added a regression notebook under `examples/Notebooks/Features/` (the suite CI runs) covering the cleanup, the map consistency, the no-op case, and the shared-node case. Also switched the connectivity check to `std::any_of` as Sonar suggested.

leonardocarreras (changes requested):
Hi @barismuutlu thanks for the PR and fixes taking the comments in account. It looks good to me in general. 

There are some small details, we could do a follow up or solve here:
- Guard the function so it does not delete GND node (you can use `node()->isGround()` ) otherwise could delete the whole circuit
- mTearComponents also need to be checked/updated, maybe something like
```
for (const auto &removed : removedComponents) {
    mTearComponents.erase(
        std::remove(mTearComponents.begin(), mTearComponents.end(), removed),
        mTearComponents.end());
  }
```

leonardocarreras (commented):
--- duplicated

barismuutlu:
Thanks for the review! Both points are addressed: `removeNode` now skips ground nodes, and removed components are also purged from `mTearComponents` (using your snippet). Both are covered in the regression notebook, and I verified the `*_split_decoupled_node` examples produce bit-identical results before and after the change.

leonardocarreras (approved):
Looks good, thanks for addressing the comments.

Lets wait for the CI.

PR #622, opened 2026-08-02
pybind: restore SystemTopology.add as a deprecated alias

### LINKED ISSUE #530: Keep the Python `add` method working
Keep the Python `add` method working

Follow-up to #423.

#423 renamed the Python `add` to `add_component` (and added `add_node`). The new names are nicer, but `add` was removed completely, so any existing script calling `system.add(...)` now breaks with an `AttributeError`:

```python
system.add(component)   # used to work, now raises AttributeError
```

`SystemTopology` is used everywhere, so this is a big break for a rename.

Overloading didn't work because the signatures differ, so let's keep `add` as a deprecated alias instead: either a small Python wrapper that picks the right method and emits a `DeprecationWarning`, or bind `add` next to the new names and mark it for removal later.

## Done when

- [ ] `system.add(...)` still works and emits a deprecation warning
- [ ] `add_component` / `add_node` stay the documented names

barismuutlu:
Closes #530. Follow-up to #423.

#423 renamed the Python `add` to `add_component`/`add_node`, which removed `add` entirely, so existing user scripts calling `system.add(...)` break with an `AttributeError`. This restores `add` as a deprecated alias, as suggested in the issue:

- Emits a `DeprecationWarning` pointing to `add_component`/`add_node`. The `PyErr_WarnEx` return value is checked, so running with `-W error::DeprecationWarning` raises cleanly instead of the warning being swallowed.
- Dispatches by type: nodes go to `addNode`, components to `addComponent`. Lists and tuples are dispatched per element, which covers the old `add(list)` overload.
- Anything else raises a `TypeError` with a descriptive message.
- `add_component` / `add_node` stay the documented names; nothing else in the API changes.

Also drops one duplicated `.def("add_component", ...)` line further down in the same binding block (a leftover duplicate of the definition a few lines above).

Verified with a self-contained check (`PYTHONPATH=build`):

```python
import warnings, dpsimpy

system = dpsimpy.SystemTopology(50)
n1 = dpsimpy.emt.SimNode("n1")
r1 = dpsimpy.emt.ph1.Resistor("r1"); r1.set_parameters(R=1)
r2 = dpsimpy.emt.ph1.Resistor("r2"); r2.set_parameters(R=2)

with warnings.catch_warnings(record=True) as w:
    warnings.simplefilter("always")
    system.add(r1)      # -> add_component, warns
    system.add(n1)      # -> add_node, warns
    system.add([r2])    # list form, warns
assert len(w) == 3 and all(issubclass(x.category, DeprecationWarning) for x in w)
assert sorted(c.name() for c in system.components) == ["r1", "r2"]
assert any(n.name() == "n1" for n in system.nodes)

# invalid input
try:
    system.add(42)
except TypeError:
    pass

# warning-as-error propagates
with warnings.catch_warnings():
    warnings.simplefilter("error", DeprecationWarning)
    try:
        system.add(r1)
    except DeprecationWarning:
        pass
```

`add_component`/`add_node` were also checked to still work without emitting any warning.

### DIFF
--- a/dpsim/src/pybind/main.cpp
+++ b/dpsim/src/pybind/main.cpp
@@ -24,6 +24,19 @@ PYBIND11_DECLARE_HOLDER_TYPE(T, CPS::AttributePointer<T>);
 namespace py = pybind11;
 using namespace pybind11::literals;
 
+/// Dispatches a single argument of the deprecated SystemTopology.add
+/// alias to add_node or add_component depending on its type
+static void systemTopologyAddSingle(DPsim::SystemTopology &sys,
+                                    const py::handle &item) {
+  if (py::isinstance<CPS::TopologicalNode>(item))
+    sys.addNode(item.cast<CPS::TopologicalNode::Ptr>());
+  else if (py::isinstance<CPS::IdentifiedObject>(item))
+    sys.addComponent(item.cast<CPS::IdentifiedObject::Ptr>());
+  else
+    throw py::type_error("SystemTopology.add expects a node, a component, "
+                         "or a list of nodes and components");
+}
+
 PYBIND11_MODULE(dpsimpy, m) {
   m.doc() = R"pbdoc(
   DPsim Python bindings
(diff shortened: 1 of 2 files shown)


barismuutlu:
Happy to add a deprecation note to the docs and a test for the `add` alias if wanted — the paths the automated review suggests don't currently exist, so just point me to where you'd prefer them.

barismuutlu:
Pushed two follow-ups: the alias dispatch now goes through a named `py::isinstance` helper, which resolves the Sonar quality-gate findings (lambda length, empty catch blocks, exceptions as control flow), and a regression notebook under `examples/Notebooks/Features/` covers the coverage gap Codecov flagged — including the `-W error::DeprecationWarning` path.

leonardocarreras (approved):
Looks good to me

leonardocarreras (approved):
--- duplicated

PR #630, opened 2026-08-03
Feat: Add Scalar EMT DC Components

matthiasmees:
## Summary

This PR adds native scalar DC support to the EMT MNA solver and introduces a set of reusable EMT DC components based on the existing state-space nodal approach.

The new public component API is available under:

```cpp
CPS::EMT::DC
```

and includes:

* `VoltageSource`
* `CurrentSource`
* `Resistor`
* `Capacitor`
* `Inductor`
* `PiLine`

The same components are exposed in Python under:

```python
dpsimpy.emt.dc
```

## Motivation

Previously, DC networks in DPsim were commonly represented using single-phase EMT components. Although this is possible for simple circuits, it does not explicitly represent a scalar DC domain and causes ambiguity regarding:

* phase handling;
* RMS-to-peak conversion;
* node indexing;
* initial-voltage interpretation;
* component naming and API semantics;
* future integration of coupled AC/DC models.

This PR introduces an explicit `PhaseType::DC` path while reusing the existing EMT MNA solver. DC and AC components can therefore still be used in the same simulation and solved within one MNA system.

### Scalar DC support

Added support for `PhaseType::DC` throughout the relevant topology and MNA infrastructure.

This includes updates to:

* `TopologicalNode`
* `TopologicalTerminal`
* `SimNode`
* `SimPowerComp`
* MNA node-index handling
* EMT node initialization
* Python `PhaseType` bindings

DC nodes use one scalar MNA unknown and preserve their voltage directly without applying an AC RMS-to-peak conversion.

The component matrix-index initialization was also updated so scalar node types use the matrix index assigned by the MNA solver instead of being dispatched through three-phase A/B/C indexing.

### Generic two-terminal DC SSN infrastructure

Added scalar two-terminal SSN implementations for:

* voltage-type components;
* current-type components.

These classes handle:

* continuous state-space matrices;
* trapezoidal discretization;
* history-vector calculation;
* MNA system-matrix stamps;
* right-side-vector stamps;
* interface voltage and current updates.

The public DC components directly own their physical parameters. No separate `Base::DC` parameter classes or additional public SSN namespace are introduced.

#### Resistor

Implements a scalar resistive branch with:

```cpp
setParameters(Real resistance)
```

#### Capacitor

Implements a scalar capacitor using a current-type state-space representation.

The capacitor voltage is retained as its dynamic state.

#### Inductor

Implements a scalar inductor using a voltage-type state-space representation.

An initial current can be specified for steady-state or hot-start initialization.

#### PiLine

Implements a DC Pi-equivalent line consisting of:

* one series resistance;
* one series inductance;
* one half of the total shunt capacitance at each terminal;
* one half of the total shunt conductance at each terminal.

The line internally creates the required subcomponents and virtual node.

Its parameter interface is:

```cpp
setParameters(
    Real seriesResistance,
    Real seriesInductance,
    Real parallelCapacitance,
    Real parallelConductance,
    Real initialCurrent
)
```

`parallelCapacitance` and `parallelConductance` represent the total line values and are divided equally between both line ends.

#### Voltage source

Implements an ideal scalar DC voltage source using an additional MNA branch-current unknown.

The voltage reference is stored as a dynamic attribute so it can be changed during a simulation.

#### Current source

Implements an ideal scalar DC current source.

The current reference is stored as a dynamic attribute and can be changed during a simulation.

### Sign convention

Two-terminal DC components use:

```cpp
connect({negativeTerminal, positiveTerminal});
```

The positive interface voltage is:

```text
v_intf = v_positive - v_negative
```

and positive interface current flows from the positive terminal to the negative terminal.

For a load connected from a positive DC node to ground, the connection is therefore:

```cpp
load->connect({EMT::SimNode::GND, positiveNode});
```

### Python bindings

Added the following Python module:

```python
dpsimpy.emt.dc
```

The available classes are:

```python
dpsimpy.emt.dc.VoltageSource
dpsimpy.emt.dc.CurrentSource
dpsimpy.emt.dc.Resistor
dpsimpy.emt.dc.Capacitor
dpsimpy.emt.dc.Inductor
dpsimpy.emt.dc.PiLine
```

`PhaseType::DC` is exposed as:

```python
dpsimpy.PhaseType.DC
```

### Validation example

Added a C++ source-step example containing:

* a 20 kV DC voltage source;
* a feeder resistor;
* a DC Pi-line;
* a resistive load;
* steady-state initialization;
* a voltage step from 20 kV to 22 kV;
* analytical pre-step and post-step operating-point calculations;
* finite-value checks;
* final relative-error reporting;
* CSV logging of node voltages and component interface quantities.

Added Jupyter Notebook with 5 small Test Circuits to validate Python expose


leonardocarreras (commented):
_Superseded, see the review below._

leonardocarreras (commented):
_Superseded, see the review below._

leonardocarreras (commented):
_Superseded, see the review below._

dpsim-models/src/SimNode.cpp
```diff
@@ -80,18 +80,32 @@ VarType SimNode<VarType>::singleVoltage(PhaseType phaseType) {
 
 template <typename VarType>
 UInt SimNode<VarType>::matrixNodeIndex(PhaseType phaseType) {
+  if (mPhaseType == PhaseType::DC) {
```
leonardocarreras, line 85: `initialize()` at line 41 did not get the same branch, so DC is scaled by `RMS3PH_TO_PEAK1PH` there. In the pi-line example the t=0 row logs `v_source_node = 16329.93` (`20000 * sqrt(2/3)`) alongside `v_source_intf = 20000`, correct from t=Δt on.

Only the initial sample, so the solution is unaffected. A DC case at line 41 would make that row consistent.

leonardocarreras (commented):
The SSN math checks out: capacitor `W = Δt/2C`, inductor `W = Δt/2L`, stamps and history signs match the Ph3 bases. I built and ran `EMT_DC_PiLine_SourceStep` and the steady state is within 1e-9 relative of the analytical operating point. Notes inline.

Separately, the `EMT::DC` vs `EMT::Ph1` rationale currently only exists in the source. A short page next to `ssn-components.md` would cover it, as a follow-up if easier.

The automated review above is mostly noise. ~40 findings are the same `DOUBLE_EPSILON` swap, which is wrong here: it is `1E-12` against capacitance in farads, so it would reject a 0.1 pF shunt. The three "critical" ones did not hold up when I checked them.

dpsim-models/src/EMT/EMT_DC_CurrentSource.cpp
```diff
@@ -0,0 +1,119 @@
...
+                                               Attribute<Matrix>::Ptr) {
+  if (!mParametersSet)
+    throw std::logic_error(
+        "setParameters() must be called before source initialization.");
+  validateDCTerminals();
+  updateMatrixNodeIndices();
+  (**mIntfCurrent)(0, 0) = **mCurrentRef;
+}
+
+void EMT::DC::CurrentSource::mnaCompApplyRightSideVectorStamp(
```
leonardocarreras, line 66: This stamps `+i`/`-i`; `EMT_Ph1_CurrentSource.cpp:45` stamps `-i`/`+i`. Hence the notebook wiring this one `connect([node, gnd])` while everything else uses `connect([gnd, node])`.

Either convention works, but it is awkward to change after `dpsimpy.emt.dc.CurrentSource` ships, so worth picking now: match Ph1, or note the difference in the class doc.

dpsim-models/src/TopologicalNode.cpp
```diff
@@ -19,12 +19,12 @@ void TopologicalNode::setInitialVoltage(MatrixComp voltage) const {
 }
 
 void TopologicalNode::setInitialVoltage(Complex voltage) const {
-  if (mPhaseType == PhaseType::Single) {
-    (**mInitialVoltage)(0, 0) = voltage;
-  } else {
+  if (mPhaseType == PhaseType::ABC) {
```
leonardocarreras, line 22: The shift now applies only to `ABC`, previously to everything except `Single`. Nothing in-tree creates A/B/C-typed nodes, so no impact that I can find. Worth a line in the description.

dpsim-models/src/SimPowerComp.cpp
```diff
@@ -197,13 +197,34 @@ void SimPowerComp<VarType>::setTerminalAt(
...
 void SimPowerComp<VarType>::updateMatrixNodeIndices() {
-  for (UInt nodeIdx = 0; nodeIdx < mNumTerminals; nodeIdx++) {
-    mMatrixNodeIndices[3 * nodeIdx] =
-        node(nodeIdx)->matrixNodeIndex(PhaseType::A);
-    mMatrixNodeIndices[3 * nodeIdx + 1] =
-        node(nodeIdx)->matrixNodeIndex(PhaseType::B);
-    mMatrixNodeIndices[3 * nodeIdx + 2] =
-        node(nodeIdx)->matrixNodeIndex(PhaseType::C);
+  for (UInt nodeIdx = 0; nodeIdx < mNumTerminals; ++nodeIdx) {
+    const auto indices = node(nodeIdx)->matrixNodeIndices();
```
leonardocarreras: `SimNode::matrixNodeIndex()` now returns slot 0 for `DC|Single|A` and 0 otherwise, which is what this writes, so the original caller already handles DC.

The change moves a phase-B index from slot 1 to slot 0. Neither version works for B/C nodes anyway, since the solver only assigns `[0]` for non-ABC. Reverting would keep the diff to `SimNode`.

dpsim/examples/cxx/CMakeLists.txt
```diff
@@ -112,6 +112,9 @@ set(CIRCUIT_SOURCES
 	Circuits/DP_SynGenTrStab_3Bus_SteadyState.cpp
 	Circuits/DP_SynGenTrStab_3Bus_Fault.cpp
 	Circuits/SP_SynGenTrStab_3Bus_Fault.cpp
+
+	# DC
+	Circuits/EMT_DC_PiLine_SourceStep.cpp
```
leonardocarreras, line 117: Examples in `CIRCUIT_SOURCES` are compiled but not run in CI, so the printed relative errors will not fail on drift. I ran it by hand and the numbers are good, and the notebook covers the validation.

leonardocarreras (approved):
Thanks for addressing the comments, looks good.

Note for ourselves: we may want to make uniform the sign convention in the domains, but that is matter for another PR.

Issue #300, opened 2024-06-06
Log at debug level


georgii-tishenin:
**Issue Summary**

It is somewhat challenging to configure logging at the debug level. Additionally, there seems to be a mismatch between the intended and actual use of the `Logger` class.

**Example Scenario**

I am working with the [DP_VS_RL1](https://github.com/sogno-platform/dpsim/blob/master/dpsim/examples/cxx/Circuits/DP_VS_RL1.cpp) example and I want to see the `SPDLOG_LOGGER_DEBUG` messages in the log files of components and the solver. I have implemented this in the [log-at-debug-level](https://github.com/sogno-platform/dpsim/tree/log-at-debug-level) branch.

**Steps Taken:**

1. **Set debug logging level for components and simulation objects in the DP_VS_RL1 example.**
   - _This step is intuitive._
2. **Set `SPDLOG_ACTIVE_LEVEL` to `DEBUG` in `Logger.h`.**
   - _This step is less intuitive. But I do understand this is necessary to compile out debug logging statements for performance reasons._
3. **Set log level for `mSLog` in `TopologicalPowerComp` and `Solver`.**
   - _This step is not intuitive to me and suggests a mismatch between the intended and actual use of the `Logger` class by `TopologicalPowerComp`, `Solver`, and other classes._

**Identified Mismatch**

- The `Logger` class has a static `Logger::setLogLevel` function to set the log level for a `spdlog::logger` object, but this function is never used (at least, I couldn't find any references).
- `TopologicalPowerComp` and `Solver` call the static `Logger::get` function to create/get the `spdlog::logger` object. However, the log level of the `spdlog::logger` object is not set within the scope of the `Logger::get` function, resulting in a default log level of `info`.

**Suggestions for Improvement**

1. **Eliminate the need for Step 3.**
   - Either adjust the `Logger` class or it's usage, so that the log level is set appropriately when creating or retrieving a `spdlog::logger` object.
   
2. **Address the less intuitive aspects of Step 2:**
   - Document the need to set `SPDLOG_ACTIVE_LEVEL` to `DEBUG`.
   - Automatically set `SPDLOG_ACTIVE_LEVEL` to `DEBUG` for Debug builds.

**Discussion**
What do you think? I look forward to your feedback and discussion on these points.

leonardocarreras:
There was an older effort in #236 to improve the logging and have some potential solutions that are relevant to here, and that is also the type of change that somehow solves #239 if we have more than one package

PR #631, opened 2026-08-03
docs: document the log level guideline and installing from master

### LINKED ISSUE #88: Guideline on logging level choice
`off` - recommend to use e.g. when running case studies where exclusively the simulation results should be logged
`info` - logging of information that are basic and that are logged only once before/after the simulation
`debug` - ...
`trace` - logging of information in each simulation step

leonardocarreras:
Adds the log level table from #88 to the coding conventions, together with the reason raising a component's level is not enough on its own: SPDLOG_ACTIVE_LEVEL is fixed at INFO in Logger.h, so the debug and trace macros are compiled out entirely, which is the surprise reported in #300.

Also documents installing from master with pip install git+https://github.com/sogno-platform/dpsim, which is what #501 asked for while there is still no Windows or macOS wheel. Built locally with the pinned Hugo 0.108, 133 pages and no errors. Closes #88 and #501.

### DIFF
--- a/docs/hugo/content/en/docs/Developer Guide/Architecture and Conventions/conventions.md
+++ b/docs/hugo/content/en/docs/Developer Guide/Architecture and Conventions/conventions.md
@@ -29,3 +29,29 @@ Current quantities are expressed either as `RMS` or as `PEAK` values:
 Debug or trace should be the default log level for information that might be nice to have but not necessary for every simulation case.
 
 Calls to the logger that might occur during simulation must use spdlog macros, like `SPDLOG_LOGGER_INFO`.
+
+### Which level to write at
+
+The level is chosen by what the message is for, not by how important it feels while writing it.
+
+| Level | What belongs at it |
+| --- | --- |
+| `off` | Nothing. Selected when running a case study where only the simulation results matter |
+| `error` | All errors |
+| `warn` | All warnings |
+| `info` | Basic information logged once, before or after the simulation |
+| `debug` | Information for debugging: extended static information such as initialisation values, matrix stamps, subcomponents |
+| `trace` | Information produced in each simulation step |
+
+The dividing line that matters is between `debug` and `trace`. Anything written once per simulation
+can be `debug`; anything written once per step has to be `trace`, because the cost is multiplied by
+the step count and a run of 100 000 steps turns a single stray line into 100 000.
+
+### Debug and trace are compiled out
+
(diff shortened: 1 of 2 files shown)


docs/hugo/content/en/docs/Developer Guide/Architecture and Conventions/conventions.md
```diff
@@ -29,3 +29,29 @@ Current quantities are expressed either as `RMS` or as `PEAK` values:
...
+can be `debug`; anything written once per step has to be `trace`, because the cost is multiplied by
+the step count and a run of 100 000 steps turns a single stray line into 100 000.
+
+### Debug and trace are compiled out
+
+`SPDLOG_ACTIVE_LEVEL` is fixed at `SPDLOG_LEVEL_INFO` in `dpsim-models/include/dpsim-models/Logger.h`,
+so `SPDLOG_LOGGER_DEBUG` and `SPDLOG_LOGGER_TRACE` calls are removed by the preprocessor and produce
+no output whatever level the logger is set to at runtime. Seeing them requires changing that line and
+rebuilding, which is deliberate: it keeps per-step logging out of the release binary rather than
+merely disabled. Raising the level on a component alone is therefore not enough, and that surprise is
```
pipeacosta, line 56: Wouldn't "Setting the level ..." sound better here? In addition, could you please consider using an alternative word instead of "surprise"?

docs/hugo/content/en/docs/User Guide/install.md
```diff
@@ -50,6 +50,28 @@ conda activate dpsim
...
+the same toolchain a source build does: CMake, a C++17 compiler, and on Windows Visual Studio with
+the C++ desktop development workload. The
+[build page]({{< ref "/docs/Developer Guide/Architecture and Conventions/build.md" >}}) lists them
+per platform, and the CMake options described there are not reachable this way, so a build that
+needs VILLAS or a solver backend turned on has to be configured directly.
+
+Append `@` and a branch or tag to pin what gets built, for example
+`pip install git+https://github.com/sogno-platform/dpsim@v1.2.1`.
+
+To work on DPsim itself, build with CMake instead. An editable install is not a substitute: it links
```
pipeacosta, line 72: What exactly does "To work on DPsim itself" mean? Does it mean extending the DPsim core? Could you please consider writing this more clearly?

pipeacosta (changes requested):
Hi @leonardocarreras. Thanks for this PR. The enhancements to the documentation are very useful. I added two minor comments.

PR #633, opened 2026-08-03
Build and publish Windows wheels

### LINKED ISSUE #368: Start building wheels for Windows

leonardocarreras:
cibuildwheel was pinned to cp3{9,10,11,12,13}-manylinux_x86_64 in a single table, so on a Windows runner it selected nothing at all, which is why #392 needed a second workflow with a competing publish job. Split into per-platform tables and one job builds both.

The matrix gains an os dimension mapped to a platform tag, and the existing publish jobs pick the new wheels up through their dist-* pattern, so there is still one publish path. Also adds a test-command importing dpsimpy and dpsim, which Linux never had.

Linux is unchanged: `manylinux-x86-64-image` only moved into `[tool.cibuildwheel.linux]`, the log still shows `sogno/dpsim:manylinux` being pulled, and the artifact is the same 57.5 MB as master. The Windows wheel was run by hand rather than trusted from a green tick: same EMT circuit, same KLUAdapter, 1002 rows, identical final values as the Linux wheel.

Three things are absent on Windows, from diffing the type stubs in both wheels: `dpsimpyvillas` (VILLASnode has no Windows port), `RealTimeSimulation` and `RealTimeDataLogger` (`WITH_RT` is gated on `Linux_FOUND`), and `SynchronGeneratorDQODE` (needs Sundials). KLU and the CGMES reader are present. The install page now says so.

Supersedes #392, closes #368.

### DIFF
--- a/.github/workflows/publish_to_pypi.yaml
+++ b/.github/workflows/publish_to_pypi.yaml
@@ -40,27 +40,37 @@ jobs:
 
   build-wheels:
     name: Build DPsim Python wheel
-    runs-on: ubuntu-24.04
+    runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
-        platform: [manylinux_x86_64]
+        os: [ubuntu-24.04, windows-latest]
         python: [cp39, cp310, cp311, cp312, cp313]
+        include:
+        - os: ubuntu-24.04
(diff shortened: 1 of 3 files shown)


pipeacosta (approved):
Thanks for the PR @leonardocarreras. Looks good from my side.

PR #634, opened 2026-08-03
removeComponent: drop the component from mComponentsAtNode too

### LINKED ISSUE #625: removeComponent leaves stale entries in mComponentsAtNode
Follow-up to #529.

`removeComponent` erases from `mComponents` only, so the component stays in `mComponentsAtNode`. This is not cosmetic: `PFSolver` and `PFSolverPowerPolar` iterate that map when classifying buses and collecting powers, so a removed component can still be picked up during power flow setup.

\#621 fixes the same leftover for the node-removal path. This is the component-removal path.

## Done when

- [ ] `removeComponent` drops the component from `mComponentsAtNode` as well
- [ ] a regression check covers it

@barismuutlu offered this as a follow-up in the description of #621. If he wants to take, is his.

barismuutlu:
Closes #625. Follow-up to #621, as offered in that PR's description.

`removeComponent` erased the component from `mComponents` only, so it stayed in `mComponentsAtNode`. Since `PFSolver`/`PFSolverPowerPolar` iterate that map when classifying buses and collecting powers, a removed component could still leak into power flow setup. This is the component-removal counterpart of #621 (node-removal path).

## Changes

- `SystemTopology::removeComponent` now also purges the removed component from every per-node list in `mComponentsAtNode` (same-named duplicates included).
- New regression notebook `examples/Notebooks/Features/SystemTopologyRemoveComponent.ipynb` covering the "Done when" items of #625:
  - the removed component disappears from `components_at_node`,
  - removing a non-existent name is a no-op,
  - all same-named components are purged from the map.

## Testing

- Regression notebook executed locally against the built `dpsimpy` — all assertions pass.
- Extra local edge checks: surviving components keep their exact per-node entries after a removal; removing every component leaves the map empty; an EMT simulation still runs on a topology modified via `remove_component`.
- Repo pre-commit hooks pass on both files (clang-format, black-jupyter, editorconfig, cleared outputs).

Note (out of scope): `addComponent` never inserts into `mComponentsAtNode` either — the map is only populated by the `(frequency, nodes, components)` constructors and `connectComponentToNodes`. Happy to look at that separately if it is considered a gap.

### DIFF
--- a/dpsim-models/src/SystemTopology.cpp
+++ b/dpsim-models/src/SystemTopology.cpp
@@ -246,6 +246,11 @@ void SystemTopology::reset() {
 void SystemTopology::removeComponent(const String &name) {
   for (auto it = mComponents.begin(); it != mComponents.end();) {
     if ((*it)->name() == name) {
+      // Drop the component from the per-node lists as well, otherwise
+      // the power flow solvers can still pick it up via mComponentsAtNode
+      for (auto &[topoNode, comps] : mComponentsAtNode) {
+        comps.erase(std::remove(comps.begin(), comps.end(), *it), comps.end());
+      }
       it = mComponents.erase(
           it); // safe: returns next valid iterator when erasing
     } else {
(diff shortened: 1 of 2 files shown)


leonardocarreras (approved):
Looks good.

PR #638, opened 2026-08-04
SystemTopology: derive mComponentsAtNode instead of hand-maintaining it

### LINKED ISSUE #626: SystemTopology constructor segfaults on components with unconnected terminals
`SystemTopology(frequency, nodes, components)` segfaults if any component in the list was never connected.

The three- and four-argument constructors call `componentsAtNodeList()`, which calls `topologicalNodes()`. `SimPowerComp` sizes `mTerminals` with `resize(num, nullptr)`, so an unconnected component holds null terminals and they get dereferenced. Adding the same component with `addComponent` is fine, since that path does not touch the map, so the crash depends only on which constructor is used.

Reproduced on `acaacf7da` with nothing applied on top.

## Done when

- [ ] an unconnected component no longer crashes the constructor, either by skipping null terminals or by failing with a descriptive error
- [ ] a regression check covers it

### LINKED ISSUE #635: Make mComponentsAtNode a derived view instead of hand-maintained state
Follow-up to #621, #634 and #626, and to the gap @barismuutlu pointed out at the end of #634.

Right now `mComponentsAtNode` is state we maintain by hand. Three places write to it: `componentsAtNodeList()` (only from the 3- and 4-arg constructors), `connectComponentToNodes()`, and `CIM::Reader`. Since #621 and #634 there are two places that remove from it. `addComponent` never inserts at all, so what we actually have is a snapshot taken at construction time that quietly drifts away from `mComponents`.

Making `addComponent` insert would not fix it. When `addComponent` runs, the component is usually not connected yet, because `SimPowerComp` sizes `mTerminals` with `resize(num, nullptr)` and `connect()` comes later. Inserting there would dereference those null terminals, which is the crash in #626. It would also give us a third writer to keep in sync, and it can insert twice if a component goes through the constructor and then through `connectComponentToNodes`.

Deriving the map instead gets rid of the whole class of bug. We can already recover node membership from `topologicalNodes()`, and every consumer sits in `PFSolver` and `PFSolverPowerPolar`, so rebuilding at power flow setup is enough. No simulation path reads the map.

## Done when

- [ ] `componentsAtNodeList()` is idempotent, clearing and rebuilding from `topologicalNodes()`
- [ ] components with null terminals are skipped, or rejected with a descriptive error, which closes #626
- [ ] the map is rebuilt where it is consumed, so the hand-maintenance in `addComponent`, `removeComponent` and `removeNode` is no longer needed
- [ ] the regression checks from #621, #634 and #626 still pass

@barismuutlu you offered to look at the `addComponent` side in #634. This is that gap, just framed a bit wider. Yours if you want it.

barismuutlu:
Closes #635. Closes #626.

`mComponentsAtNode` was a snapshot taken by the 3- and 4-arg constructors that quietly drifted away from `mComponents`: `addComponent` never inserted, the removal paths had to purge it by hand, and repeated calls of `componentsAtNodeList()` duplicated entries. The map is now a view derived from `mComponents`, rebuilt where it is consumed.

## Changes

- `componentsAtNodeList()` is idempotent: it clears the map and rebuilds it from `topologicalNodes()`.
- `SimPowerComp::topologicalNodes()` skips unconnected (null) terminals, so a component that is not fully wired no longer segfaults the `SystemTopology(frequency, nodes, components)` constructors — that closes #626. `componentsAtNodeList()` is its only caller, so no other behaviour changes.
- `PFSolver::initialize()` rebuilds the map before the solver classifies buses, so `PFSolver`/`PFSolverPowerPolar` never see entries for removed components or miss components added after construction.
- The Python binding exposes `components_at_node` as a read-only property that rebuilds before every read (nothing in the repo assigned to it).
- The manual purges in `removeComponent` (from #634) and in `removeNode`/`removeComponentsConnectedTo` (from #621) are gone; the derived view keeps the same visible behaviour, which both regression notebooks still verify.
- New regression notebook `examples/Notebooks/Features/SystemTopologyComponentsAtNode.ipynb` covering: the #626 constructor repro, a component connected after construction appearing in the map, no duplicates on repeated reads, and removal staying visible through the derived view.

`CIM::Reader` and `connectComponentToNodes` still write the map so that direct C++ reads (e.g. the CIGRE load-profile example) keep working without a rebuild call; the rebuild simply recreates the same content wherever it runs.

## Testing

- New regression notebook executed locally against the built `dpsimpy` — all assertions pass, including the #626 repro that previously crashed the interpreter.
- The regression notebooks from #621 (`SystemTopologyRemoveNode.ipynb`), #634 (`SystemTopologyRemoveComponent.ipynb`) and #622 (`SystemTopologyAddAlias.ipynb`) all still pass unchanged, plus local edge suites for the removal paths.
- Master comparison: `EMT_Ph3_VSIVoltageControlVCO_PF_Init.ipynb` (power flow init + EMT run) executed on master and on this branch — both CSV outputs are bit-identical, so power flow results are unaffected for well-formed topologies.
- Repo pre-commit hooks pass on all changed files.

### DIFF
--- a/dpsim-models/src/SimPowerComp.cpp
+++ b/dpsim-models/src/SimPowerComp.cpp
@@ -218,6 +218,10 @@ template <typename VarType>
 TopologicalNode::List SimPowerComp<VarType>::topologicalNodes() {
   TopologicalNode::List nodes;
   for (typename SimTerminal<VarType>::Ptr term : mTerminals) {
+    // mTerminals is preallocated with nullptr and only filled by connect(),
+    // so an unconnected component must not crash the lookup (see #626)
+    if (!term)
+      continue;
     nodes.push_back(term->node());
   }
   return nodes;
--- a/dpsim-models/src/SystemTopology.cpp
+++ b/dpsim-models/src/SystemTopology.cpp
@@ -69,6 +69,9 @@ void SystemTopology::connectComponentToNodes(
 }
 
 void SystemTopology::componentsAtNodeList() {
+  // Idempotent: the map is a view derived from mComponents, so rebuild it
+  // from scratch instead of accumulating entries across calls (see #635)
+  mComponentsAtNode.clear();
   for (auto comp : mComponents) {
     auto powerComp = std::dynamic_pointer_cast<TopologicalPowerComp>(comp);
     if (powerComp)
@@ -246,11 +249,8 @@ void SystemTopology::reset() {
 void SystemTopology::removeComponent(const String &name) {
   for (auto it = mComponents.begin(); it != mComponents.end();) {
     if ((*it)->name() == name) {
-      // Drop the component from the per-node lists as well, otherwise
-      // the power flow solvers can still pick it up via mComponentsAtNode
-      for (auto &[topoNode, comps] : mComponentsAtNode) {
-        comps.erase(std::remove(comps.begin(), comps.end(), *it), comps.end());
-      }
+      // mComponentsAtNode needs no manual cleanup here: it is rebuilt from
+      // mComponents by componentsAtNodeList() wherever it is consumed (#635)
       it = mComponents.erase(
           it); // safe: returns next valid iterator when erasing
     } else {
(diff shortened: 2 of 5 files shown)


leonardocarreras (approved):
@barismuutlu thanks for the PR, looks good

If you feel like doing a small improvement, logging the name of the component that is ignored because terminal is null might be informative (as maybe there is actually an error in the circuit data)

PR #644, opened 2026-08-06
fix: re-pivot fully when no variable matrix entries are declared

FGH-wirtz:
partialRefactorize() calls klu_partial_factorization_path(), which needs a factorization path. That path only exists if a component declared variable system-matrix entries - factorize() skips computing it otherwise. With an empty list KLU finds Numeric->path == NULL, sets KLU_PATH_INVALID and returns without touching the factorization. Only KLU_PIVOT_FAULT is checked here. --> recomputed matrix is discarded, every later time step is wrong

Found on a network with switched lines: after a trip, all following results were wrong, with no error or warning.

Only VBR synchronous generators fill mVariableSystemMatrixEntries - so empty list is the common case switches e.g. declare none.
Proposed fix: Fall back to a full, re-pivoted factorization when list is empty. Runs only at switch events, so cost is negligible

### DIFF
--- a/dpsim/src/KLUAdapter.cpp
+++ b/dpsim/src/KLUAdapter.cpp
@@ -137,6 +137,10 @@ void KLUAdapter::partialRefactorize(
   if (systemMatrix.nonZeros() != nnz) {
     preprocessing(systemMatrix, listVariableSystemMatrixEntries);
     factorize(systemMatrix);
+  } else if (mPartialRefactorizationMethod !=
+                 PARTIAL_REFACTORIZATION_METHOD::NO_PARTIAL_REFACTORIZATION &&
+             (mVaryingColumns.empty() || mVaryingRows.empty())) {
+    factorize(systemMatrix);
   } else {
     auto Ap =
         Eigen::internal::convert_index<Int *>(systemMatrix.outerIndexPtr());


dpsim/src/KLUAdapter.cpp
```diff
@@ -134,7 +134,10 @@ void KLUAdapter::refactorize(SparseMatrix &systemMatrix) {
 void KLUAdapter::partialRefactorize(
     SparseMatrix &systemMatrix,
     std::vector<std::pair<UInt, UInt>> &listVariableSystemMatrixEntries) {
-  if (systemMatrix.nonZeros() != nnz) {
+  // Without declared variable entries no factorization path was computed, so
+  // partial refactorization would silently do nothing
+  if (systemMatrix.nonZeros() != nnz ||
+      listVariableSystemMatrixEntries.empty()) {
     preprocessing(systemMatrix, listVariableSystemMatrixEntries);
     factorize(systemMatrix);
```
leonardocarreras, line 139: Thanks for the fix, I think it solves the problem, but it might be replacing for everyone/every solvero.

I have a suggestion, to avoid full refactorization and keep the simulations a bit faster

```suggestion
  if (systemMatrix.nonZeros() != nnz) {
    preprocessing(systemMatrix, listVariableSystemMatrixEntries);
    factorize(systemMatrix);
  } else if (mPartialRefactorizationMethod !=
                 PARTIAL_REFACTORIZATION_METHOD::NO_PARTIAL_REFACTORIZATION &&
             (mVaryingColumns.empty() || mVaryingRows.empty())) {
    factorize(systemMatrix);
```

Eventually the problem also has something to do with the way we are constructing the DiakopticsSolver (we do not have the correct setConfiguration() call and it will be ignored if set) but that can be done in a different PR if needed because is a refactor of other file. The other solvers do, that is why this problem did not show.

leonardocarreras:
Hi @FGH-wirtz! Thanks for the fix, I left a small comment, tell me what do you think, or if you have a minimal config that we could use to reproduce the exact bug you are having (if the suggestion does not work for it).

leonardocarreras:
Note: had to close-open to make the CI work, it got stale for some reason.

dpsim/src/KLUAdapter.cpp
FGH-wirtz: sure, thats actually the better solution as it only applies to the problematic case.

leonardocarreras: @FGH-wirtz feel free to implement/test it and we can merge it

leonardocarreras (approved):
Hi @FGH-wirtz! Thanks for the fix, looks good.

PR #645, opened 2026-08-07
add: run binder container test by running examples/villas/dpsim-file.py

### LINKED ISSUE #432: import dpsimpy fails in Docker/Binder
**Describe the bug**
When using the Python wrapper of DPsim via Binder (e.g. Quickstart Guide.ipynb example), the import of dpsimpy fails. 

**To Reproduce**
1. Go to [DPsim GitHub](https://github.com/sogno-platform/dpsim)
2. Click on **launch Binder** badge in README file.
3. Open **QuickStart Guide.ipynb**
4. Run `import dpsimpy` cell -> produces an error: `ImportError: /home/jovyan/.local/lib/python3.13/site-packages/dpsimpy.cpython-313-x86_64-linux-gnu.so: undefined symbol: villas__node__message__unpack`

**Expected behavior**
no error in step 4.

**Additional context**
This seems to happen only in the Docker environment using the [Dockerfile](https://github.com/sogno-platform/dpsim/blob/master/.binder/Dockerfile). Hence:

- Local (not Docker-based) installation of dpsimpy works without issues.
- CI passes. This may be worth to add a test to catch this problem in the future.

### LINKED ISSUE #435: mybinder is not automatically checked in the CI before a PR is approved
**Is your feature request related to a problem? Please describe.**
There is currently no way to catch errors in Binder before they get to the master branch.

**Describe the solution you'd like**
Somehow this should be automated.

**Describe alternatives you've considered**
A manual checklist that makes you try this on the mybinder website

**Additional context**
This is somehow also associated with #301, we do not have a simple way to compile the docker files in the current setup. We might want to think about it a bit more.

SystemsPurge:
Closes #435 . 
Also encountered the very situation where import dpsimpy ( or dpsimpyvillas ) doesn't work. It was when the container is ran with a user different than the one used to build the image. The python package ends up in ~/.local. So might also close #432

### DIFF
--- a/.github/workflows/container.yaml
+++ b/.github/workflows/container.yaml
@@ -118,10 +118,26 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USER }}
           password: ${{ secrets.DOCKERHUB_SECRET }}
 
-      - name: Build and push
-        id: docker_build_release
+      - name: Build
         uses: docker/build-push-action@v7
         with:
-         file: .binder/Dockerfile
-         push: true
-         tags: sogno/dpsim:binder
+          context: .
+          file: .binder/Dockerfile
+          load: true
+          push: false
+          tags: sogno/dpsim:binder
+
+      - name: Run test
+        run: |
+          mkdir -p logs
+          sudo chown 1000:1000 logs
+          docker run --rm \
+            -u 1000:1000 \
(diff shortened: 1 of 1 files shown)


leonardocarreras (approved):
Lets try it out

Issue #647, opened 2026-08-07
Pin Python dependencies with checked-in lockfiles instead of resolving at build time


leonardocarreras:
Every dev image and dependency script runs `pip-compile` against `pyproject.toml` at build time and deletes the output right after (packaging/Docker/Dockerfile.dev:69-72 and the five sibling Dockerfiles, packaging/Shell/install-*-deps*.sh). Nothing is pinned, so each build resolves the graph from scratch and a new upstream release can turn CI red without a single commit touching the repo. The check-python-deps job in build_test_linux_fedora.yaml re-resolves the extras but throws the result away as well, so it catches a hard resolution conflict and nothing else.

Proposal (input from @SystemsPurge ): compile each extra once into a checked-in lockfile, `uv pip compile pyproject.toml --extra dev -o requirements-dev.txt` and the same for manylinux and test, and have the Dockerfiles and shell scripts install with `-r` instead of compiling. check-python-deps then becomes a freshness gate: recompile and fail if the lockfiles differ, so a dependency bump arrives as a reviewable PR rather than a random red build. uv also makes the resolve step fast enough to keep as a standalone CI step.

PR #648, opened 2026-08-07
fix(ci): drop the build-time Python dependency layer from the manylinux image

leonardocarreras:
pip-compile ran under the image's Python 3.6 and stopped resolving when pybind11-stubgen and numpy entered [build-system].requires. The layer is unnecessary anyway: cibuildwheel, the image's only consumer, installs everything itself, so remove it. Verified locally: cp39 and cp312 wheels build and pass the import test. Related to #647.

### DIFF
--- a/packaging/Docker/Dockerfile.manylinux
+++ b/packaging/Docker/Dockerfile.manylinux
@@ -14,6 +14,8 @@ ARG MAKE_OPTS=-j4
 
 ENV PLAT=manylinux_2_28_x86_64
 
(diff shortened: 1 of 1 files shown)


SystemsPurge (dismissed):
This seems to resolve the issue of the manylinux image build not finding the proper pybind11-stubgen version. Keep in mind however that we might need to consider a uniform mechanism to guarantee python/pip versions accross all CI's , as it seems to be causing many problems lately.

PR #660, opened 2026-08-07
refactor!: remove the CUDA and MAGMA linear solvers

### LINKED ISSUE #164: Linear solvers using CUDA do not work or yield wrong results
**Describe the bug**
When compiling DPsim with CUDA-support (e.g. using the cmake-flag `-DWITH_CUDA=ON`), the linear solvers implemented in files `MNASolverGpuSparse.cpp` and `MNASolverGpuDense.cpp` yield segmentation faults, if an example with a variable system matrix is used, or wrong results - possibly due to the presence of zeros on the system matrix' diagonal. In case of `CUDADense`, some values are `NaN`

**To Reproduce**
1. modify example files of `WSCC_9bus_mult_coupled` and `DP_WSCC9bus_SGReducedOrderVBR` to use `CUDADense` (for example)
2. build
   1. `mkdir build && cd build`
   2. `cmake .. -DWITH_CUDA=ON`
   3. `make WSCC_9bus_mult_coupled DP_WSCC9bus_SGReducedOrderVBR`
3. cd dpsim/examples/cxx
4. run
   1. `./WSCC_9bus_mult_coupled -ocopies=0`
   2. `./DP_WSCC9bus_SGReducedOrderVBR`

**Expected behavior**
No segmentation faults and correct results, if compared to other solvers, such as `EigenSparse`

### LINKED ISSUE #646: GPU - CUDA and DPsim: delete
The solvers are most likely not used, and do not work as of 2022 like mentioned in #164.

If nobody is using, I think we can make the codebase a bit slimmer. If someone is using, it would be great to fix them soon.

leonardocarreras:
Closes #164, closes #646. 

Context: The GPU adapters are broken and no GitHub Actions job builds them.

Removes the three adapters, the MAGMA plugin and the CUDA enum values, and drops CUDA, MAGMA and cricket from the Rocky container. dpsimpy and the C++ examples build clean; the image was not rebuilt.


leonardocarreras:
Hi @n-eiling, hi @fwege, would this set of changes be okay for you? 

We would appreciate your input, or heads-up about your plans or any comment. Of the people I know, you are the most familiar with the GPU work in the software.

Other points: 
- if you are using the GPU solvers and plan to update it, we are happy to keep them and drop this PR.
- we could keep things like dependencies (for example cricket) if they are needed for any other purpose.

I keep it as draft for the moment.

PR #662, opened 2026-08-07
refactor: make the decoupling components CompositePowerComp

### LINKED ISSUE #527: Make the decoupling components `CompositePowerComp`
Follow-up to #423.

Today every decoupling component makes the caller wire its internals into the system by hand. For the ITM:

```cpp
sys.addComponent(idealTrafo);
sys.addComponents(idealTrafo->getComponents());   // manual
sys.addNode(idealTrafo->getVirtualNode());        // manual, easy to forget
```

The existing decoupling line (`DecouplingLine` / `DecouplingLineEMT` / `DecouplingLineEMT_Ph3`) has the same thing with `getLineComponents()`. If you forget one of these lines the system is silently wrong (a node ends up floating). This is what caused the missing-resistor and missing `addNode` issues found in #423.

`CompositePowerComp` is made for this: the component owns its internal parts and virtual nodes and registers them itself, so you just add the one component. This also matches the migration done elsewhere (see #508 and #511).

Applies to both families: the ITM components (`DecouplingIdealTransformer_EMT_Ph1` / `EMT_Ph3` / `SP_Ph1` / `DP_Ph1`) and the line components.

## Done when

- [ ] All decoupling components (ITM and line) are `CompositePowerComp`
- [ ] Examples no longer call `getComponents()` / `getLineComponents()` / `getVirtualNode()` by hand

leonardocarreras:
Every decoupling component made the caller add its internals and virtual node by hand, and forgetting either silently floated a node. All seven are now CompositePowerComp, so one addComponent is enough and the terminals come in through connect().

The seven decoupling examples give the same results as before. setParameters() changes signature, and the ITM's v_intf/i_intf become v_src_intf/i_src_intf since SimPowerComp was shadowing them. Closes #527.


leonardocarreras:
Reopening in-repo so the namespace follow-up can stack on it.

PR #670, opened 2026-08-07
fix: resolve Python dependencies with uv in the remaining images and scripts

leonardocarreras:
pip-tools 7.6.0 is broken on every Python (undeclared typing_extensions below 3.11, removed pip internals from 26.2 up). The rocky and manylinux images already tripped and were fixed; these six spots carry the same pattern and fail the same way on their next base refresh, dev-debian (Python 3.9) already would. Convert Dockerfile.dev, dev-minimal, dev-debian, Dockerfile.rpm and both install-fedora-deps scripts to uv, as #648/#651 did. Verified the converted stage on fedora:42 and debian:11, dev extra installs and imports. Related to #647.


SystemsPurge (approved):
Good, so long as we consider version pinning and/or lockfiles in the future

PR #676, opened 2026-08-10
ci(llm-review): stop repeating findings on every push

leonardocarreras:
Unconfirmed and sub-50 findings go to the log, not the PR. A run with nothing new edits its previous review instead of posting again, which sends no mail. On PR #581 history: 12 findings instead of 171, 4 mails instead of 10.


georgii-tishenin (approved):
@leonardocarreras, yes LLM review seemed to noisy. Let's try this out. Thanks for working further on this!

PR #678, opened 2026-08-10
Fix the real Sonar findings and scope out the pybind false positive

leonardocarreras:
Clears the Sonar bug findings worth clearing before the 1.3.0 baseline resets. TurbineGovernor left twelve parameters uninitialized, AttributeUpdateTask called std::forward on a by-value pack where it means std::move, and Dockerfile.shmem put spaces around the LABEL assignments, which collapses all four labels into a single one carrying the rest of the block as its value.

The other 25 findings are all cpp:S5184 on pybind11's py::class_<T>(mod, "T") registration idiom, which holds no resource, so the rule is scoped out for dpsim/src/pybind rather than renaming 25 temporaries. The notebook float-equality reports and the CSVReader float loop counter are left alone, both need a closer look than a sweep.


georgii-tishenin (approved):
The changes are small, okay to merge from my point of view.

PR #681, opened 2026-08-11
build(python): drop 3.9 and build wheels for 3.14

leonardocarreras:
Drops CPython 3.9 and adds 3.14 in pyproject, both cibuildwheel matrices and the CI checks. The published manylinux image already carries cp314, so no image rebuild is needed.

The install page gains a version matrix and a table of what the Windows wheel lacks; README and the build page no longer claim wheels are Linux only.

Succesful runs:
https://github.com/leonardocarreras/dpsim/actions/runs/31490363600/job/93785444971
https://github.com/leonardocarreras/dpsim/actions/runs/31490363600/job/93785444971

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -131,7 +131,7 @@ jobs:
       - name: Install uv
         uses: astral-sh/setup-uv@v9.0.0
         with:
-          python-version: '3.9'
+          python-version: '3.10'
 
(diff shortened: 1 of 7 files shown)


leonardocarreras:
Tested here, can only be green after merging:
https://github.com/leonardocarreras/dpsim/actions/runs/31490363600

SystemsPurge (approved):

PR #683, opened 2026-08-12
Feature: Added EMT classes GFL, refactored SSN_GFL and SSN_GFL_Split 

matthiasmees:
## Summary

This PR extends the EMT grid-following converter implementations and adds a common benchmark for comparing the different modeling approaches.

The main goal is to provide equivalent GFL implementations with different numerical formulations so that their dynamic behavior and computational performance can be compared under identical operating conditions.

The implementations covered by this PR are:

- `EMT::Ph3::GFL`
- `EMT::Ph3::SSN_GFL`
- `EMT::Ph3::SSN_GFL_Split`

They are benchmarked against the existing:

- `EMT::Ph3::AvVoltSourceInverterStateSpace`

### Composite EMT GFL model

Adds / extends `EMT::Ph3::GFL` as a physical composite EMT, it is exactly the same class as EMT::Ph3::AvVoltageSourceInverterDQ but without hte optional transformer and connection resistor R_c which was the reason for many failed power flow runs.
Also the wrongly measured voltage point was fixed.

### SSN_GFL model

This model of the SSN_GFL is the refactored version of Georgiis AvVoltagseSourceInverterStateSpace with numerical derivations now for speed increment, it works exactly the same like other class but the Jacobian derivations are done now with equations

### SSN_GFL_Split

This model represents the same model as SSN_GFL but with two seperate state-spaces split up: COnstant Network State-space and Variable Control state-space. this way speed can be improved by around 45 % because of the lower linearization effort

Pybind and comparison example cxx included


dpsim-models/src/EMT/EMT_Ph3_GFL.cpp
georgii-tishenin, line 1: Need to add SPDX header for this class.

examples/Notebooks/Components/EMT_Ph3_GFL_Comparison.ipynb
```diff
@@ -0,0 +1,1309 @@
...
+    "    \"SSN_GFL\",\n",
+    "    \"SSN_GFL_Split\",\n",
+    "    \"AvVoltSourceInverterStateSpace\",\n",
+    "]\n",
+    "\n",
+    "for cls_name in required_classes:\n",
+    "    cls = getattr(dpsimpy.emt.ph3, cls_name)\n",
+    "    print(f\"{cls_name:34s}: {cls}\")\n",
+    "\n",
+    "REPO_ROOT = Path(\"/home/mms/dpsim\")\n",
```
georgii-tishenin: hm, I think hard-coded repo root can fail on other machines. In the other notebooks the usual way is to use relative \logs path.

georgii-tishenin (commented):
Hi @matthiasmees, thanks for the PR. Looks good.
From my point of view there are two minor issues worth fixing before merging (see my comments).

leonardocarreras (commented):
Hi Matthias! Thanks for the PR, just some very general comments
- You maybe want to use the Math::parkTransformPowerInvariant, however there are many other models with duplication in the same type
- The examples look a bit similar, you may want to think of a function/template/factory, not critical but may help with duplication and testing
- It looks like the measuring point and the operating points might not be exactly the same for all cases, just to keep in account when comparing
- You may want to test with logging off for the measurements (the time difference of only the calculation is even better), the time of logging dilutes a bit the total contribution of your change
- What do you say if we trim a bit the comments in the files? I think some of the stuff that is long can go to documentation, and in other cases a short explanation is fine because the code is not so clear

Unrelated, for the conventional commits, you may want to split the things like:
- feat(GFL): add new GFL model
- refactor(GFL_SSN): change blah

For the test, include it in TEST_SOURCES for the notebook to run it (and as Georgii said, better not use hardcoded paths)

georgii-tishenin (approved):
LGTM. The concrete review comments have been addressed. In my opinion remaining points are follow-up cleanup/design decisions rather than blockers.

PR #684, opened 2026-08-13
EMT_Ph3_SSN_InductionMotor

MarvinTollnitschRWTH:
Implements an EMT_Ph3_SSN_InductionMotor model for use in the MNASolver/MNASolverDirect. Theory is based on Section 3. B. of https://ieeexplore.ieee.org/document/8536236.

As of now, a cold start load switch example exists as a first test. In the future, powerflow simulation by use of the power flow solver and powerflow initialization of the component are further implementation goals.


leonardocarreras (approved):
Hi @MarvinTollnitschRWTH thanks for the PR!

I think we can see better when using the component if there is something missing or how to improve, lets try it.

PR #687, opened 2026-08-16
Add current-zero and exponential switching models for EMT and DP three-phase switches

matthiasmees:
## Description

This PR extends the three-phase switch models in the EMT and dynamic phasor
(DP) domains with more realistic opening behavior.

The existing binary switch behavior is retained as the default. Two additional
switching modes are introduced:

- `CurrentZero`
- `ExponentialZCSEmulation` (Approach from Wirtz: https://ieeexplore.ieee.org/abstract/document/10408527) 

The implementation supports independent phase opening for current-zero
switching and variable switch conductances during the exponential switching
process.

In addition, this PR fixes the initialization of three-phase DP passive
components used by the benchmark example, so that the DP simulation starts from
a consistent power-flow operating point.

## Motivation

The existing switch model changes directly between the closed and open
resistance states. For breaker opening, this can interrupt a non-zero current
instantaneously and produce unrealistic transients.

In EMT simulations, a more realistic approximation is to keep each breaker pole
closed until the corresponding phase current reaches its next natural zero.

For dynamic phasor simulations, the physical fundamental-frequency zero crossing
is not directly visible in the complex envelope. This PR therefore adds two
different approaches:

1. reconstruction of the physical instantaneous phase current and detection of
   its zero crossing;
2. an exponential variable-resistance transition that emulates zero-current
   switching behavior without explicit zero-crossing detection.

## Switching modes

The following modes are available for both `EMT::Ph3::Switch` and
`DP::Ph3::Switch`:

```cpp
enum class SwitchingMode {
  Ideal = 0,
  CurrentZero = 1,
  ExponentialZCSEmulation = 2,
};

### DIFF
--- a/dpsim-models/include/dpsim-models/Base/Base_Ph3_Switch.h
+++ b/dpsim-models/include/dpsim-models/Base/Base_Ph3_Switch.h
@@ -10,10 +10,12 @@
 
 #include <dpsim-models/AttributeList.h>
 #include <dpsim-models/Definitions.h>
+
 namespace CPS {
 namespace Base {
 namespace Ph3 {
-/// Dynamic Phasor Three-Phase Switch
+
+/// Common base for three-phase switches.
 class Switch {
 protected:
   Bool mIsClosedPrev = false;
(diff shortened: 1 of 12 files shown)


leonardocarreras:
Thanks Matthias! I will have a look soon, meanwhile we can also invite @FGH-wirtz to comment this PR if he wishes, maybe we can see if he has suggestions.

FGH-wirtz:
First of all: I think this is a great contribution. three-phase support and EMT variant are missing today, nice

Independently of the switch models, the DP::Ph3::Inductor and DP::Ph3::Resistor power-flow initialization fixes (missing RMS3PH_TO_PEAK1PH, and the resistor building phases B/C from an unassigned mIntfVoltage) are valid bug fixes

On the exponential mode: why not extend varResSwitch instead? DP::Ph1::varResSwitch already implements a geometric resistance ramp, just single-phase, DP/SP only, with an unparameterized transition duration and mutation of R_open/R_closed. I think the approach here is sound and the explicit switching time is an improvement - but then the follow-up question is: should varResSwitch be deprecated and everything based on the new implementations? @leonardocarreras 

DP CurrentZero mode: detecting the physical zero doesnt remove the transient. So we have the correct point in time, but still the problem that the trapezoidal companion model cannot represent an envelope discontinuity, so we have the transient. In my work i thought about zeroing the history terms to get EMT like continous behaviour here, but as switching would need topology information i decided to skip this approach and went with the Exponential approximation
So if u have any idea on how this could be done differently, that would be great - otherwise we should think about naming it differently then "CurrentZero" mode in DP - because at least my expectation would be a "EMT like behaviour" based on the naming

georgii-tishenin (approved):
Looks good to me.

I agree with @FGH-wirtz, we could in the future deprecate `DP::Ph1::varResSwitch`.
I think the way @matthiasmees did it is very usable, because now you have different options you can select with a single component. Yes, Current Zero does not fix the numerical oscillations (especially for DP), but this is an option consistent in switching times with EMT.
I think this is very nice step forward, I would merge it and then discuss further what can be improved.

PR #690, opened 2026-08-17
Support state space extraction for EMT split SSN components

georgii-tishenin:
## Description

This PR adds state-space extraction support for split SSN components whose electrical plant and controller are evaluated separately with a one-step interface delay.

It introduces reusable split V-type SSN base classes and refactors `EMT::Ph3::SSN_GFL_Split` to use them. The extractor now maintains constant and updated contribution matrices, allowing both variable components and split SSN components to refresh their contributions when required.

A logarithmic discrete-to-continuous pole mapping is added for models containing explicit discrete delays.

The included C++ example compares the split and no-delay GFL models with analytical references for time steps from 1 µs to 1 ms.

### DIFF
--- a/dpsim-models/src/EMT/EMT_VTypeSplitSSNComp.cpp
+++ b/dpsim-models/src/EMT/EMT_VTypeSplitSSNComp.cpp
@@ -1,0 +1,250 @@
+// SPDX-FileCopyrightText: 2026 Institute for Automation of Complex Power Systems, EONERC, RWTH Aachen University
+// SPDX-License-Identifier: MPL-2.0
+
+#include <dpsim-models/EMT/EMT_VTypeSplitSSNComp.h>
+
+#include <stdexcept>
+
+using namespace CPS;
+
+EMT::VTypeSplitSSNComp::VTypeSplitSSNComp(String uid, String name,
+                                          Int inputSize, Int outputSize,
+                                          Int controllerStateSize,
+                                          Int controllerInputSize,
+                                          Int controllerOutputSize,
+                                          Logger::Level logLevel)
+    : VTypeSSNComp(uid, name, inputSize, outputSize, logLevel),
+      mControllerStateSize(controllerStateSize),
+      mControllerInputSize(controllerInputSize),
+      mControllerOutputSize(controllerOutputSize),
+      mMeasurementState(Matrix::Zero(controllerInputSize, 0)),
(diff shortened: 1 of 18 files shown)


matthiasmees (approved):

dpsim-models/src/EMT/EMT_VTypeSplitSSNComp.cpp
```diff
@@ -0,0 +1,250 @@
...
+void EMT::VTypeSplitSSNComp::recomputeDiscreteModel() {
+  SSNComp::recomputeDiscreteModel();
+
+  const Matrix identity = Matrix::Identity(mA.rows(), mA.cols());
+  const Matrix lhs = identity - 0.5 * mTimeStep * mA;
+  mdBControllerOutput =
+      lhs.fullPivLu().solve(0.5 * mTimeStep * mBControllerOutput);
+
+  recomputeControllerDiscreteModel();
+  rebuildSplitDiscreteModel(mControllerBd);
```
leonardocarreras, line 181: Do we need to do this every time, or only for the extraction?

dpsim/examples/cxx/StateSpace/EMT_Ph3_GFL_Split_StateSpaceExtraction.cpp
```diff
@@ -0,0 +1,529 @@
...
+#include <cmath>
+#include <filesystem>
+#include <fstream>
+#include <iomanip>
+#include <iostream>
+#include <limits>
+#include <stdexcept>
+#include <vector>
+
+#include <dpsim-models/EMT/EMT_Ph3_SSN_GFL.h>
```
leonardocarreras, line 17: I think DPsim.h should already include this

dpsim/examples/cxx/StateSpace/EMT_Ph3_GFL_Split_StateSpaceExtraction.cpp
```diff
@@ -0,0 +1,529 @@
...
+#include <dpsim-models/EMT/EMT_Ph3_SSN_GFL_Split.h>
```
leonardocarreras, line 18: I think DPsim.h should already include this

leonardocarreras (approved):
Hi @georgii-tishenin! We can try this, I leave some comments but you can evaluate if you want to have a follow-up. 
Some are cosmetic, some are about probably with chances to have faster performance (like gating behaviour for extraction).

PR #691, opened 2026-08-19
Further enhancement of switch model

### LINKED ISSUE #688: Further enhancements to switch models
"First of all: I think this is a great contribution. three-phase support and EMT variant are missing today, nice

Independently of the switch models, the DP::Ph3::Inductor and DP::Ph3::Resistor power-flow initialization fixes (missing RMS3PH_TO_PEAK1PH, and the resistor building phases B/C from an unassigned mIntfVoltage) are valid bug fixes

On the exponential mode: why not extend varResSwitch instead? DP::Ph1::varResSwitch already implements a geometric resistance ramp, just single-phase, DP/SP only, with an unparameterized transition duration and mutation of R_open/R_closed. I think the approach here is sound and the explicit switching time is an improvement - but then the follow-up question is: should varResSwitch be deprecated and everything based on the new implementations? @leonardocarreras 

DP CurrentZero mode: detecting the physical zero doesnt remove the transient. So we have the correct point in time, but still the problem that the trapezoidal companion model cannot represent an envelope discontinuity, so we have the transient. In my work i thought about zeroing the history terms to get EMT like continous behaviour here, but as switching would need topology information i decided to skip this approach and went with the Exponential approximation
So if u have any idea on how this could be done differently, that would be great - otherwise we should think about naming it differently then "CurrentZero" mode in DP - because at least my expectation would be a "EMT like behaviour" based on the naming"

_Originally posted by @FGH-wirtz in https://github.com/sogno-platform/dpsim/issues/687#issuecomment-5314708855_

FGH-wirtz:
Follow-up to #687, covers the first two points of #688.

Three commits, because they are independent of each other:

  1. feature (models): add switching modes to DP::Ph1::Switch — #688 point 1. DP::Ph3 has Ideal / CurrentZero / ExponentialZCSEmulation,
  DP::Ph1 had none of them. Base::Ph1::Switch gets the virtual closeSwitch()/openSwitch()
  hooks that Base::Ph3::Switch already has; close()/open() call them, so existing users are unaffected. The mode benchmark now also runs DP::Ph1, 18 runs in total, and is renamed since it is no longer Ph3 only.

  2. fix (models): correct DP Ph3 capacitor initialization — the same defect 2efee228 fixed for the inductor and resistor: the missing
  RMS3PH_TO_PEAK1PH conversion. Unrelated to the switch models, but commit 3 puts a capacitor at the switched bus, and without this fix  its startup transient is larger than the switching transient that should be measured there. Kept separate

  3. feature (models): ramp the resistance on closing in exponential ZCS mode — #688 point 2. Closing into capacitance has the same problem as interrupting inductive current: the DP envelope cannot represent the step. The exponential mode now traverses the same resistance path in both directions in EMT::Ph3, DP::Ph3 and DP::Ph1. Ideal and CurrentZero still close immediately. Benchmark and documentation cover both directions.
  
  Closes #688

### DIFF
--- a/docs/hugo/content/en/docs/Concepts/Models/switches.md
+++ b/docs/hugo/content/en/docs/Concepts/Models/switches.md
@@ -48,8 +48,9 @@ asks the network to move the capacitor voltage discontinuously, and the trapezoi
 of the capacitor answers with the same alternating error.
 
 This is a property of the discretisation, not of the physical circuit. A real breaker interrupts at a
-current zero and closes at a voltage zero, and the arc across real contacts dissipates what is left;
-a two-valued resistance switched at an arbitrary instant has neither mechanism.
+current zero, while controlled switching can aim to close at a favourable voltage instant. The arc
+across real contacts dissipates what is left; a two-valued resistance switched at an arbitrary
+instant has neither mechanism.
 
 ## What a current zero means per domain
 
@@ -58,7 +59,8 @@ Whether that is available depends on what the domain represents.
 
 In EMT the state variables are instantaneous quantities, so the current zero is directly visible.
 Waiting for it means the interrupted current is zero to within one time step and there is nothing
-left for the companion model to disagree about. This is physically and numerically correct.
+left for the companion model to disagree about. This is consistent with the intended ideal
+current-zero interruption model.
 
 In DP the state variables are complex envelopes of a carrier at the reference frequency
 $\omega_s$, and the instantaneous current can be reconstructed from them,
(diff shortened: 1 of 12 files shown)


FGH-wirtz:
Still open for discussion: for EMT::Ph3 the emulation of ZCS behaviour through an exponential resistance model is questionable, as this e.g. supresses inrush current for the closing side and EMT CurrentZero already perfectly implements ZCS behaviour. In DP we have a clear benefit here as explained above, in EMT this could potentially have some upsides when using a large timestep which could lead to artificial transients. To keep the options equal i kept it for now and tried to explain this in the new documentation, maybe making this option unavailable or giving a warning could be beneficial for EMT. Opinions on this are welcome :)

georgii-tishenin (approved):
Hi @FGH-wirtz, I reviewed and it looks good to me. Also, thanks for fixing the `DP::Ph3::Capacitor` initialization. I added a commit with minor typo corrections and rephrasing in documentation (hope, this is okay for you).

In my opinion it is okay to have `ExponentialZSCEmulation` option for EMT to keep the same options between EMT and DP.

Let us know if you have further ideas for enhancement of switch model behavior, especially for DP.

PR #693, opened 2026-08-21
Fix Norton matrix stamping for mixed DP Ph1 SSN components

georgii-tishenin:
This PR fixes the Norton-equivalent stamping of `DP::Ph1::MixedVTypeVariableSSNComp`.

Previously, the full real matrix

$$W = C B_d + D$$

was reduced to one complex scalar. This is only valid when $W$ has the complex-multiplication structure 

$$
\begin{bmatrix}
a & -b \\
b & a
\end{bmatrix}
$$

Mixed real/envelope models such as controlled converters do not generally satisfy this condition, so the reduction discarded part of their $d-q$ coupling.

The component now retains and stamps the full packed-real $2\times2$ Norton matrix. The same matrix is also used to reconstruct the interface current.

The DP Ph1 inverter state-space extraction example is extended to compare the extracted discrete eigenvalues against an independently assembled and trapezoidally discretized $dq$ model for time steps from $1 \mu s$ to $1\mathrm{ms}$. With full-matrix stamping, the models agree within numerical precision.

The documentation is updated to distinguish complex-linear DP components, whose Norton matrix can be represented by one complex scalar, from mixed real/envelope components that require a full real block.


leonardocarreras (approved):
Hi @georgii-tishenin! Thanks for the fix.

This error is present in the DP 3ph, eventually we would need a follow-up. Also there is likely the chance to assert against a smaller value for the tolerances.

PR #695, opened 2026-08-23
Fix DP Ph3 mixed SSN Norton stamping and initialization

### LINKED ISSUE #694: DP::Ph3::MixedVTypeVariableSSNComp folds a Norton matrix that mixed models do not satisfy
#693 fixed the Norton stamping of `DP::Ph1::MixedVTypeVariableSSNComp`: the real block `W = C * Bd + D` was reduced to one complex scalar, which is only valid when `W` has the `[[a, -b], [b, a]]` complex-multiplication structure. Mixed real/envelope models do not generally satisfy it, as it was explained in the PR.

`DP::Ph3::MixedVTypeVariableSSNComp` still performs the same reduction.

georgii-tishenin:
## Description

This PR fixes the Norton-equivalent representation and power-flow initialization of three-phase dynamic-phasor SSN components.

### Norton-equivalent stamping

`DP::Ph3::MixedVTypeVariableSSNComp` previously reduced the packed-real Norton matrix

$$
W = C B_d + D
$$

from $6\times6$ real form to a $3\times3$ complex matrix. This reduction is valid only when every $2\times2$ real block has the complex-multiplication structure

$$
\begin{bmatrix}
a & -b \\
b & a
\end{bmatrix}.
$$

Mixed real/envelope models such as controlled converters do not generally satisfy this condition. The reduction therefore discarded part of the coupling between real and imaginary envelope components, resulting in incorrect simulation and extracted eigenvalues.

The component now retains the full packed-real $6\times6$ Norton matrix. A corresponding MNA stamping function stamps its $2\times2$ real blocks directly, and the same full matrix is used when reconstructing the interface current.

### DP Ph3 initialization scaling

Power-flow node voltages are represented as three-phase line-to-line RMS phasors, while DP Ph3 simulation quantities use phase-to-ground peak complex envelopes. Components initialized from power-flow nodes must therefore apply

$$
V_{\mathrm{phase,peak}} = \sqrt{\frac{2}{3}},V_{\mathrm{LL,RMS}}.
$$

This conversion is added to:

* `DP::Ph3::MixedVTypeVariableSSNComp`,
* `DP::Ph3::TwoTerminalVTypeSSNComp`,
* `DP::Ph3::VoltageSource`, and
* `DP::Ph3::SeriesResistor`.

This prevents inconsistent initial states and startup transients when these components are initialized from a power-flow solution.

The change follows the phase-to-ground peak convention already used by the relevant DP Ph3 network components, including `PiLine`, `Resistor`, `Inductor`, `Capacitor`, and `Switch`. It is also consistent with the DP Ph3 initialization work introduced in PRs #687 and #691.

### State-space extraction and examples

State-space extraction support is added for the DP Ph3 inductor and mixed SSN components.

`EMT_Ph3_Inverter_StateSpaceExtraction.ipynb` is extended with the DP Ph3 averaged inverter. The DP model is included in the discrete- and continuous-eigenvalue comparisons, simulated-power plots, and participation-factor analysis.

`DP_Ph3_AvVoltSourceInverterStateSpace.ipynb` is updated so that its assertions, reconstructed waveforms, and plots consistently treat DP Ph3 quantities as phase-to-ground peak envelopes without applying an additional scaling factor.

Finally, the coding conventions are updated to distinguish:

* RMS quantities in the `SP` and `DP::Ph1` domains, and
* phase-to-ground peak quantities in the `DP::Ph3` and `EMT` domains.

Closes #694


leonardocarreras (commented):
Hi Georgii! Thanks for the fixes. Two small things that we might think about
- DP 3ph Resistor is not making yet the conversion the way the other components are doing, so is the different one
- Vref in EMT and in DP now might not match in the way they are calculated or have a different outcome. From power flow it is Peak 1ph, from setParameters is RMS 3ph, but maybe a bigger check and alignment in the meaning in all domains is needed.

georgii-tishenin:
> Hi Georgii! Thanks for the fixes. Two small things that we might think about
> 
>     * DP 3ph Resistor is not making yet the conversion the way the other components are doing, so is the different one
> 
>     * Vref in EMT and in DP now might not match in the way they are calculated or have a different outcome. From power flow it is Peak 1ph, from setParameters is RMS 3ph, but maybe a bigger check and alignment in the meaning in all domains is needed.

Hi @leonardocarreras. Thanks for the review. I applied the scaling convention for DP 3ph SeriesResistor. I agree, let's handle the clarity and consistency of voltage source Vref as a separate issue/PR.

leonardocarreras (approved):
Hi! Looks fine to me, thanks for the additional effort, lets try it. Yes, we can solve the rest separately.

PR #696, opened 2026-08-24
feat(models): EMT Ph3 PQLoad

matthiasmees:
Signed-off-by: SimonS0406 <simonschr06@gmx.de>

- Implemented new EMT_Ph3_PQLoaad with constant power behavior
- Tested against Load and voltage steps, including pybind


georgii-tishenin (approved):
Looks good to me!

PR #699, opened 2026-08-25
test(notebooks): skip slow notebook tests in CI

georgii-tishenin:
## Description

This PR marks the longest-running notebook tests as skipped using the existing `metadata.tests.skip` mechanism.

The skipped notebooks are simulation-heavy examples that currently dominate the pytest runtime in CI. They remain available as examples, but are no longer executed as part of the regular notebook test job.

## Motivation

The notebook pytest job currently takes around 40 minutes, with a small number of notebooks accounting for most of the runtime. Skipping these notebooks keeps the CI feedback loop shorter while preserving the notebooks in the repository.


leonardocarreras (approved):
Hi! I think we can try it

Issue #680, opened 2026-08-11
# Proposal / feature: event-driven variable time step for the MNA solvers


FGH-wirtz:
# Proposal / feature: event-driven variable time step for the MNA solvers

**Problem.** One step size has to satisfy the worst moment of a run as well as the best. On my 194-bus IEEE-118 grid version, DP at 20 µs costs 3.9x DP at 100 µs — and the coarse run is off significantly at fault clearing. A huge runtime or being wrong at fault event are the only options today.

**Proposal.** Event driven variable time step could be a solution here as we can take the benefits of large stepsize in dp in settled state and still be correct after fault events. I implemented and tested this locally and then found some descriptions of a similar method already in literature: Henschel 1999 ("Variable Integration Step Sizes"), Demiray Dissertation 2008 (ETH 17607), Gao & Strunz 2009 (FAST), Rupasinghe 2021 (Manitoba, BFAST)., so this seems to be a established feature 

**Issue / Problems**: dt is not solver-local in DPsim — it sits in every companion model, in the loggers' constant dt assumption, .... Realistically this is a multi-release project to get it done cleanly. Furthermore this implementation contradicts the proposal to use DP for real-time sim, as the fine steps could take too long - so worth discussing if main dpsim should be developed in this direction - should be implemented as "opt-in"

**Implementation** (tested / done in a fork for DP::Ph1, (DP::Ph3 and EMT::Ph3 partly)): 
(1) `mnaUpdateTimeStep(Real)` per component — rebuild every dt-dependent quantity
(2) a refinement window in `Simulation` that opens a lead time before a listed event or before an armed switch operates, and re-stamps on change - here not only the switch stamps but all components dependent on dt like every subcomp with companion model
(3) tested two options: a pre defined "fine run time" after event and a adaptive change. a continued "re-coarsement" would require constant refactorization so imo using max a few distinct dt values is beneficial 

**Measurements**, 194 buses / 494 components, fault 0.25–0.30 s, deviation of |V| per bus against the fine run on the shared time grid:

| DP::Ph1 | steps | time [s] | ms/step | vs coarse | vs fine | p95 deviation | worst deviation |
|---|---|---|---|---|---|---|---|
| coarse 1e-4 | 4 000 | 25.1 | 5.03 | 1.00x | 0.25x | 0.8227 % | 800 % at 0.3012 s (fault clearing) |
[...]

PR #701, opened 2026-08-26
feat: Variable time step for the MNA solver (DP domain)

FGH-wirtz:
First stage of #680. There i proposed implementing an event-driven variable step so a run can use a large step in the settled state and a fine step after a fault. This branch starts with DP domain and scheduled events.

What it adds

  setTimeStep() only works before a run starts, because components build their companion models from the step at initialisation and keep them. New: Simulation::updateTimeStep(Real) changes the step properly between steps, and setEventRefinement(fine, lead, follow) uses a finer step around scheduled events.

  A change travels Simulation::updateTimeStep() → MnaSolver::updateTimeStep() → MNAInterface::mnaUpdateTimeStep() per component. A component returns false if it cannot follow; the solver counts those, logs the declining types and returns the count, so a stale discretisation is reported rather than silently simulated.

  Two solver details:

  - The base matrix holds the static components with the old step, and recomputeSystemMatrix() only runs when a variable component reports a change. refreshStaticMatrixStamp() re-stamps it and factorises fully, since partialRefactorize() only covers variable components.
  - With precomputed switch matrices every factorisation carries the old step, so the change is refused with a warning. Matrix recomputation must be enabled.

  Coverage and limits

  All of DP::Ph1 and DP::Ph3: passive elements, ideal sources, containers, switches, and the reduced-order and PCM/TPM machines. Machines also re-anchor the rotor angle by mBase_OmMech * (oldTimeStep - timeStep), otherwise the dq transform argument jumps by w0*dt and the machine rotates against the network.

  mnaUpdateTimeStep() defaults to false, so EMT and SP decline and warn, and existing runs are untouched. Still declining in DP on purpose: the converter families, SynchronGeneratorTrStab, SynchronGeneratorIdeal, VoltageSourceRamp, Ph3::SynchronGeneratorDQODE. 
Refinement covers scheduled events only for now. Important: Logs get non-uniform time stamps, so anything downstream has to read the time column instead of assuming a constant step.

  Testing

  DP_VarTimeStep in dpsim/examples/cxx/Components/, six checks: 
- nothing changes on an unchanged step 0 
- coarse-to-fine against a run held fine 4.4e-10 
- decline count 0 passive and 1 with a ramped source 
- refined window on a switching transient 2.9e-2 at the base step against
  5.3e-8 refined; 
machine rotor angle 2.3e-11.

  The machine case is a 24 kV SMIB with a 4th-order VBR generator. Remove the rotor re-anchoring and the same check reads 2.3e-1, so it measures that correction rather than passing by construction.

  Documented under Developer Guide, Solvers, next to the MNA solver page.

An LLM was used to rework my local implementation for this merge request - i verified changes and modified some things manually.

### DIFF
--- a/dpsim/examples/cxx/Components/DP_VarTimeStep.cpp
+++ b/dpsim/examples/cxx/Components/DP_VarTimeStep.cpp
@@ -1,0 +1,224 @@
+/* Author: Christoph Wirtz <christoph.wirtz@fgh-ma.de>
+ * SPDX-FileCopyrightText: 2026 FGH e.V.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+// Checks Simulation::updateTimeStep() on a network of DP passive elements.
+//
+//   [1] Changing to the step already in use must reproduce the undisturbed run.
+//       Any deviation is produced by the conversion itself.
+//   [2] Changing from a coarse to a fine step must stay close to a run held at
+//       the fine step throughout.
+//   [3] The return value must count the components that did not convert.
+
+#include <algorithm>
+#include <cmath>
+#include <iomanip>
+#include <iostream>
+#include <vector>
+
+#include <DPsim.h>
+
+using namespace DPsim;
+using namespace CPS;
+
+namespace {
+
+struct Parameters {
+  Real frequency = 50.0;
+  Real sourceVoltage = 100e3;
+  Real sourceResistance = 10.0;
+  Real inductance = 50e-3;
+  Real capacitance = 1e-6;
+  Real loadResistance = 100.0;
+
+  Real coarseStep = 200e-6;
+  Real fineStep = 50e-6;
+  Real changeTime = 0.1;
+  Real finalTime = 0.2;
+};
+
+/// Source - R - L - probe, with C and a load resistor from probe to ground.
+SystemTopology build(const Parameters &p,
+                     std::shared_ptr<SimNode<Complex>> &probe) {
+  auto n1 = SimNode<Complex>::make("n1", PhaseType::Single);
+  auto n2 = SimNode<Complex>::make("n2", PhaseType::Single);
+
+  n1->setInitialVoltage(Complex(p.sourceVoltage, 0.0));
+
+  auto vs = DP::Ph1::VoltageSource::make("vs", Logger::Level::off);
+  vs->setParameters(Complex(p.sourceVoltage, 0.0));
+
+  auto res = DP::Ph1::Resistor::make("R_src", Logger::Level::off);
+  res->setParameters(p.sourceResistance);
+
+  auto ind = DP::Ph1::Inductor::make("L", Logger::Level::off);
+  ind->setParameters(p.inductance);
+
+  auto cap = DP::Ph1::Capacitor::make("C", Logger::Level::off);
+  cap->setParameters(p.capacitance);
+
+  auto load = DP::Ph1::Resistor::make("R_load", Logger::Level::off);
+  load->setParameters(p.loadResistance);
+
+  vs->connect({SimNode<Complex>::GND, n1});
+  res->connect({n1, n2});
+  ind->connect({n2, SimNode<Complex>::GND});
+  cap->connect({n2, SimNode<Complex>::GND});
+  load->connect({n2, SimNode<Complex>::GND});
+
+  probe = n2;
+
+  return SystemTopology(p.frequency, SystemNodeList{n1, n2},
+                        SystemComponentList{vs, res, ind, cap, load});
+}
+
+struct Trace {
+  std::vector<Real> time;
+  std::vector<Complex> voltage;
+  UInt unconverted = 0;
+};
+
+/// changeTo <= 0 runs at startStep throughout.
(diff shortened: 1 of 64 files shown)


leonardocarreras:
@FGH-wirtz thanks for the PR and the nice work! I will need a bit of time to understand and test, will get back to you with questions soon.

dpsim/src/MNASolverDirect.cpp
```diff
@@ -136,6 +136,32 @@ void MnaSolverDirect<VarType>::solveWithSystemMatrixRecomputation(
...
 }
 
+template <typename VarType>
+void MnaSolverDirect<VarType>::refreshStaticMatrixStamp() {
+  if (!mSystemMatrixRecomputationEnabled) {
+    // The precomputed matrices were factorised with the old timestep.
+    SPDLOG_LOGGER_WARN(mSLog,
+                       "Time step changed while using precomputed switch "
+                       "matrices. They keep the factorisation of the old step; "
+                       "enable system matrix recomputation to change it.");
```
georgii-tishenin, line 146: I suggest throwing an exception with clear message (maybe same as you wrote) rather than logging a warning. This is clearly a wrong combination of settings. The warning can be overlooked by the user.

dpsim/src/MNASolver.cpp
georgii-tishenin, line 469: in one of the recent PRs we have added this logic to automatically decide whether recomputation should be enabled or not based on the components in the topology. Could variable timestep be integrated into the existing `SystemMatrixRecomputationMode::Auto` logic?

If event refinement / variable timestep is enabled before initialization, `Auto` could select system-matrix recomputation automatically, because timestep-dependent companion stamps need to be rebuilt when `dt` changes. If the user explicitly selects `SystemMatrixRecomputationMode::Disabled` together with variable timestep, I would prefer throwing a `SystemError` instead of warning, since continuing with a factorization from the old timestep is not a valid configuration.

For manual `updateTimeStep()` calls after start, the solver should probably still throw if the current matrix mode cannot support timestep changes.

georgii-tishenin (approved):
Hi, thanks for the PR. From my point of view looks good. I have added two optional comments (if you do not agree with them, just resolve). Same goes for the comments from the GitHub actions bot.

Let's also wait for the review from @leonardocarreras

dpsim/src/Simulation.cpp
```diff
@@ -472,12 +472,46 @@ void Simulation::run() {
...
+
+  const Real wanted = wantFine ? mRefineTimeStep : mRefineBaseTimeStep;
+  const UInt unhandled = updateTimeStep(wanted);
+
+  mRefineInWindow = wantFine;
+  SPDLOG_LOGGER_INFO(mLog, "t = {:e}s: time step {:e}s, {:s} refinement window",
+                     mTime, wanted, wantFine ? "entering" : "leaving");
+
+  if (unhandled > 0)
+    SPDLOG_LOGGER_WARN(mLog,
```
leonardocarreras, line 499: I would  say we want here a logger error (and maybe a throw), so is clear for the user that the result is not good. Do we want to list somehow the components that failed, for the user?

dpsim/src/Simulation.cpp
```diff
@@ -472,12 +472,46 @@ void Simulation::run() {
...
+  return unhandled;
+}
+
 Real Simulation::step() {
   std::chrono::steady_clock::time_point start;
   if (mLogStepTimes) {
     start = std::chrono::steady_clock::now();
   }
 
+  applyEventRefinement();
```
leonardocarreras, line 513: is this now running unconditionally on every step? if yes please just confirm that it does not add a lot of operations in the critical path for the case when is inactive (looks like it does not)

dpsim/examples/cxx/Components/DP_VarTimeStep.cpp
```diff
@@ -0,0 +1,534 @@
...
+
+#include <algorithm>
+#include <cmath>
+#include <iomanip>
+#include <iostream>
+#include <vector>
+
+#include "../Examples.h"
+#include <DPsim.h>
+#include <dpsim-models/DP/DP_Ph1_Switch.h>
```
leonardocarreras, line 19: DP_Ph1_Switch is not part of DPsim.h?

leonardocarreras (commented):
Hi @FGH-wirtz! After a short delay, some comments and questions on my side, but in general I say that I agree with the changes and approach:
- What is your opinion about just failing if the components cannot do it? I think for the composite components is problematic, because the subcomponent changes, then the parent may not support and then they even stay in some middle inconsistent state, or that is what I understand from following the code.
- I would prefer that we have some automatic mode to select or even just make sure that we always activate `doSystemMatrixRecomputation()` when this characteristic is active, so we avoid strange errors for the user.
- Some components (that should not be signals, but we did not merge the refactor yet, like the Decoupling Lines see #666 #667) also have subcomponents, and then it mostly likely will fail silently because they (Decoupling Line should be a Composite Component / parent) are not reached now by the updating but the internal resistor and source will do. The delay is here related to the timestep, and the model may not be able to run in certain conditions. This is probably something that we should carry into its own issue.

PR #702, opened 2026-08-27
ci: update test- pypi publishing version resolution and triggers

SystemsPurge:
Notable changes:
Resolve version substituted in pyproject.toml, depending on use case. 
Packaging-python is called from two different workflows: ci.yaml and publish_to_pypi.
If its called from ci, default version is '' and substitution is skipped, and no subsequent publishing seems to be happening
If its called from publish_to_pypi then:
If its a manually triggered workflow, the version is given the <version>.dev<run_number> treatment
Publishing to testpypi will trigger if the version string contains either dev or rc
Otherwise the provided tag is substituted into the pyproject.toml anyway , and the publishing to pypi triggers only if the tag ( NOT THE CALCULATED VERSION STRING ) doesnt contain -rc
This should cover all use cases, but please do check whether it is correct, as i cannot test it easily

### DIFF
--- a/.github/workflows/packaging-python.yaml
+++ b/.github/workflows/packaging-python.yaml
@@ -10,6 +10,11 @@ on:
         description: 'Revision under test'
         required: true
         type: string
+      version:
+        description: 'Version to set in the pyproject.toml'
+        required: false
+        type: string
+        default: ''
       image:
         description: 'Development image the source distribution is built in'
         required: true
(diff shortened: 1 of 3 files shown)


SystemsPurge:
This should now do the following:
Version resolution also takes into account rc tag push
First: Publish to testpypi and publish to pypi are mutually exclusive: testpypi if workflow_dispatch or tag contains rc, and pypi the opposite

If workflow_dispatched or tag contains rc , version is resolved as follows:
If ref_type is branch, fetch the latest tag, otherwise use the provided tag
If tag is rc, remove dash
If tag is not rc, split, increment z ( vx.y.z ), add .dev<run_number>

SystemsPurge:
Will handle this in a larger refactor of the workflows, where triggers are split, and context is clearer

PR #704, opened 2026-08-30
fix(models): correct the SP::Ph3 voltage source current readback and stamp

leonardocarreras:
Fixes to the Voltage Source in SP 3 phase domain
- Stamp all phases
- Update as a complex value in the current


georgii-tishenin (approved):
Thanks for the fix!

PR #705, opened 2026-08-30
feat(models): replace the transformer snubbers with a magnetizing branch

leonardocarreras:
- Replace transformer with a T-equivalent
- Remove auto-flip, replaced with a resolver (finds the highest voltage side)
- Remove snubbers
- Sizing of the defaults give a sensible parameter of magnetization
- Update tests, examples, docs, notebooks


dpsim-models/src/CIM/Reader.cpp
```diff
@@ -619,6 +619,59 @@ Reader::mapPowerTransformer(CIMPP::PowerTransformer *trans) {
     resistance = end1->r.value / std::pow(ratioAbsNominal, 2);
   }
 
+  CIMPP::PowerTransformerEnd *referenceEnd =
+      (voltageNode1 >= voltageNode2) ? end1 : end2;
```
georgii-tishenin, line 623: Non-blocking / future check: currently the magnetizing `g`/`b` values are read from the selected reference end only. If some CIM files provide these values on the other `PowerTransformerEnd`, we may silently fall back to the defaults. Or are they always provided in CIM for the high voltage end?

dpsim-models/src/SP/SP_Ph3_Transformer.cpp
```diff
@@ -0,0 +1,272 @@
...
+          Math::complexFromVectorElement(leftVector, matrixNodeIndex(0, phase));
+  }
+}
+
+void SP::Ph3::Transformer::mnaCompUpdateCurrent(const Matrix &leftVector) {
+  for (UInt phase = 0; phase < 3; phase++)
+    (**mIntfCurrent)(phase, 0) = Math::complexFromVectorElement(
+        leftVector,
+        mVirtualNodes[1]->matrixNodeIndex(static_cast<PhaseType>(phase)));
+}
```
georgii-tishenin, line 271: Non-blocking / future check: for the Ph1 and EMT Ph3 transformer implementations, `mIntfCurrent` is adjusted with the orientation sign when the reference winding is not terminal 0. Should we check whether the same convention is needed here for the DP/SP Ph3 transformer as well, especially for `setReferenceWinding(Secondary)`?

georgii-tishenin (dismissed):
Thanks, this looks good to me overall. The transformer refactor towards a T-equivalent with an explicit magnetizing branch is a clear improvement over the previous snubber-based workaround.

I left a few non-blocking comments/questions that may be useful to check in a follow-up, but I would not require them for this PR.

dpsim-models/src/SP/SP_Ph3_Transformer.cpp
leonardocarreras: Fixed

dpsim-models/src/CIM/Reader.cpp
leonardocarreras: Rule from ENTSO-E CGMES implementation guidelines says that parameters are provided for end numbered as 1, and highest voltage should be on the end 1, so lets say that they provide in a different place, it would be a file non-conform to standard. 
As those kind of files exist (even in our own example data, but they are zero so we do not test), yes, we should know that we would be providing a default at the moment, eventually we could move it to other end and use the ratio to scale it.

georgii-tishenin (dismissed):

georgii-tishenin (approved):

PR #714, opened 2026-09-15
build: add native macOS support

matthiasmees:
# build: add native macOS support

## Summary

This PR adds native macOS support for building and developing DPsim, including Apple Silicon support.

It provides a reproducible Homebrew-based setup while keeping the regular CMake workflow for normal development.

The user and developer documentation have also been updated to describe the native macOS installation and development workflow.

## Changes

* add `packaging/Shell/install-macos.sh`
* add native Apple Silicon / `arm64` support
* add Homebrew dependency setup for:
  * CMake
  * Ninja
  * Eigen 3
  * Graphviz
  * OpenMP (`libomp`)
  * Python 3.14
  * Ruby
  * Node.js
  * `pkg-config`
* add macOS OpenMP detection for AppleClang and Homebrew `libomp`
* improve Graphviz compatibility and header detection
* improve native `std::filesystem` detection
* fix Eigen-dependent template compilation on macOS
* add macOS handling for the real-time library
* improve Python packaging and CMake argument forwarding
* create a local `dpsim-python` virtual environment
* build and install the Python bindings via `pybind11`
* install Jupyter and register a `DPsim Python` kernel
* install and validate pre-commit dependencies
* add a native macOS installation section to the User Guide
* update the Developer Guide with the native macOS CMake workflow
* document clean rebuilds, Python environment activation and normal incremental rebuilds

## Usage

Make the setup script executable:

```bash
chmod +x packaging/Shell/install-macos.sh
```

Run the initial setup:

```bash
./packaging/Shell/install-macos.sh
```

For a completely clean setup:

```bash
CLEAN=1 ./packaging/Shell/install-macos.sh
```

After the initial configuration, activate the Python environment and use the regular CMake workflow:

```bash
source dpsim-python/bin/activate
cmake --build build --parallel "$(sysctl -n hw.ncpu)"
```

The required macOS-specific CMake configuration is stored in the build directory and does not need to be supplied again for normal rebuilds.

If changes affect the Python bindings or Python package, reinstall it into the active environment:

```bash
python -m pip install .
```

## Documentation

The documentation now includes a native macOS installation path for users and a corresponding development workflow for contributors.

The User Guide documents:

* installation through `packaging/Shell/install-macos.sh`
* the local `dpsim-python` environment
* the registered `DPsim Python` Jupyter kernel
* clean rebuilds with `CLEAN=1`
* the default optional-feature limitations on macOS

The Developer Guide documents:

* the macOS setup script and Homebrew dependencies
* the configured CMake build tree
* incremental rebuilds with `cmake --build build`
* Python package reinstallations after binding changes
* the current status of CIM/CGMES and VILLASnode support on macOS

## Tested

Tested successfully with:

* macOS on Apple Silicon (`arm64`)
* AppleClang
* Homebrew
* Python 3.14
* Ninja
* OpenMP via Homebrew `libomp`
* Graphviz
* SuiteSparse / KLU
* C++ examples
* `dpsimpy`
* Jupyter
* pre-commit

The native CMake build successfully produces the Python extension, and the installed environment successfully imports:

```python
import dpsim
import dpsimpy
```

Validation performed with:

```bash
cmake --build build --parallel "$(sysctl -n hw.ncpu)"
python -m pip install .
pre-commit run --all-files
```


setup.py
```diff
@@ -1,26 +1,93 @@
...
+#
+# Core DPsim:
+#   pip install .
+#
+# DPsim + VILLAS:
+#   DPSIM_PYTHON_WITH_VILLAS=1 pip install .
+#
+# This avoids assuming that VILLASnode exists simply because the platform
+# is Linux or macOS.
+WITH_VILLAS = env_flag("DPSIM_PYTHON_WITH_VILLAS", default=False)
```
leonardocarreras, line 50:
```suggestion
WITH_VILLAS = env_flag("DPSIM_PYTHON_WITH_VILLAS", default=platform.system() == "Linux")
```

setup.py
```diff
@@ -31,96 +98,186 @@ def build_extension(self, ext):
...
+            "-DDPSIM_BUILD_EXAMPLES=OFF",
+            "-DDPSIM_BUILD_DOC=OFF",
+            # Reproducible dependencies for package builds.
+            "-DFETCH_SPDLOG=ON",
+            "-DFETCH_SUITESPARSE=ON",
+            # Eigen is supplied by the platform/development environment.
+            # Windows already fetches Eigen in the top-level CMake logic.
+            "-DFETCH_EIGEN=OFF",
+            # VILLAS is explicitly optional.
+            f"-DWITH_VILLAS={'ON' if WITH_VILLAS else 'OFF'}",
```
leonardocarreras, line 114: We need to mark it the other way, I think. I would argue for true by default, or true if linux.

setup.py
```diff
@@ -31,96 +98,186 @@ def build_extension(self, ext):
...
             "-DCMAKE_POLICY_VERSION_MINIMUM=3.5",
+            # Python bindings are the purpose of this build.
+            "-DWITH_PYBIND=ON",
+            "-DFETCH_PYBIND=OFF",
+            # Keep Python package builds small.
+            "-DDPSIM_BUILD_EXAMPLES=OFF",
+            "-DDPSIM_BUILD_DOC=OFF",
+            # Reproducible dependencies for package builds.
+            "-DFETCH_SPDLOG=ON",
+            "-DFETCH_SUITESPARSE=ON",
```
leonardocarreras, line 109: We do not want to fetch the things that are already installed (but you could do it for Mac if needed). Eventually the idea would be not to fetch anymore but only find.

cmake/FindGraphviz.cmake
```diff
@@ -6,14 +6,27 @@ find_library(CGRAPH_LIBRARY
...
+# Typical locations:
+#   Linux:   /usr/include/graphviz
+#   macOS:   /opt/homebrew/include/graphviz
+#   Windows: <Graphviz>/include/graphviz
+#
+# Some installations may place the headers directly in include/, so keep
+# that as a fallback.
 find_path(GRAPHVIZ_INCLUDE_DIR
 	NAMES
-		graphviz/cgraph.h
```
leonardocarreras, line 15: why is this needed?

dpsim-models/src/Graph.cpp
```diff
@@ -1,12 +1,12 @@
...
+/*
+ * SPDX-FileCopyrightText: 2017-2021 Institute for Automation of Complex Power Systems,
+ *                         EONERC, RWTH Aachen University
+ * SPDX-License-Identifier: MPL-2.0
+ */
 
-#include <graphviz/gvc.h>
+#include <cstddef>
+
+#include <gvc.h>
```
leonardocarreras, line 9: is this needed

pyproject.toml
```diff
@@ -74,6 +74,7 @@ test-command = "python -c \"import dpsimpy; import dpsim\""
 [tool.cibuildwheel.linux]
 build = "cp3{10,11,12,13,14}-manylinux_x86_64"
 manylinux-x86_64-image = "sogno/dpsim:manylinux"
+environment = { DPSIM_PYTHON_WITH_VILLAS = "1" }
```
leonardocarreras, line 77: This might be not needed if we keep the defaults as they were

dpsim/src/MNASolver.cpp
```diff
@@ -747,7 +747,7 @@ void MnaSolver<VarType>::steadyStateInitialization() {
     diff = prevLeftSideVector - **mLeftSideVector;
     prevLeftSideVector = **mLeftSideVector;
     maxDiff = diff.lpNorm<Eigen::Infinity>();
-    max = (**mLeftSideVector).lpNorm<Eigen::Infinity>();
+    max = (**mLeftSideVector).template lpNorm<Eigen::Infinity>();
```
leonardocarreras, line 750: why is it required for `max` and not for `maxDiff`

leonardocarreras (commented):
Hi Matthias! Thanks for the PR, there are quite a lot of nice additions here. You catched a lot of small bugs and errors, like the OpenMP import and the RealTimeDatalogger mismatches. The setup script also became a bit better in general. Nice things overall!

Just a few nits, related to how things work or were changed here
- I would avoid to flip the normal behaviour, which is having the villas support. It might make the pip install or the nix build to avoid having villas without the user realizing. It also breaks the default convention (defaults are in cmake).
- setup.py is not the best place in my view to put defaults, also because that only affects python builds, so if any, that should be the cmake options in the cmakelists. I think changing the defaults is a nice discussion, but is a separate thing from macOS support
- in general, the mac script does a bit more than just setting up things for dpsim (for example, you can compare to the spirit on the fedora one), I would suggest we can move it to somewhere not in packaging, as we are not exactly installing deps only or maybe split in, and also changes here affect when we publish containers (or we need to change that filter)... maybe more opinions here are better...
- do you want to pin to 3.14 exacty for mac? is there anything that requires this in the script? we said we support 3.10+, so if this is not true for mac is also something good to know
- the pre-commit line might be deleting some other existing stuff from the user, might be good to ask first
- in general, the comments are long or a bit wordy

stv0g:
With some minor fixes, our Nix-build also works on aarch64-darwin (inluding dpsimpy, but without CIMpp and VILLAS):

#715

leonardocarreras (approved):
Once again thanks to Matthias! Did some changes agreed with him myself, will merge, we make any other fixes separately in follow-up PRs if needed.

PR #715, opened 2026-09-16
Fix Nix builds on Darwin (and dpsimpy Nix build on Linux)

stv0g:
Related to #714


stv0g:
Tested with

```
PYTHONPATH=$(nix build --print-out-paths github:sogno-platform/dpsim/nix-darwin#dpsimpy)/lib/python3.12/site-packages \
python3
```

leonardocarreras:
Thanks for the PR @stv0g ! I think the failures in CI here are not caused by this PR and are solved in #653 and #654, will try to get them going and come back to this

